PR #14, opened 2024-05-25
[Feature] Allow professor to create their own semester

d3li0n:
# Description

Allow professors to create their own semesters instead of selecting them from the dropdown list

## Type of change

# How Has This Been Tested?

- [x] Updated existing unit tests
- [x] Created a new course with a semester that doesn't exist. Confirmed that the course is created and semester is added to database
- [x] Created a new course with the existing semester; confirmed that the course is created and the new semester record is not added. Confirm via course edit page that semester is correct
- [x] Modified existing course with the new semester and confirmed that information about course updated
- [x] Modified existing course with the same semester that does exist and confirmed that the information about course updated

# Checklist:


packages/app/pages/organization/course/[cid]/edit.tsx
```diff
@@ -113,7 +111,8 @@ export default function Edit(): ReactElement {
         sectionGroupNameField === courseData.course.sectionGroupName &&
         zoomLinkField === courseData.course.zoomLink &&
         courseTimezoneField === courseData.course.timezone &&
-        semesterIdField === courseData.course.semesterId &&
+        semesterNameField ===
```
Comment (d3li0n) [inline packages/app/pages/organization/course/[cid]/edit.tsx:114]: Unfortunately, that's how the semester model is structured. We don't want to modify the existing model as it can cause production to break. We would need to eliminate the semester model to make the new version of the model. Given the time, it's easier to accommodate with this approach.

I will probably refactor the model in T2 of the summer.

wskksw (approved):

packages/app/pages/organization/course/[cid]/edit.tsx
```diff
@@ -153,12 +152,10 @@ export default function Edit(): ReactElement {
         return
       }
 
-      // if semesterIdField is not a number or not in semesters
-      if (
-        isNaN(semesterIdField) ||
-        !semesters.find((semester) => semester.id === semesterIdField)
-      ) {
-        message.error('Semester is invalid')
+      if (semesterNameField.split(',').length !== 2) {
```
Comment (AdamFipke) [inline packages/app/pages/organization/course/[cid]/edit.tsx:155]: might want to add a check to make sure that the second item is a number. I managed to cause an error by putting `nuh,uh` as my semester

packages/app/pages/organization/course/[cid]/edit.tsx
```diff
@@ -273,16 +270,10 @@ export default function Edit(): ReactElement {
...
+                      name="semesterName"
                       tooltip="Semester of the course"
                     >
-                      <Select>
-                        {semesters.map((semester) => (
-                          <Select.Option value={semester.id} key={semester.id}>
-                            {semester.season + semester.year}
-                          </Select.Option>
-                        ))}
-                      </Select>
```
Comment (AdamFipke) [inline packages/app/pages/organization/course/[cid]/edit.tsx:285]: honestly I'm not sure if it matters, but is there a way to list all other semesters as well? That way, we don't have like multiple of the same semester (e.g. "Winter,2024" and "Winter,24", and maybe "winter,2024")

packages/server/src/semester/setCourse.command.ts
```diff
@@ -1,104 +0,0 @@
...
-import { Season } from '@koh/common';
-import { SemesterService } from './semester.service';
-import { SemesterModel } from './semester.entity';
-
-@Injectable()
-export class SetSemesterCommand {
-  constructor(private readonly semService: SemesterService) {}
-  @Command({
-    command: 'semester:toggleActiveSemester  <mode> <semester> <year>',
-    describe: '(disable or enable) all the classes in a given semester',
```
AdamFipke, line 12: Wait this seems like a useful feature?

Unless i'm mistaken and it doesn't actually work

AdamFipke (changes requested):
left some comments

packages/app/pages/organization/course/[cid]/edit.tsx
Comment (d3li0n) [reply packages/app/pages/organization/course/[cid]/edit.tsx:285]: Ramon specifically asked to allow professors to input any semester name or variation.

packages/server/src/semester/setCourse.command.ts
d3li0n: This shouldn't really be used in production + we have seed controller. So I don't think we need a command.

packages/app/pages/organization/course/[cid]/edit.tsx
Comment (AdamFipke) [reply packages/app/pages/organization/course/[cid]/edit.tsx:285]: oh i see, aight

Comment (d3li0n) [reply packages/app/pages/organization/course/[cid]/edit.tsx:285]: Yeah, that's my understanding. I am also okay with duplicate semesters as it would require a more complex logic to check for duplicates

packages/app/pages/organization/course/[cid]/edit.tsx
Comment (d3li0n) [reply packages/app/pages/organization/course/[cid]/edit.tsx:155]: Fixed

AdamFipke (approved):

PR #15, opened 2024-05-25
[Bug Fix / Feature] Allow admins to change user roles when sso is enabled for organization

d3li0n:
# Description
- Allows organization admins to change user roles in the organization
- Disables any course modifications on the front end when the user is not a professor in the organization
- Fixes an issue where code in the backend incorrectly handled async and await calls
- Fixes an issue where the empty user's last name of the user was displayed as `null` instead of an empty string

## Type of change

# How Has This Been Tested?

- [x] Manually tested frontend and backend functions to verify that user roles can be modified when SSO is enabled

# Checklist:

### DIFF
--- a/packages/app/pages/organization/course/[cid]/edit.tsx
+++ b/packages/app/pages/organization/course/[cid]/edit.tsx
@@ -72,7 +72,7 @@ export default function Edit(): ReactElement {
     const [formGeneral] = Form.useForm()
     const semesters = useSemester()
 
-    const { data: courseData, error } = useSWR(
+    const { data: courseData, _error } = useSWR(
       `api/v1/organization/[oid]/course/[cid]`,
       async () =>
         await API.organizations.getCourse(organization.id, Number(cid)),
--- a/packages/app/pages/organization/course/[cid]/edit.tsx
+++ b/packages/app/pages/organization/course/[cid]/edit.tsx
@@ -215,14 +215,6 @@ export default function Edit(): ReactElement {
                 }}
                 onFinish={updateGeneral}
               >
-                {courseData.profIds.length !== professors?.length && (
-                  <Alert
-                    message="System Notice"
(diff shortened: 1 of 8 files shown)


packages/app/pages/organization/course/[cid]/edit.tsx
```diff
@@ -214,14 +215,27 @@ export default function Edit(): ReactElement {
...
                 onFinish={updateGeneral}
               >
+                {courseData.profIds.length !== professors?.length && (
+                  <Alert
+                    message="System Notice"
+                    description="One or more professors in this organization is not assigned a professor role. Please assign a professor role to all professors in this course."
+                    type="error"
+                    style={{ marginBottom: 20 }}
+                  />
+                )}
```
Comment (AdamFipke) [inline packages/app/pages/organization/course/[cid]/edit.tsx:225]: Bug: This seems to lock me out of editing anything on my newly created course.

Steps to reproduce:
1. Create a course with Ramon Lawrence as a professor
2. Click edit course (warnings should show up)

Comment (AdamFipke) [reply packages/app/pages/organization/course/[cid]/edit.tsx:225]: After looking at this file, it seems like `professors` is the total professors for the *organization*, not the course. So it seems like doing `courseData.profIds.length !== professors?.length` is comparing the total professors in an *organization* and the number of professors in the *course* and seeing if they are the same (and they basically never would be)

Lines 81-84: 
```ts
 const { data: professors } = useSWR(
      isAdmin ? `/api/v1/organization/[oid]/get_professors` : null,
      async () => await API.organizations.getProfessors(organization.id),
    )
```

AdamFipke (changes requested):

packages/app/pages/organization/course/[cid]/edit.tsx
Comment (d3li0n) [reply packages/app/pages/organization/course/[cid]/edit.tsx:225]: Okay, I will remove this check since I asked Ramon to update Scott's role to prof

Comment (d3li0n) [reply packages/app/pages/organization/course/[cid]/edit.tsx]: Reverted

AdamFipke (approved):

PR #18, opened 2024-06-01
[Performance Optimization] Improve API and database response on large number of requests

d3li0n:
# Description

Adds restrictions on most demanded API endpoints to prevent system crash both on front and back-end.

## Type of change

# How Has This Been Tested?

WIP

# Checklist:

- [ ] Any dependent changes have been merged and published in downstream modules


packages/server/src/course/course.controller.ts
```diff
@@ -59,23 +57,16 @@ import { ConfigService } from '@nestjs/config';
 @Controller('courses')
 @UseInterceptors(ClassSerializerInterceptor)
 export class CourseController {
+  readonly MAX_ASYNC_QUESTIONS = 100;
+  readonly MAX_QUEUES_PER_COURSE = 30;
```
AdamFipke, line 61: Okay this might sound weird, but with some of the work i'm doing with demos and stuff, we're basically making it so each queue is an assignment (Ramon strongly recommend doing it like this), and 30 *might* be a little too low, or at least for some of the larger courses. Right now, I have it so it's recommended to have each queue be an assignment (so maybe upwards of 15 queues or something), but some professors may prefer to create a new queue for each lab section and assignment, leading to probably upwards of over 100 queues (e.g. 10 labs with 10 lab sections). Granted, when my feature comes, it would need to come with a way to easily hide/show/archive queues, but I thought maybe this is worth mentioning to maybe start a discussion about it.

packages/server/src/queue/queue.service.ts
```diff
@@ -22,10 +22,9 @@ import { AlertsService } from '../alerts/alerts.service';
  */
 @Injectable()
 export class QueueService {
-  constructor(
-    private connection: Connection,
-    private alertsService: AlertsService,
-  ) {}
+  readonly MAX_QUESTIONS_PER_QUEUE = 30;
```
AdamFipke, line 25: 30 will probably be too low for Scott's use case. Also, one of the things with my work is adjusting the system to allow students to have 2 questions (one "task/demo" question and one regular question). So, in a lab of maybe 30 people (I will double-check with Scott and Ramon on Wednesday if it can realistically be larger than this), each student can have 2 questions, so maybe a max of 60 could be better but that's also kinda crazy amount of questions. I'll see what Ramon and Scott say

packages/server/src/semester/semester.controller.ts
```diff
@@ -4,8 +4,12 @@ import { SemesterModel } from './semester.entity';
 
 @Controller('semesters')
 export class SemesterController {
+  readonly MAX_SEMESTERS = 40;
```
AdamFipke, line 7: Could this be exploitable?

Let's say you manage to get an admin role in an organization.
Theoretically, you could spam a ton of fake semesters and then professors would be stuck when they try to create a new course.

Though tbh that's kinda silly, I think I'm more worried about running out of semesters. If there's 4 semesters (one for each season), that means an organization can support 10 years of courses. However, more likely a bunch of profs will put in different stuff into the semester field (e.g. lowercase versions vs uppercase such as Fall and fall, Winter and winter, etc.). I'm willing to guess that profs are more likely to end up making 10+ semesters per year, meaning there's more like 4 years of semesters.

packages/server/src/course/course.controller.ts
d3li0n: For performance reasons, having 100 queues seems like overkill. I don't like the idea that each assignment should have its own queue instead of repurposing it. I will chat with Ramon about that

packages/server/src/semester/semester.controller.ts
d3li0n: We assume that there are no bad actors to cause "this issue"

packages/server/src/queue/queue.service.ts
d3li0n: Agree, lets chat with Ramon and Scott about this. Allowing more queues can simply crash frontend based on my tests. Unless we implement pagination

packages/server/src/semester/semester.controller.ts
AdamFipke: Well even if there are no bad actors, won't we eventually run out of semesters? 

Although, I guess one way to fix this is to make it so a limited number of semesters can be created each year, like maybe having MAX_SEMESTERS be a function of the number of years since the organization was created, allowing gradually more each year (maybe allow 8 on the first year, 16 on second year, etc.). 

Unless of course, you have other ideas on what you want to do with semester.

packages/server/src/queue/queue.service.ts
AdamFipke: Ah I see yeah, we'll probably need to implement pagination at some point... eugh.

packages/server/src/semester/semester.controller.ts
d3li0n: Ideally, we need admin to create semesters and set start and expiry date so that only active semester will show up

AdamFipke (approved):
new code changes look good. Didn't manually test them

AdamFipke (approved):

PR #28, opened 2024-06-20
Fix Windows yarn dev

AdamFipke:
# Description

Hopefully makes it so `yarn dev` works on windows while also not breaking mac/unix.

@d3li0n @wskksw please test on your end, i don't want this to break your environment

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build --webpack --webpackPath webpack.config.js",
-    "dev": "cross-env NODE_OPTIONS=--openssl-legacy-provider cross-env TYPEORM_LOGGING=true nest build --webpack --webpackPath webpack-hmr.config.js",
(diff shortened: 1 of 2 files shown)


packages/server/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build --webpack --webpackPath webpack.config.js",
-    "dev": "cross-env TYPEORM_LOGGING=true nest build --webpack --webpackPath webpack-hmr.config.js",
+    "dev": "cross-env NODE_OPTIONS=--openssl-legacy-provider cross-env TYPEORM_LOGGING=true nest build --webpack --webpackPath webpack-hmr.config.js",
```
d3li0n, line 10: Is it a typo? we are calling cross-env twice?

AdamFipke: ah you're right, didn't realise we can combine the two

d3li0n (approved):

PR #30, opened 2024-07-03
[Frontend] Tag Groups & useQuestionTypes hook

AdamFipke:
# Description

- Adds ability for both students and TAs to group the queue's questions together by tag/task.
- Also allows students to "join" a tag group by essentially creating a new question or editing their question with that tag/task
- Minor-ish change: added a useQuestionTypes hook to do a bunch of caching instead of calling getQuestionTypes all the time.

Depedent on #19 

image
image
image

## Type of change

# How Has This Been Tested?

This is basically all frontend, so just the usual manual testing of all new features and changes.

# Checklist:

- [ ] My changes generate no new warnings - NOTE: this actually generates several new non-breaking warnings in the console, but that's because of how antd's collapse menu is implemented. Nothing can be done to fix this, for now.
- [x] Any dependent changes have been merged and published in downstream modules


packages/common/index.ts
```diff
@@ -1136,26 +1097,15 @@ export class QuestionTypeParams {
...
-  name!: string
-
-  @IsString()
-  @IsOptional()
-  color?: string
-
-  @IsInt()
-  @IsOptional()
-  queueId?: number
+// named QuestionTypeType to not conflict with the UI component QuestionType
```
d3li0n, line 1100: I feel like we should rename this type, it's more confusing than it was before

packages/common/index.ts
```diff
@@ -1115,7 +1076,7 @@ export class UpdateQueueParams {
 }
 
 export class QuestionTypeParams {
-  @IsInt()
+  @IsInt() // when updating a question with new questionTypes, the question type's id is required
```
d3li0n, line 1079: If it's required, we need to remove IsOptional?

packages/app/components/Today/ChatbotToday.tsx
```diff
@@ -81,6 +81,9 @@ export const ChatbotToday: React.FC = () => {
           ])
         })
       })
+      .catch((err) => {
+        console.error(err)
```
d3li0n, line 85: probably better to `throw err`?

packages/app/components/Questions/Shared/SharedComponents.tsx
```diff
@@ -35,15 +35,15 @@ export const Photo = styled(Avatar)`
...
   @media (min-width: 1000px) {
-    margin: 0 32px;
+    margin: 0 2rem;
   }
 `
 export const QueueInfoColumnButtonStyle = styled(Button)<{
-  hasdemos?: boolean
-  isstudent?: boolean
+  hasdemos?: string
+  isstudent?: string
```
d3li0n, line 46: Variable names sound like boolean instead of string?

packages/app/components/Questions/Queue/JoinTagGroupButton.tsx
```diff
@@ -0,0 +1,220 @@
...
+      message.error(e.response?.data?.message || 'An unexpected error occurred')
+    }
+    setIsLoading(false)
+  }
+
+  return (
+    <CustomButton
+      size="small"
+      className={isDone ? '' : 'join-or-leave-tag-group-button'}
+      colorstyle={isDone ? 'Done' : isJoined ? 'Leave' : 'Join'}
```
d3li0n, line 198: Is there a better way to do this? I feel like nester ternary operators not great

packages/app/components/Questions/Queue/JoinTagGroupButton.tsx
```diff
@@ -0,0 +1,220 @@
...
+            aria-hidden="true"
+            style={{ transform: 'rotate(180deg)' }}
+            className="mr-1 shrink-0"
+          />
+        ) : (
+          <LogIn size={15} aria-hidden="true" className="mr-1 shrink-0" />
+        )
+      }
+    >
+      {isDone ? 'Done' : isJoined ? 'Leave' : 'Join'}
```
d3li0n, line 217: Repeats same condition as above

packages/app/components/Questions/Queue/JoinTagGroupButton.tsx
```diff
@@ -0,0 +1,220 @@
...
+              studentQuestion.text,
+              newQuestionTypes,
+              studentQuestion.groupable,
+              studentQuestion.isTaskQuestion,
+              studentQuestion.location,
+            )
+          }
+        } else if (taskId) {
+          if (!studentDemo) {
+            console.log('Error: Student does not have a task to leave')
```
d3li0n, line 166: Can we remvove any mentions of console log as it's mainly for development purposes

d3li0n (changes requested):
Left some comments

packages/app/components/Today/ChatbotToday.tsx
AdamFipke: If we don't catch the error, it causes that annoying uncaught error exception whenever we go on the Today page and we don't have chatbot running. 
Plus this is only for the suggested questions, which i don't think is a critical component that's worth informing us if it throws a 500 error.

packages/app/components/Questions/Shared/SharedComponents.tsx
AdamFipke: that's correct, though if i make them booleans, it gives me react warnings in the console that they can't be booleans n stuff.

packages/app/components/Questions/Queue/JoinTagGroupButton.tsx
AdamFipke: I don't think there's a better way of doing it. Also nested ternary operators are used in tons of places in the frontend (usually for whether or not to render certain components).

packages/app/components/Questions/Queue/JoinTagGroupButton.tsx
AdamFipke: This one is for displayed text, the other is for providing a string for specifying why styles to use.

packages/app/components/Questions/Queue/JoinTagGroupButton.tsx
AdamFipke: right, i will replace them with console.error. I feel like its important to log any error states when using the component since that means something is wrong.

packages/common/index.ts
AdamFipke: it's used in other places without id being required (such as when creating questionTypes). I put that comment there since I was analyzing some code and realised that the question type id was required for updating questions.
I mean the real solution is to just have different request types, but that was kinda out of what i wanted to do with this PR

packages/common/index.ts
AdamFipke: yeah it's not ideal for sure lol.
Ideally, we could rename all instances of questionType into questionTag or something like that (since `Type` gets confused with our types and sounds more mutually-exclusive than `tag`), but that's not worth the effort atm imo.

I mean it'd probably be better just to rename the type into QuestionType and then rename the component that displays a questionType, but meh.

d3li0n (approved):

PR #31, opened 2024-07-05
[UX] Allow TAs to mark individual tasks of a demo

AdamFipke:
# Description

Allows TAs individually mark specific tasks.

Dependent on #19 

Use case: 
1. Student makes demo asking for "task1" and "task2" to be marked
2. TA looks at their "task1" and thinks it looks good but "task2" is not done yet.
3. TA clicks/taps "task1" and clicks/taps "done"

In cases where all tasks are good, the TA may either click on each task that is good to go and click "done", or they may simply click "done" right away (as it was before). 

The tooltip of the "Done" button will show up for a couple seconds to show what pressing the button will do

image
image
image
image
Also adjusted the "Your Question has ended" text to display more helpful messages on what was marked.

## Type of change

# How Has This Been Tested?

Lots of manual testing.
New tests for question.integration.ts:
TaskQuestions: Allows TAs to edit a task question's text while it is being marked and only marks the new tasks
TaskQuestions: When TAs modify the questions text, it checks for valid tasks

# Checklist:

- [x] Any dependent changes have been merged and published in downstream modules


packages/app/styles/global.css
```diff
@@ -198,6 +198,35 @@ body {
     }
 }
 
+/* used to add a fancy highlight animation in TaskMarkingSelector */
+.highlight-spin:before {
+    content: '';
+    position: absolute;
+    left: -2.5px;
```
d3li0n, line 205: I feel like we stepped on the same issues with before with using `px` should we account for mobile version?

packages/app/components/Questions/Queue/TaskMarkingSelector.tsx
```diff
@@ -0,0 +1,158 @@
...
+  // for making it so you can press enter to toggle it
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key === 'Enter') {
+      handleClick()
+    }
+  }
+
+  // for applying hover and focus styles
+  const [isHovered, setIsHovered] = useState(false)
+  const handleMouseEnter = () => {
```
d3li0n, line 53: We have 4 methods that call same useState, should we reduce amount of methods and make it more general?

packages/app/components/Questions/Queue/TaskMarkingSelector.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+function CheckableMarkingTask({
+  taskId,
+  taskName,
+  taskColor,
+  onChange,
+  checked,
+  decorative,
+}: CheckableMarkingTaskProps): React.ReactElement {
+  function getBrightness(color: string): number {
```
d3li0n, line 27: Do we have this method already?

packages/app/components/Questions/Queue/TAQueueDetailButtons.tsx
```diff
@@ -161,16 +228,34 @@ export default function TAQueueDetailButtons({
             <CantFindButton shape="circle" icon={<CloseOutlined />} />
           </Tooltip>
         </Popconfirm>
-        <Tooltip title="Finish Helping">
+        <Tooltip
+          title={
+            question.isTaskQuestion
+              ? tasksSelectedForMarking.length > 0
```
d3li0n, line 234: Is there a way we can reduce complexity of this logic? More than condition seems excessive?

d3li0n (commented):
Left some comments

packages/app/components/Questions/Queue/TaskMarkingSelector.tsx
AdamFipke: it is already more general. If someone wanted to change a state differently for focus/blur compared to mouseEnter/mouseLeave, they can. Also, the function's names are accurate to what they are doing (handling an action).  It could be probably changed to 2 tiny functions instead of 4, but meh, I would argue it does not matter much in this case

packages/app/styles/global.css
AdamFipke: in this instance where it's just adding a small visual effect that extends just beyond the elements size, it's actually a better idea to use absolute units like `px` since otherwise it will increase in size with relative units. 

Also, I tested with it on mobile and also with it on 400% zoom, it looks fine.

packages/app/components/Questions/Queue/TaskMarkingSelector.tsx
AdamFipke: yeah i mean technically it's within QuestionType (which is what this file was originally based off of) and technically they can be refactored to both use the same function in a shared file but meh i don't think it's worth changing at this point. Maybe if we use getBrightness in a 3rd area, i'll refactor it

d3li0n (approved):

PR #62, opened 2024-08-30
fix: rate limit does not apply to professor

karel-harjono:
kevin pls approve :)

### DIFF
--- a/packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
@@ -475,7 +475,7 @@ const Chatbot: React.FC<ChatbotProps> = ({
                     Ask
                   </Button>
                 </Space.Compact>
-                {userInfo.chat_token && (
+                {userInfo.chat_token && questionsLeft < 100 && (
                   <Card.Meta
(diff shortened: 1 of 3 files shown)


wskksw:
Tests are failing now for the changed end point. Need to update test cases.

AdamFipke:
Also the populate chat token endpoint does something very different than resetting the token uses. The populate chat token looks to have been there to ensure that all users have a chat token associated with them (since i guess maybe they won't always), hence why this was an endpoint in developer settings

AdamFipke (changes requested):
wohoo I have internet again and can properly request changes now

wskksw (approved):
I made changes to revert the old endpoint

AdamFipke:
no tests? :((((

wskksw:
I will be writing some today with my other branch. I needed to quickly push these to production yesterday because class started today.

AdamFipke:
oh okay. While writing the tests, the following should also be changed:
- The backend endpoint is going to update for all users rather than just the users in the organisation. 
- Also you probably want a try catch there in case the query fails for some reason.
- Also this new endpoint you made can probably just set used to 0. If it's also resetting the max_uses, it should say as such (since not all users may have their max_uses set the same).

wskksw:
sounds good!

PR #125, opened 2024-09-21
Hopefully fix redirect to queue invite page after registering with google

AdamFipke:
# Description

The system won't redirect you to the queue invite page after registering with google for some reason. 

I am unsure if this actually fixes it, I am unable to test if google oath works on localhost


AdamFipke:
Since I can't seem to get google oath working on my dev environment, I may just need to have this pushed to prod to check to make sure it works

wskksw (approved):
I looked the problem up, yes we can merge it, but try it out and see whether it does fix the issue, otherwise revert

PR #131, opened 2024-09-22
add create subcriptions on registration and fix mail service template…

wskksw:
… bugs

# Description

user subscriptions are not created currently upon registration. That is fixed in this PR. 

Fixed bug of email not sending the extra info due to bad string concats. 

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/server/src/asyncQuestion/asyncQuestion.controller.ts
+++ b/packages/server/src/asyncQuestion/asyncQuestion.controller.ts
@@ -269,12 +269,6 @@ export class asyncQuestionController {
 
     const updatedQuestion = await question.save();
 
-    // Update Redis queue
-    await this.redisQueueService.updateAsyncQuestion(
-      `c:${question.course.id}:aq`,
-      updatedQuestion,
-    );
-
     if (body.status === asyncQuestionStatus.StudentDeleted) {
       await this.redisQueueService.deleteAsyncQuestion(
         `c:${question.course.id}:aq`,
--- a/packages/server/src/auth/auth.controller.ts
+++ b/packages/server/src/auth/auth.controller.ts
@@ -417,6 +417,8 @@ export class AuthController {
         sid ? sid : -1,
         organizationId,
       );
+      // create student subscriptions
+      await this.authService.createStudentSubscriptions(userId);
 
       const authToken = await this.createAuthToken(userId);
 
--- a/packages/server/src/auth/auth.service.ts
+++ b/packages/server/src/auth/auth.service.ts
@@ -1,6 +1,8 @@
 import { AccountType, OrganizationRole } from '@koh/common';
 import {
   BadRequestException,
+  HttpException,
+  HttpStatus,
   Injectable,
   InternalServerErrorException,
 } from '@nestjs/common';
@@ -45,7 +47,10 @@ export class AuthService {
 
       await UserSubscriptionModel.save(subscriptions);
     } catch (err) {
-      console.log(err);
+      throw new HttpException(
+        'There was a error saving user mail subscriptions',
+        HttpStatus.INTERNAL_SERVER_ERROR,
+      );
     }
   }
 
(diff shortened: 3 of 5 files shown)


packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -269,12 +269,6 @@ export class asyncQuestionController {
 
     const updatedQuestion = await question.save();
 
-    // Update Redis queue
-    await this.redisQueueService.updateAsyncQuestion(
-      `c:${question.course.id}:aq`,
-      updatedQuestion,
-    );
-
```
AdamFipke, line 277: erm, why are we deleting this? Won't this mean that when you update an async question, it won't display the update since redis won't know of the new question?

packages/server/src/auth/auth.controller.ts
```diff
@@ -417,6 +417,8 @@ export class AuthController {
         sid ? sid : -1,
         organizationId,
       );
+      // create student subscriptions
+      await this.authService.createStudentSubscriptions(userId);
```
AdamFipke, line 421: note that register() is not the only method where users are created. Login with sso as well as login with google will create a user account if they don't have one yet.

packages/server/src/auth/auth.service.ts
```diff
@@ -25,6 +27,22 @@ export class AuthService {
     );
   }
 
+  async createStudentSubscriptions(userId: number): Promise<void> {
+    const memberMailServices = await MailServiceModel.find({
+      where: { mailType: 'member' },
+    });
+
```
AdamFipke, line 34:
```suggestion
if (!memberMailServices) {
  return
}
```

packages/server/src/auth/auth.service.ts
```diff
@@ -25,6 +27,22 @@ export class AuthService {
...
+    const subscriptions = memberMailServices.map((service) => {
+      const subscription = new UserSubscriptionModel();
+      subscription.userId = userId;
+      subscription.serviceId = service.id;
+      subscription.isSubscribed = true;
+      return subscription;
+    });
+
+    await UserSubscriptionModel.save(subscriptions);
```
AdamFipke, line 43: might want to put a try-catch either around this .save or around anything that calls createStudentSubscription() (I suggest the latter since it catches anything with the .find as well)

packages/server/test/auth.integration.ts
```diff
@@ -46,6 +46,10 @@ const mockAuthService = {
     }
   },
 
+  createStudentSubscriptions: async (userId: number) => {
+    return;
+  },
+
```
AdamFipke, line 52: Just a general question, why is this here?

AdamFipke (changes requested):
see comments

packages/server/src/auth/auth.service.ts
```diff
@@ -28,21 +28,25 @@ export class AuthService {
...
+        subscription.serviceId = service.id;
+        subscription.isSubscribed = true;
+        return subscription;
+      });
 
-    await UserSubscriptionModel.save(subscriptions);
+      await UserSubscriptionModel.save(subscriptions);
+    } catch (err) {
+      console.log(err);
+    }
```
AdamFipke, line 49: This will not inform the user that any error has happened. I was thinking more like a try catch around any call to createStudentSubscriptions that will then return something like "There was a error saving user mail subscriptions"

AdamFipke: Something like this:
 ```ts
 try {
      this.createStudentSubscription()
    } catch (err) {
      throw new HttpException(
        'There was a error saving user mail subscriptions',
        HttpStatus.INTERNAL_SERVER_ERROR,
      );
    }
```

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
AdamFipke: okay turns out this is duplicate logic below

AdamFipke (approved):

Issue #95, opened 2024-09-05
get rid of useSWR on the profile page


AdamFipke:
**Description**

useSWR was originally added there as part of making sure the profile data is up-to-date (and i guess dima couldn't figure out how to update the userInfo context). I managed to properly update the userInfo context and now useSWR can be safely removed and replaced with a getUser here and there.

Using useSWR is best if you use it everywhere (since I think there's a decent bit of initial setup it has to do behind the scenes), and really probably isn't that necessary in the profile page.

PR #136, opened 2024-09-24
Image Cropping for Org. Banners and Logos and User Avatars

### LINKED ISSUE #90: Image compression + Cropper
**Description**

If you try to take a photo of yourself to upload to helpme, it will tell you that the image is too large and the max size is 1MB.

We should be using a compression library for photo uploads
- Adding a compression library can also help with performance, since there could be maybe 50 profile pictures on the page at the same time. And 1MB is really overkill for these photos too since most are just these tiny 48x48 pictures (100x100 at most). Heck, you could probably specifiy in the compression library to compress it to only 100x100 (though maybe do 200x200 just to be safe, since the profile pictures can get a little larger on the profile page on mobile)
	- Also in the options for the compression library, we will have the option between auto-cropping the images or to squish the images so that they are the correct aspect ratio. Maybe there is a larger discussion to be had here on what to pick. Ideally we would incorperate one of those interactive spotlight croper things that let you select what area of the photo becomes the profile picture. I don't believe antd has one, so you may have to go searching. But this may take additional effort that might be better for a seperate PR after getting the compression working.
	- And if you can, try to put the compression on the frontend to save some backend resources :) (especially since uploading a huge image just to get compressed is a bad idear)

- Also, the image compressor may also be a good thing to add for the organization banner + logo, though this may be less priority since only us developers have been adding that so far

An example of the "interactive splotlight croper things" is this (from google's profile picture selector thing)
image

### LINKED ISSUE #93: Support for more image types for pfp, org logo, and org banner
**Description**

Our image inputs only support png and jpg iirc. This is an somewhat-unnecessary restriction that may cause unnecessary issues for users. There are plenty of image types out there besides that (webp, apng (same as png), jpeg (same as jpg), bmp, svg, gif, maybe some others).

https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Image_format_support

There may be some configuration or conversion required with the compression library #90

patrickma6199:
# Description

This PR tackles two issues (linked to this PR below). The first was to add a spotlight cropper to allow primary users to be able to resize their profile avatars and organization banners and logos. The second was to allow for more image file types when uploading these images and to increase the image file size limit from 1MB to 5MB. Additionally, I also converted each image into WEBP before storage in the uploads directory to reduce space usage, and speed up image retrieval. This is to generally improve the user experience and lessen users' potential reliance on other software to compress or resize these images.

Outside of issue #90 and #93, I also changed the file name generation of images saved in the uploads directory from random string generation to being reliant on the integer representation of date stamps. This greatly enforces that each filename will be unique. I also tackled a little of #95 but have not removed the calls to useSWR() in the notifications section of the profile settings page as that was a lot more challenging.

Dependencies:
- sharp (node module for the backend to replace JIMP) - handles lightweight image resizing and file type conversion to WEBP
- react-easy-crop (node module) - a premade react component for lightweight image spotlight cropping

Closes #90
Closes #93 

## Type of change

# How Has This Been Tested?

- [x] Manually tested all added functionality
  - [x] Test that uploading a profile/logo/banner image results in:
    - [x] A success message
    - [x] The image file in the uploads directory
    - [x] The image updating in the UI without a refresh
  - [x] All newly uploaded image files are WEBP files
  - [x] Naming conventions of each file end with an integer representation of the date of upload
  - [x] Upload size limit is universally 5MB
  - [x] Uploads support these file types: JPG, PNG, WEBP, TIFF, AVIF, GIF, SVG

# Checklist:


packages/frontend/app/(dashboard)/components/ImageCropperModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+    'image/gif',
+    'image/svg+xml',
+    'image/tiff',
+  ]
+
+  const beforeUpload = (file: any): boolean => {
+    const isValidMimeType = viableFileTypes.includes(file.type)
+
+    if (!isValidMimeType) {
+      message.error('You can only upload JPGs or PNGs!')
```
AdamFipke, line 74:
```suggestion
      message.error('Only the following image formats are supported: jpeg/jpg, png, webp, avif, gif, svg, and tiff')
```

packages/frontend/app/(dashboard)/components/ImageCropperModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+        className="mb-2"
+        showUploadList={false}
+        maxCount={1}
+      >
+        <button className="min-w-[85vw] rounded-lg border-2 bg-white p-2 lg:min-w-[30vw]">
+          <UploadOutlined />
+          <span className="ml-2">{`Upload ${imgName}`}</span>
+        </button>
+      </Upload>
+      <div className="relative mt-2 h-[300px] w-full">
```
AdamFipke, line 149:
```suggestion
      <div className="relative mt-2 h-80 w-full">
```
I suggest using relative units where possible. Though if this is a component should always be 300px than you can disregard this comment.

packages/frontend/app/(dashboard)/profile/components/AvatarSettings.tsx
```diff
@@ -15,59 +15,14 @@ const AvatarSettings: React.FC = () => {
...
-
-    const isLt1M = file.size / 1024 / 1024 < 1
-    if (!isLt1M) {
-      message.error('Image must smaller than 1MB!')
-    }
-
-    return isJpgOrPng && isLt1M
-  }
-
-  const handleUpload = async (file: any) => {
```
AdamFipke, line 44: i like that you standardized the photo upload

packages/frontend/app/(dashboard)/profile/components/EditProfile.tsx
```diff
@@ -62,8 +60,6 @@ const EditProfile: React.FC = () => {
         })
     }
 
-    const newUser = await mutate() //update the context
-    setUserInfo({ ...userInfo, ...newUser })
```
AdamFipke, line 66: You might not want to delete this setUserInfo since you want to update the global userInfo context to the new information. I would probably just do something like setUserInfo(newProfile) or something. Make sure to manually test it.

packages/frontend/app/components/UserAvatar.tsx
```diff
@@ -41,9 +41,16 @@ export default function UserAvatar({
...
+        photoURL && photoURL.startsWith('http') ? (
+          <img src={photoURL} alt={username} loading="lazy" decoding="async" />
+        ) : (
+          <img
+            src={`/api/v1/profile/get_picture/${photoURL}`}
+            alt={username}
+            loading="lazy"
+            decoding="async"
+          />
+        )
```
AdamFipke, line 53: huh i didn't know you could provide an img element. What if we used Next.js's Image tag? Since I believe it does some fancy caching things that could be nice maybe.

packages/server/src/organization/organization.controller.ts
```diff
@@ -808,18 +809,21 @@ export class OrganizationController {
...
-    organization.bannerUrl = fileName;
+    // Create the upload location if it doesn't exist
+    if (!fs.existsSync(process.env.UPLOAD_LOCATION)) {
+      fs.mkdirSync(process.env.UPLOAD_LOCATION, { recursive: true });
+    }
+
+    const targetPath = path.join(process.env.UPLOAD_LOCATION, fileName);
+
+    try {
+      await sharp(file.buffer).resize(1920, 1080).webp().toFile(targetPath);
```
AdamFipke, line 822: I think 1920x1080 might be larger than the org banner needs to be. Since the org banner's height takes up 20% of the screen, I would make the height something like 200 or 300px.

packages/server/src/organization/organization.controller.ts
```diff
@@ -889,19 +893,21 @@ export class OrganizationController {
...
-    const image = await Jimp.read(file.buffer); // Load the image
-    image.resize(256, Jimp.AUTO); // Resize the image to 256 pixels (width, maintaining aspect ratio)
-    await image.writeAsync(
-      path.join(process.env.UPLOAD_LOCATION as string, fileName),
-    ); //same old
+    const targetPath = path.join(process.env.UPLOAD_LOCATION, fileName);
 
-    organization.logoUrl = fileName;
+    try {
+      await sharp(file.buffer).resize(256).webp().toFile(targetPath);
```
AdamFipke, line 906: I realise that previously we also resized to 256x256, but I think we can resize it to 48x48 since that's the size of it in the image elements

packages/server/src/profile/profile.controller.ts
```diff
@@ -218,27 +222,25 @@ export class ProfileController {
...
         fs.mkdirSync(process.env.UPLOAD_LOCATION, { recursive: true });
       }
 
       const targetPath = path.join(process.env.UPLOAD_LOCATION, fileName);
 
       try {
-        const image = await Jimp.read(file.buffer);
-        image.resize(256, Jimp.AUTO);
-        await image.writeAsync(targetPath);
+        await sharp(file.buffer).resize(256).webp().toFile(targetPath);
```
AdamFipke, line 235: Maybe let's try resizing it to 128. This may make it a little blurry on the profile page on mobile but otherwise I think should be fine

AdamFipke (changes requested):
I left some comments. Overall pretty solid job!

There is one thing though, I was wondering if we could make the banner cropper a super long rectangle (something like 1920x300) since that's roughly the size of the banner (roughly)

packages/frontend/app/(dashboard)/profile/components/AvatarSettings.tsx
patrickma6199: Thanks!

packages/server/src/profile/profile.controller.ts
patrickma6199: Screenshot 2024-10-02 111727
Screenshot 2024-10-02 111756

I think that it might be an unnecessary optimization that would hinder the visual experience of the application. I think resizing the organization logo makes more sense as there are no instances where it is displayed as any more detailed than 48x48, but that's not the case with the profile photos.

patrickma6199: The experience of changing profile photos is especially affected because the user goes from a high-res version of their image in the cropper to an instantly low-res version on display within the profile page.

packages/server/src/organization/organization.controller.ts
patrickma6199: I have reduced this to 100x100 because 48x48 was still looking too obviously pixelated on my 1080p monitor.

packages/frontend/app/components/UserAvatar.tsx
patrickma6199: I tried it out and sadly, it messes with the API call for the image from the backend so I'll keep it as is for now.

AdamFipke (approved):
Sweet looks good! I also realised that the file upload was missing an accept attribute (so that when you go to select an image to upload, only the images will show up in your file explorer) so i just quickly added that

PR #147, opened 2024-09-30
Backups

### LINKED ISSUE #52: Add backup system
**Description**

Let's not lose our data forever guys

AdamFipke:
# Description

Adds a backup system with docs for restoring backups. NOTE: this works for both chatbot and helpme databases (i think, since they're both using postgres).

There are three types of backups:
- **Semi-Hourly**: Every 3 hours. These are rolling (we only keep 5 days of these). We also only take these from 7am to 10pm.
- **Daily**: Every day at midnight. These are rolling (we only keep 1 month of these)
- **Monthly**: The first of every month at midnight. These are kept indefinitely.

Closes #52 

## Type of change

# How Has This Been Tested?

I added a bunch of tests for it:

describe('handleDailyBackup'
it('should create a daily backup when there is sufficient disk space'
it('should not create a backup if there is insufficient disk space'
it('should delete old backups older than 30 days'
describe('handleSemiHourlyBackup'
it('should create a semi-hourly backup when there is sufficient disk space'
it('should not create a backup if there is insufficient disk space'
it('should delete old backups older than 5 days'
describe('handleMonthlyBackup'
it('should create a monthly backup when there is sufficient disk space'
it('should not create a backup if there is insufficient disk space'
describe('checkDiskSpace'
it('should return true if there is enough disk space'
it('should return false if there is not enough disk space'

I have also tested making the interval 1minute, and then restoring one of those backups, which works fine.

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ yarn-error.log*
 husky.sh
 
 .vscode/
+
+# ignore backups 
+/backups/**/*.gz
--- a/backups/daily/daily-backups.md
+++ b/backups/daily/daily-backups.md
@@ -1,0 +1,1 @@
+see [Backups](../../docs/backups.md) 
--- a/backups/monthly/monthly-backups.md
+++ b/backups/monthly/monthly-backups.md
@@ -1,0 +1,1 @@
+see [Backups](../../docs/backups.md) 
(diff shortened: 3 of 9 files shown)


AdamFipke:
Note that ramon suggested storing the backup logs in the database. I don't think that is necessary store the log files anywhere. I'm hoping sentry can catch any failed backup errors. If not, I'll set up the email service and probably just write an error log. For now this is fine

wskksw (approved):

PR #148, opened 2024-10-02
Queue Location Settings (Online, Hybrid, In-Person)

patrickma6199:
# Description

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context. 
List any dependencies that are required for this change.

Closes # (issue number)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -1217,6 +1217,10 @@ export class TACheckoutResponse {
 export class UpdateQueueParams {
   @IsString()
   @IsOptional()
+  type?: QueueTypes
+
+  @IsString()
+  @IsOptional()
   notes?: string
 
   @IsBoolean()
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -312,6 +312,8 @@ export interface Queue {
  * @param isOpen - A queue is open if it has staff and is not disabled.
  * @param config - A JSON object that contains the configuration for the queue. Contains stuff like tags, tasks, etc.
  */
+export type QueueTypes = 'online' | 'hybrid' | 'inPerson'
+
 export class QueuePartial {
   id!: number
   room!: string
@@ -332,6 +334,8 @@ export class QueuePartial {
   endTime?: Date
 
   allowQuestions!: boolean
+
+  type?: QueueTypes
 
   isProfessorQueue!: boolean
 
--- a/packages/frontend/app/(dashboard)/course/[cid]/components/CreateQueueModal.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/components/CreateQueueModal.tsx
@@ -1,5 +1,5 @@
 import { QuestionCircleOutlined } from '@ant-design/icons'
-import { Role, validateQueueConfigInput } from '@koh/common'
+import { QueueTypes, Role, validateQueueConfigInput } from '@koh/common'
 import {
   Modal,
   Form,
@@ -9,6 +9,7 @@ import {
   Tooltip,
   Button,
   message,
+  Segmented,
 } from 'antd'
 import TextArea from 'antd/lib/input/TextArea'
 import { useState } from 'react'
@@ -23,7 +24,7 @@ type ConfigPresets = 'default' | 'lab'
 
 interface FormValues {
   officeHourName: string
-  isOnline: boolean
+  type: QueueTypes
   allowTA: boolean
   notes: string
   queue_config: string
@@ -49,12 +50,19 @@ const CreateQueueModal: React.FC<CreateQueueModalProps> = ({
   const [isLoading, setIsLoading] = useState(false)
   const [hoveredConfig, setHoveredConfig] = useState<ConfigPresets>('default')
 
+  const queueTypeOptions = [
+    { label: 'Online', value: 'online' },
+    { label: 'Hybrid', value: 'hybrid' },
+    { label: 'In-Person', value: 'inPerson' },
+  ]
+
   const onFinish = async (values: FormValues) => {
     setIsLoading(true)
     await API.queues
       .createQueue(
         cid,
         values.officeHourName,
+        values.type,
         !values.allowTA,
         values.notes ?? '',
         values.queue_config ? JSON.parse(values.queue_config) : {},
@@ -114,7 +122,7 @@ const CreateQueueModal: React.FC<CreateQueueModalProps> = ({
           form={form}
           name="form_in_modal"
           initialValues={{
-            isOnline: false,
+            type: 'hybrid',
             allowTA: true,
           }}
           clearOnDestroy
@@ -125,13 +133,12 @@ const CreateQueueModal: React.FC<CreateQueueModalProps> = ({
       )}
     >
       <Form.Item
-        label="Online?"
-        name="isOnline"
+        label="Queue Location"
+        name="type"
         layout="horizontal"
-        valuePropName="checked"
-        tooltip="Online queues have the option for a zoom/teams link"
+        tooltip="Online queues are for virtual office hours. In-person queues are for physical locations. Hybrid queues are for a mix of both."
       >
-        <Switch />
+        <Segmented options={queueTypeOptions} />
       </Form.Item>
 
       <Form.Item
--- a/packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
@@ -9,28 +9,43 @@ import { Button, Card, Divider, Input, message, Row, Tag, Tooltip } from 'antd'
 import Link from 'next/link'
 import { ReactElement, useState } from 'react'
 import UserAvatar from '@/app/components/UserAvatar'
-import { QueuePartial } from '@koh/common'
+import { QueuePartial, QueueTypes } from '@koh/common'
 import { useCourse } from '@/app/hooks/useCourse'
 import { API } from '@/app/api'
 import { cn, getErrorMessage } from '@/app/utils/generalUtils'
 
 interface QueueCardProps {
   cid: number
+  type: QueueTypes
   queue: QueuePartial
-  isTA: boolean
+  isStaff: boolean
   linkId: string
 }
 
 const QueueCard: React.FC<QueueCardProps> = ({
   cid,
+  type,
   queue,
-  isTA,
+  isStaff,
   linkId,
 }): ReactElement => {
   const { mutateCourse } = useCourse(cid)
   const [editingNotes, setEditingNotes] = useState(false)
   const [updatedNotes, setUpdatedNotes] = useState(queue.notes)
   const [isLinkEnabled, setIsLinkEnabled] = useState(true) // for enabling/disabling the link to the queue when editing notes
+
+  const getQueueTypeLabel = (type: QueueTypes) => {
+    switch (type) {
+      case 'online':
+        return 'Online'
+      case 'hybrid':
+        return 'Hybrid'
+      case 'inPerson':
+        return 'In-Person'
+      default:
+        return 'Invalid Queue Type'
+    }
+  }
 
   const handleSaveQueueNotes = async (e: { preventDefault: () => void }) => {
     e.preventDefault()
@@ -81,6 +96,14 @@ const QueueCard: React.FC<QueueCardProps> = ({
             <div>
               {queue.room}
               <div className="flex">
+                {queue?.type && (
+                  <Tag
+                    color="#2a9187"
+                    className="m-0 mr-1 leading-4 text-gray-200"
+                  >
+                    {getQueueTypeLabel(queue.type)}
+                  </Tag>
+                )}
                 {queue?.isProfessorQueue && (
                   <Tag
                     color="#337589"
@@ -149,7 +172,7 @@ const QueueCard: React.FC<QueueCardProps> = ({
             <div className="whitespace-pre-wrap break-words text-[rgb(125,125,125)]">
               <NotificationOutlined /> <i>{queue.notes}</i>
             </div>
-          ) : isTA ? (
+          ) : isStaff ? (
             <i className="font-light text-gray-400"> no notes provided </i>
           ) : null}
           <div className="flex">
@@ -164,7 +187,7 @@ const QueueCard: React.FC<QueueCardProps> = ({
             )}
             {!editingNotes && (
               <Row className="pt-2.5">
-                {isTA && (
+                {isStaff && (
                   <Button
                     className="rounded-md border px-4 py-2 text-base font-medium"
                     onClick={(e) => {
--- a/packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx
@@ -15,6 +15,7 @@ import {
   ColorPickerProps,
   GetProp,
   Dropdown,
+  Segmented,
 } from 'antd'
 import {
   QuestionTypeParams,
@@ -123,6 +124,13 @@ const EditQueueModal: React.FC<EditQueueModalProps> = ({
     }
   }, [open, queue?.config, resetQueueConfig])
 
+  // Values and labels for Segmented component for queue types
+  const queueTypeOptions = [
+    { label: 'Online', value: 'online' },
+    { label: 'Hybrid', value: 'hybrid' },
+    { label: 'In-Person', value: 'inPerson' },
+  ]
+
   const onFinish = async (values: FormValues) => {
     let errorsHaveOccurred = false
     const deletePromises =
@@ -178,10 +186,12 @@ const EditQueueModal: React.FC<EditQueueModalProps> = ({
             })
         : Promise.resolve()
     const updateQueueParams: UpdateQueueParams = pick(values, [
+      'type',
       'notes',
       'allowQuestions',
     ])
     const updateQueuePromise =
+      updateQueueParams.type !== queue?.type ||
       updateQueueParams.notes !== queue?.notes ||
       updateQueueParams.allowQuestions !== queue?.allowQuestions
         ? API.queues
@@ -241,6 +251,10 @@ const EditQueueModal: React.FC<EditQueueModalProps> = ({
         </Form>
       )}
     >
+      <Form.Item label="Queue Type" name="type">
+        <Segmented options={queueTypeOptions} />
+      </Form.Item>
+
       <Form.Item label="Queue Notes" name="notes">
         <TextArea
           className="rounded-md border border-gray-400 font-normal"
--- a/packages/frontend/app/(dashboard)/course/[cid]/page.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/page.tsx
@@ -119,6 +119,7 @@ export default function CoursePage({ params }: CoursePageProps): ReactElement {
                   sortedQueues?.map((q) => (
                     <QueueCard
                       cid={cid}
+                      type={q.type!}
                       linkId={
                         skipLinkTarget == 'first-queue' &&
                         q.id === sortedQueues[0].id
@@ -127,7 +128,7 @@ export default function CoursePage({ params }: CoursePageProps): ReactElement {
                       }
                       key={q.id}
                       queue={q}
-                      isTA={role === Role.TA || role === Role.PROFESSOR}
+                      isStaff={role === Role.TA || role === Role.PROFESSOR}
                     />
                   ))}
 
--- a/packages/frontend/app/api/index.ts
+++ b/packages/frontend/app/api/index.ts
@@ -22,6 +22,7 @@ import {
   UpdateQuestionParams,
   UpdateQuestionResponse,
   UpdateQueueParams,
+  QueueTypes,
   QueuePartial,
   Role,
   GetCourseUserInfoResponse,
@@ -445,6 +446,7 @@ class APIClient {
     createQueue: async (
       courseId: number,
       room: string,
+      type: QueueTypes,
       isProfessorQueue: boolean,
       notes: string,
       config: QueueConfig,
@@ -453,7 +455,7 @@ class APIClient {
         'POST',
         `/api/v1/courses/${courseId}/create_queue/${room}`,
         QueuePartial,
-        { notes, isProfessorQueue, config },
+        { notes, type, isProfessorQueue, config },
       ),
   }
 
--- a/packages/server/migration/1727805441958-AddQueueLocationTypesColumn.ts
+++ b/packages/server/migration/1727805441958-AddQueueLocationTypesColumn.ts
@@ -1,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddQueueLocationTypesColumn1727805441958
+  implements MigrationInterface
+{
+  name = 'AddQueueLocationTypesColumn1727805441958';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "queue_model" ADD "type" character varying NOT NULL DEFAULT 'hybrid'`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "queue_model" DROP COLUMN "type"`);
+  }
+}
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -11,6 +11,7 @@ import {
   QueueConfig,
   QueueInvite,
   QueuePartial,
+  QueueTypes,
   Role,
   TACheckinTimesResponse,
   TACheckoutResponse,
@@ -521,6 +522,7 @@ export class CourseController {
     @Body()
     body: {
       notes: string;
+      type: QueueTypes;
       isProfessorQueue: boolean;
       config: QueueConfig;
     },
@@ -586,6 +588,7 @@ export class CourseController {
             .create(QueueModel, {
               room,
               courseId,
+              type: body.type,
               staffList: [],
               questions: [],
               allowQuestions: true,
--- a/packages/server/src/queue/queue.controller.ts
+++ b/packages/server/src/queue/queue.controller.ts
@@ -128,6 +128,7 @@ export class QueueController {
     if (queue === undefined) {
       throw new NotFoundException();
     }
+    queue.type = body.type;
     queue.notes = body.notes;
     queue.allowQuestions = body.allowQuestions;
     try {
--- a/packages/server/src/queue/queue.entity.ts
+++ b/packages/server/src/queue/queue.entity.ts
@@ -20,6 +20,7 @@ import {
   ERROR_MESSAGES,
   OpenQuestionStatus,
   QueueConfig,
+  QueueTypes,
   StatusInQueue,
 } from '@koh/common';
 import { QuestionTypeModel } from '../questionType/question-type.entity';
@@ -58,6 +59,9 @@ export class QueueModel extends BaseEntity {
 
   @Column({ default: false })
   allowQuestions: boolean;
+
+  @Column({ default: 'hybrid' })
+  type: QueueTypes;
 
   @Column({ default: false })
   isProfessorQueue: boolean;
--- a/yarn.lock
+++ b/yarn.lock
@@ -75,10 +75,19 @@
   dependencies:
     "@ctrl/tinycolor" "^3.6.1"
 
-"@ant-design/cssinjs@^1.21.0":
-  version "1.21.0"
-  resolved "https://registry.yarnpkg.com/@ant-design/cssinjs/-/cssinjs-1.21.0.tgz#de7289bfd71c7a494a28b96569ad88f999619105"
-  integrity sha512-gIilraPl+9EoKdYxnupxjHB/Q6IHNRjEXszKbDxZdsgv4sAZ9pjkCq8yanDWNvyfjp4leir2OVAJm0vxwKK8YA==
+"@ant-design/cssinjs-utils@^1.1.0":
+  version "1.1.0"
+  resolved "https://registry.yarnpkg.com/@ant-design/cssinjs-utils/-/cssinjs-utils-1.1.0.tgz#5e85c16c1695e692f550b3e4b156cf70cac0b873"
+  integrity sha512-E9nOWObXx7Dy7hdyuYlOFaer/LtPO7oyZVxZphh0CYEslr5EmhJPM3WI0Q2RBHRtYg6dSNqeSK73kvZjPN3IMQ==
+  dependencies:
+    "@ant-design/cssinjs" "^1.21.0"
+    "@babel/runtime" "^7.23.2"
+    rc-util "^5.38.0"
+
+"@ant-design/cssinjs@^1.21.0", "@ant-design/cssinjs@^1.21.1":
+  version "1.21.1"
+  resolved "https://registry.yarnpkg.com/@ant-design/cssinjs/-/cssinjs-1.21.1.tgz#7320813c5f747e0cde52c388eff5198d78d57230"
+  integrity sha512-tyWnlK+XH7Bumd0byfbCiZNK43HEubMoCcu9VxwsAwiHdHTgWa+tMN0/yvxa+e8EzuFP1WdUNNPclRpVtD33lg==
   dependencies:
     "@babel/runtime" "^7.11.1"
     "@emotion/hash" "^0.8.0"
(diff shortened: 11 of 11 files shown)


packages/common/index.ts
```diff
@@ -312,6 +312,8 @@ export interface Queue {
  * @param isOpen - A queue is open if it has staff and is not disabled.
  * @param config - A JSON object that contains the configuration for the queue. Contains stuff like tags, tasks, etc.
  */
+export type QueueTypes = 'online' | 'hybrid' | 'inPerson'
```
AdamFipke, line 315: perhaps a better name for this is QueueLocations?

Also, you might want to move this since it's right below QueuePartial's jsdoc

packages/common/index.ts
```diff
@@ -1212,12 +1221,20 @@ export class TACheckoutResponse {
 }
 
 export class UpdateQueueParams {
+  @IsString()
+  @IsOptional()
+  type?: QueueTypes
```
AdamFipke, line 1226: and i guess you would want to rename this as well

packages/frontend/app/(dashboard)/course/[cid]/components/CreateQueueModal.tsx
```diff
@@ -125,13 +133,12 @@ const CreateQueueModal: React.FC<CreateQueueModalProps> = ({
...
     >
       <Form.Item
-        label="Online?"
-        name="isOnline"
+        label="Queue Location"
+        name="type"
         layout="horizontal"
-        valuePropName="checked"
-        tooltip="Online queues have the option for a zoom/teams link"
+        tooltip="Online queues are for virtual office hours. In-person queues are for physical locations. Hybrid queues are for a mix of both."
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/CreateQueueModal.tsx:139]:
```suggestion
        tooltip="Online and Hybrid queues have an option for a zoom link, and students will be prompted to join this zoom link once you start helping them"
```

packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
```diff
@@ -80,7 +83,15 @@ const QueueCard: React.FC<QueueCardProps> = ({
...
-              <div className="flex">
+              <div className="mb-1 flex flex-wrap gap-y-1 sm:mb-0">
+                {queue?.type && (
+                  <Tag
+                    color="#2a9187"
+                    className="m-0 mr-1 leading-4 text-gray-200"
+                  >
+                    {getQueueTypeLabel(queue.type)}
+                  </Tag>
+                )}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:94]: I spent some more time thinking about this, and I'm actually unsure how much value there is in showing the queue location.

Like for my use case, my students don't really gain much by knowing its in-person, because I have already told them where the labs take place.
I imagine it's a similar case in other courses, since the prof normally tells them where it is.
The part that I'm having trouble visualising is if showing this helps students in the _Help Desk_ use case. And for that, maybe it would be best to discuss this on our next meeting.

I had a similar thought process a while back with whether to show the "Queue is up to date" component on mobile since I was unsure how much value it gave. I decided to keep it there though since:
- it gives reassurance to the user that what they are seeing is what the queue currently is
- it also tells the user that this page live-updates, so they aren't expected to refresh the page

So, maybe if we can go without displaying the queue location, we can probably revert how it looks on mobile to how it was before.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/JoinZoomButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Button, ButtonProps, Tooltip } from 'antd'
+import { PhoneOutlined } from '@ant-design/icons'
+import { PropsWithChildren } from 'react'
+
+interface JoinZoomButtonProps extends ButtonProps {
+  zoomLink?: string
+  onJoin?: () => void
+}
+
+const JoinZoomButton: React.FC<PropsWithChildren<JoinZoomButtonProps>> = ({
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/JoinZoomButton.tsx:11]: I noticed that there is also a JoinZoomButton in QueueInfoColumn. I feel one of these was meant to be deleted?

Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/JoinZoomButton.tsx:11]: No, that was intentional. I put it there because the pop up modal for joining a zoom call would close after you clicked the link. So if you accidentally lose the tab and need the link again, that button will remain in the info column until the student is no longer being helped.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx
```diff
@@ -59,141 +63,184 @@ const QuestionCard: React.FC<QuestionCardProps> = ({
...
-            <div className="text-sm font-medium text-green-700">
-              {servedTime}
-            </div>
+          {question.status === LimboQuestionStatus.ReQueueing && (
+            <Col flex="1 1">
+              <div className="text-md h-full italic text-gray-600">
+                In the process of requeuing...
+              </div>
+            </Col>
+          )}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx:204]: This *cannot* be put here, as it will cause question text to be truncated early and also cause a lot of ugly wrapping.
image
image

For a solution I have two ideas:
1. Floating text, similar to the "Currently Being Helped". I like this since this idea is already being used for showing question status, but I also don't like it since it gives these requeuing questions more attention than I would like (ideally, requeing questions get less emphasis compared to other questions, not more).
2. *Remove the TA buttons* since they are all disabled anyway, and put the text there instead

I think in option 2 may be better for TA on desktop, though it may look a little weird for TA on mobile. Though, the lack of buttons may also draw more attention to it as well.

For the student's side, I think the best option may be just to put it as the floating text, as much as it's not an ideal solution. This is simply because there is so little space on mobile.
image

For consistency, you could just do floating text for both TA and student side. (This might be the best option overall). This may require some more discussion on Tuesday

Also, I would change the text to just say "Requeing..." or possibly "Not Ready" (the latter may be more universally understandable)

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx
```diff
@@ -185,22 +208,16 @@ const QueueInfoColumn: React.FC<QueueInfoColumnProps> = ({
               />
             ) : null}
           </Row>
+          {queue?.notes && (
+            <div className="flex max-h-[200px] w-full items-center overflow-y-auto px-2 text-xl text-[#5f6b79] sm:hidden">
+              <Linkify>
+                <div className="min-w-0 whitespace-pre-wrap break-words text-sm italic md:text-base">
+                  {`Notes: ${queue?.notes}`}
+                </div>
+              </Linkify>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx:217]: The new notes location kinda felt weird at first but I think it makes a little more sense this way.

Only comment here is if I have no notes, it still says Notes:
image

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx
```diff
@@ -97,8 +112,8 @@ const QueueInfoColumn: React.FC<QueueInfoColumnProps> = ({
         {buttons}
       </div>
 
-      <div className="flex">
-        <h3 className="my-0 text-2xl font-semibold md:mt-10">Staff</h3>
+      <div className="mt-3 flex">
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx:115]:
```suggestion
      <div className="md:mt-3 flex">
```

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/StaffList.tsx
```diff
@@ -8,11 +8,12 @@ import { formatWaitTime } from '@/app/utils/timeFormatUtils'
 
 interface StaffListProps {
   queueId: number
+  mobile?: boolean
 }
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/StaffList.tsx:12]: this mobile prop seems unused

packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:94]: Oh i realised I put this comment on the QueueCard component instead of on the QueueInfoColumn whoops ;P

For the QueueCard component in particular, would it be possible to not show the queue location if all queue locations are the same? 

Since I imagine it may be redundant to know a queue is Online when a whole course is Online, or vise-versa with in-person. 

But in the case where there may be hybrid office hours and online labs or something, that info can be valuable (e.g. if students were told prior that their course has hybrid office hours and in-person labs, they can immediately make that connection when viewing the queue card to get more re-assurance that they're going to the right place). 

Though maybe this ties into the discussion above of how much *value* students get by displaying queue location. 

image

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+  onJoin,
+  setRequeuing,
+}) => {
+  return (
+    <Modal
+      open={open}
+      footer={[]}
+      style={{ top: '30%' }}
+      closable={false}
+      title={`${taName ?? 'A TA'} is Ready for You Now!`}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx:27]:
```suggestion
      title={`${taName ?? 'A TA'} is ready for you now!`}
```
imo I think sentence case reads a little better, maybe because it puts emphasis on the name, idk. Small thing
image
image

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type JoinZoomNowModalProps = {
+  open: boolean
+  zoomLink?: string
+  taName?: string
+  onJoin: () => void
+  setRequeuing: () => void
+}
+
+const JoinZoomNowModal: React.FC<JoinZoomNowModalProps> = ({
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx:14]: I noticed that if the queue doesn't have a zoomlink set, the only option that students have is to requeue themselves.

Instead of showing a disabled button, what if we just showed some text saying "A TA has begun helping you. Be respectful of the TA's time. Be prepared with your question! \nOnce finished, this popup will automatically close.".

ALSO, I just realised it might be a really good idea to include the queue notes inside this modal, since some profs may decide to put critical information inside of it (or in the case of the help desk, they put the zoom link inside it i guess).

Lastly, I realised that if a TA forgets to mark a question as done, the student is basically trapped in this modal, locking them from interacting with the rest of the site. Maybe after 10s or something, show a close button?

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <div className="mt-8 flex h-full w-full flex-col items-center gap-2">
+        <JoinZoomButton zoomLink={zoomLink} onJoin={onJoin}>
+          Join Meeting Now
+        </JoinZoomButton>
+        <Popconfirm
+          title="Are you sure you want to temporarily leave the queue?"
+          onConfirm={setRequeuing}
+        >
+          <ReQueuingButton />
+        </Popconfirm>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx:38]: Add a loading state to this pls (e.g. setRequeueing.finally(() => {setRequeueingLoading(false)} or somethin along those lines)

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx:27]: I also experimented with putting a green "!" icon here.
image

I think I like it, but maybe you want to play around with it? Code is here:
```suggestion
      title={<div className='flex items-center gap-x-2'>
        <InfoCircleFilled className='text-green-500 text-2xl' /> {`${taName ?? 'A TA'} is ready for you now!`}
      </div>}
```

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx
```diff
@@ -181,13 +189,19 @@ const AddStudentsToQueueModal: React.FC<AddStudentsToQueueModalProps> = ({
...
-          <Form.Item name="location" label="Is the question in-person?">
-            <Radio.Group className="mb-1">
-              <Radio value="In Person">Yes</Radio>
-              <Radio value="Online">No</Radio>
-            </Radio.Group>
-          </Form.Item>
+          {isQueueHybrid && (
+            <Form.Item
+              name="location"
+              label="How would you like them to meet you?"
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx:195]: nice touch <3

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx
```diff
@@ -1,8 +1,13 @@
 import { useCallback } from 'react'
 import Modal from 'antd/lib/modal/Modal'
-import { Form, message, Checkbox, Input, Radio, Select } from 'antd'
+import { Form, message, Checkbox, Input, Radio, Select, Segmented } from 'antd'
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx:3]:
```suggestion
import { Form, message, Checkbox, Input, Select, Segmented } from 'antd'
```

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx
```diff
@@ -1,8 +1,13 @@
...
 import { useEffect, useState } from 'react'
-import { OpenQuestionStatus, UserTiny } from '@koh/common'
+import {
+  OpenQuestionStatus,
+  QuestionLocations,
+  QueueTypes,
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx:8]:
```suggestion
```

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx
```diff
@@ -181,13 +189,19 @@ const AddStudentsToQueueModal: React.FC<AddStudentsToQueueModalProps> = ({
...
-          {/* TODO: change this to only be an option if the queue is hybrid. Strictly in-person or online queues should not have this option */}
-          <Form.Item name="location" label="Is the question in-person?">
-            <Radio.Group className="mb-1">
-              <Radio value="In Person">Yes</Radio>
-              <Radio value="Online">No</Radio>
-            </Radio.Group>
-          </Form.Item>
+          {isQueueHybrid && (
+            <Form.Item
+              name="location"
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx:194]: You will want to add the following to the <Form> element, otherwise location will be null if the user doesn't click anything
```
initialValues={{
            location: isQueueHybrid ? 'Online' : undefined,
          }}
```
image

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/CreateQuestionModal.tsx
```diff
@@ -202,13 +213,16 @@ const CreateQuestionModal: React.FC<CreateQuestionModalProps> = ({
...
 
-      {/* TODO: change this to only be an option if the queue is hybrid. Strictly in-person or online queues should not have this option */}
-      <Form.Item name="location" label="Are you joining the queue in-person?">
-        <Radio.Group className="mb-1">
-          <Radio value="In Person">Yes</Radio>
-          <Radio value="Online">No</Radio>
-        </Radio.Group>
-      </Form.Item>
+      {isQueueHybrid && (
+        <Form.Item name="location" label="Is the question in-person?">
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/CreateQuestionModal.tsx:217]: Similarly with AddStudentsToQueueModal, you will want to add the following to the `<Form>` element, otherwise location will be null if the user doesn't click anything
```
initialValues={{
            location: isQueueHybrid ? 'Online' : undefined,
          }}
```
image

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
```diff
@@ -587,6 +617,15 @@ export default function QueuePage({ params }: QueuePageProps): ReactElement {
...
           ) : (
             <>
+              {((queue.type === 'hybrid' && // Show the "Join Zoom" button if staff is ready and student already clicked on the modal
+                studentQuestion?.location === 'Online') ||
+                queue.type === 'online') &&
+                clickedZoomModal &&
+                studentQuestion?.status === OpenQuestionStatus.Helping && (
+                  <JoinZoomButton zoomLink={queue.zoomLink ?? course?.zoomLink}>
+                    Need The Link?
+                  </JoinZoomButton>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:627]: This is really smart that you added this, just a couple small things with it:

```suggestion
                  <JoinZoomButton zoomLink={queue.zoomLink ?? course?.zoomLink}>
                    Join Zoom
                  </JoinZoomButton>
```

Also, the join zoom button is slightly more bold than the other QueueInfoColumn buttons?
image

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx
```diff
@@ -296,12 +287,16 @@ const EditQueueModal: React.FC<EditQueueModalProps> = ({
             })
         : Promise.resolve()
     const updateQueueParams: UpdateQueueParams = pick(values, [
+      'type',
       'notes',
       'allowQuestions',
+      'zoomLink',
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx:293]: I'm noticing that I can't clear the zoomLink, it seems to still call the edit course endpoint?

image
image

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx
```diff
@@ -679,11 +680,25 @@ const EditQueueModal: React.FC<EditQueueModalProps> = ({
...
+            </Tooltip>
+          </div>
+        }
+        name="zoomLink"
+      >
         <Input
           allowClear={true}
           className="text-sky-800"
-          placeholder="[No Zoom/Teams link set]"
+          placeholder="[No queue-specific Zoom/Teams link set]"
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx:701]: I wonder if there is a good way of showing the course zoom link here. What if the placeholder was changed to be the course zoom link? And if there is no course zoom link, it just says [No Zoom/Teams link set] like before

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx:194]: also update with this, i realised I was on an old version, hence why it says [null] instead of [unselected]. It seems the issue is still there though

AdamFipke (changes requested):
See comments. Overall, pretty awesome job actually! I am surprised you got this much done in this much time. 

I did a bunch of testing and for the most part works pretty well. 

I also committed one change that makes all requeueing questions appear more transparent.

AdamFipke (changes requested):
See comments. Overall, pretty awesome job actually! I am surprised you got this much done in this much time. 

I did a bunch of testing and for the most part works pretty well. 

I also committed one change that makes all requeueing questions appear more transparent.

AdamFipke:
> See comments. Overall, pretty awesome job actually! I am surprised you got this much done in this much time.
> 
> I did a bunch of testing and for the most part works pretty well.
> 
> I also committed one change that makes all requeueing questions appear more transparent.

And this is a perfect example why we add loading states to our buttons everyone lol

I clicked twice and it submitted two reviews lol

packages/common/index.ts
patrickma6199: Whoops, didn't notice the jsdoc hahah.

packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:94]: I see your point that if all the queues for a course are the same type, then this would likely not be necessary. Still, I think for consistency in general styling, there will be instances where some queues are of different types, and I don't think it "costs" the users (or us) much of anything to have more feedback about the general information of queues. The cards and general interface are not losing anything in terms of aesthetics or usability to have this extra label on the QueueCards or to have the locations in the QueueInfoColumn for desktop and mobile in my opinion.

I do also think generally that the revamped mobile interface for the queue page is more readable and easier on the eyes. Before my refactor, the flex-row with the "Queue up to date" message and queue group switch were squished together on the left with the notes oddly above the StaffList component. I will likely just run through these design choices in the meeting before merging to see what everyone thinks, but for now, I guess I'm just a little lost as to what we lose for having more feedback and assurance for users and keeping the styling of queue cards uniform in all instances for us to have an easier time working on top of this implementation in the future.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx:211]: Yea, when I added this for feedback, I didn't like it either but just didn't think of a good alternative. I agree that your latter suggestion makes more sense. Thanks for the idea! I'll add this in.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx:217]: Yea, I need a stricter check for not just whether its null, but whether it is falsy since and empty string is also falsy in JS unless I'm remembering that wrong. Thanks for pointing this out.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/AddStudentsToQueueModal.tsx:195]: thanks! hahaha

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/CreateQuestionModal.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/CreateQuestionModal.tsx:217]: I missed this, thanks!

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx:701]: Thats a good idea. I'll switch this over to that instead.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/StaffList.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/StaffList.tsx:12]: Nice catch, probably something I was doing with the styling before I abandoned it and forgot to clean up.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx:14]: I will change the disabled button to text and add the notes in as well. As for the student being trapped, if they click the requeue button or the join button, the modal automatically closes (the reason i added a button to get the zoom link on the infocolumn). I agree though that in the cases where the student wants to join online but the course doesn't have a zoom link, having the cancel button might still be useful since it makes no sense for the only option to be to requeue so ill add it in just to be safe.

Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx:14]: Here's the final styling here:

image

packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:94]: I think adhering to visibility of the system status by having the tags on each queue is helpful; if they are all the same, removing the label it's going to lose all information viability. In this case I think more information is as helpful as anything

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/JoinZoomButton.tsx
```diff
@@ -0,0 +1,51 @@
+import { cn } from '@/app/utils/generalUtils'
+import { Button, ButtonProps, Tooltip } from 'antd'
+import { PhoneOutlined } from '@ant-design/icons'
+import { PropsWithChildren } from 'react'
+import { Mr_Dafoe } from 'next/font/google'
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/JoinZoomButton.tsx:5]: importing Mr_Dafoe himself haha

packages/frontend/app/(dashboard)/components/EditCourseForm.tsx
```diff
@@ -37,6 +37,7 @@ const EditCourseForm: React.FC<EditCourseFormProps> = ({
     const courseNameField = formValues.courseName
     const coordinatorEmailField = formValues.coordinatorEmail
     const sectionGroupNameField = formValues.sectionGroupName
+    console.log('zoomLink: ' + formValues.zoomLink) // PAT TODO: remove
```
AdamFipke, line 40: hi

packages/frontend/app/(dashboard)/components/EditCourseForm.tsx
```diff
@@ -192,7 +188,7 @@ const EditCourseForm: React.FC<EditCourseFormProps> = ({
         <Form.Item
           label="Zoom Link"
           name="zoomLink"
-          tooltip="Link to the zoom meeting for queues. Currently, this is shared between all queues. When a student is helped, they will have the option to click this link."
+          tooltip="Default link to the zoom meeting for queues. Each queue can also have a unique zoom link which will automatically overwrite this one. When a student is helped, they will have the option to click this link."
```
AdamFipke, line 191: might want to also make this change on the organization/course/add/page.tsx

packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
```diff
@@ -80,7 +83,15 @@ const QueueCard: React.FC<QueueCardProps> = ({
           <span className="mr-8 flex flex-row flex-wrap items-center justify-between">
             <div>
               {queue.room}
-              <div className="flex">
+              <div className="mb-1 flex flex-wrap gap-y-1 sm:mb-0">
+                {queue?.type && (
+                  <Tag
+                    color="#2a9187"
+                    className="m-0 mr-1 leading-4 text-gray-200"
+                  >
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:91]: image
I noticed the tags' colors are the same for each location. I believe I may have briefly mentioned when you were starting this issue to try having a different color for each location (like green for in-person, teal for hybrid, and blue for online). I don't blame you if you forgot though, or I may be misremembering completely anyway.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx
```diff
@@ -59,141 +63,192 @@ const QuestionCard: React.FC<QuestionCardProps> = ({
...
+                  hasUnresolvedRephraseAlert={false}
+                  tasksSelectedForMarking={tasksSelectedForMarking}
+                  className="align-center flex items-center justify-around"
+                />
+              </Col>
+            ) : null
+          ) : (
+            <Col className="w-full sm:w-auto">
+              <div className="text-md ml-0 h-full italic text-gray-600 sm:ml-2">
+                In the process of requeuing...
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx:236]: Doing this kinda breaks the column of times (looks especially weird for students).
image
image

On mobile, this text ends up ends up getting wrapped around, causing the question card to grow larger, which is not really ideal since i'd argue we're wanting to draw less attention to these types of questions and that increasing their size gives them more attention.
image
Though surprisingly it doesn't look too bad on the TA-side of things
image

Nonetheless, I'd recommend doing the floating text instead of this like I had mentioned in my original comment, as that will kinda fix these issues with lining things up and consistency. And by floating text i mean like the "currently being helped text" but instead it says "Not Ready" or "Requeueing..." or something. As much as the floating text isn't ideal, it's kinda the only solution as conditionally adding anything to a question card is surprisingly hard to do while still making everything look consistent and well-aligned.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx
```diff
@@ -162,17 +177,25 @@ const QueueInfoColumn: React.FC<QueueInfoColumnProps> = ({
       )}
 
       {/* mobile only */}
-      <div className="mt-3 flex items-center justify-between sm:hidden">
-        {!isStaff && hasDemos && buttons}
+      <div className="mt-5 flex items-center justify-around sm:hidden">
+        {!isStaff && buttons}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx:181]: oh yeah were you going to try having a large "Join Queue" button here? With width something like 80 or 90%?
image

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }, [open])
+
+  return (
+    <Modal
+      open={isClosable ? localOpen : open}
+      footer={[]}
+      style={{ top: '30%' }}
+      closable={isClosable}
+      onCancel={() => setLocalOpen(false)}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx:45]: Okay I found a very interesting bug with this! 
Steps to reproduce:
- Start getting helped
- Before the 10s is over, click somewhere on the gray-ed out background
- Notice that once the 10s is up, the modal closes automatically
image
This is pretty interesting, and I guess it's because when you click on the grayed out background, it will run this onCancel which will setLocalOpen to False, meaning that once isClosable is true, it will  immediately set open to false. Luckily, I think all that's require to fix this is by adding a check to make sure isClosable is true before setting setLocalOpen to false.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx
```diff
@@ -679,11 +680,25 @@ const EditQueueModal: React.FC<EditQueueModalProps> = ({
           </Form.List>
         </>
       )}
-      <Form.Item label="Zoom/Teams Link" name="zoomLink">
+      <Form.Item
+        label={
+          <div className="flex flex-row items-center gap-1">
+            Zoom/Teams Link
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx:686]: Oh yeah I guess if the queue is in-person, we could probably hide this field.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx
```diff
@@ -679,11 +680,25 @@ const EditQueueModal: React.FC<EditQueueModalProps> = ({
...
+            </Tooltip>
+          </div>
+        }
+        name="zoomLink"
+      >
         <Input
           allowClear={true}
           className="text-sky-800"
-          placeholder="[No Zoom/Teams link set]"
+          placeholder={course?.zoomLink ?? '[No Zoom/Teams link set]'}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EditQueueModal.tsx:701]: Huh interesting thing with this. So I guess nullish coalescing operator only checks if the value is null or undefined, not if it's falsy. I guess in this case all you would want to do is replace it with an `||`
```suggestion
          placeholder={course?.zoomLink || '[No Zoom/Teams link set]'}
```
image
image

This made me realise I may have been using the nullish coalescing operator in places I shouldn't have been using it lol, and I should have been using || instead

packages/frontend/app/(dashboard)/course/[cid]/components/QueueInfoColumnButton.tsx
```diff
@@ -1,6 +1,12 @@
 import { cn } from '@/app/utils/generalUtils'
 import { Button, ButtonProps } from 'antd'
-import { PropsWithChildren } from 'react'
+import React, { PropsWithChildren } from 'react'
+
+const JoinZoomButton: React.FC<
+  PropsWithChildren<QueueInfoColumnButtonProps>
+> = (props) => (
+  <QueueInfoColumnButton type="primary" onClick={props.onClick} {...props} />
+)
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueInfoColumnButton.tsx:9]: With this, do we think we can rename this to something else purely because there is another component named the exact same?

Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueInfoColumnButton.tsx:9]: Actually wait, after further investigation, I realised that this component actually isn't being used anywhere. The one on the queue info column is using the other JoinZoomButton
image

packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:94]: ~~Idk I still kinda disagree. I think knowing the location type of the queue is not useful information in a lot, if not most siutations. And every bit of information on the page adds more cognitive load to process. And the two pages with the worst cognitive load imo is the course page and the queue page.~~

I have been thinking more about it, it isn't *that* much more cognitive load i guess. I have mixed feelings about it I guess. I guess the location on the Queue Cards anyway can actually give a bit of a signifier that this is something with a location, which contrasts with other parts of the system (e.g. the anytime question hub) in a good way. Ugh I've put a lot of thought into this. Yeah, it's probably fine. Maybe if another idea comes up in the future we could make a new issue for that. But otherwise yeah I guess it is fine to display the queue location.

Also as a side note, I guess the reason why I had the queue notes where they were before was to sorta have a seperation of the "queue details" part of the page from the "questions" part of the page. Though it did look pretty jank up there, and I'm not unhappy with it's new spot.

Sorry if I'm dragging my feet when it comes to adding new stuff like this. I guess I just spent so much time trying to reduce unneeded information or cleaning it up around the course and queue page (every word counts). So when we add information or visual elements that I feel might be unnecessary or don't add enough value for the amount of cognitive load induced, then I guess you get long windy rants from me lol

TL;DR: I changed my mind, displaying the queue locations is fine 👍

AdamFipke (changes requested):
See comments. I like the changes, especially with the JoinZoomModal.

As a small thing I added a line break in the text like so:
image
(had to use a <br> tag instead of \n)

packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:91]: image

What do you think about these colors?

Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:91]: i tried to keep them fitting the theme but realized its really hard to distinguish them (which would be the point of this) unless we have some warm colors in there too.

Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:91]: Ah I see yeah fair enough. I like your color choice for these, though I think maybe it stands out a little too much maybe? Hybrid tag looks good, but I wonder how it would look like if we chose some shades of color that are closer to the color of the Queue Card header for In-Person and Online?

image
image
Or in other words, maybe try what it looks like if we choose colors that don't contrast as much

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/JoinZoomNowModal.tsx:45]: I just used another prop: "maskClosable" which just disables the ability to close the modal by clicking off entirely which I think is much better since there is no way to bring the modal back. So having a close button be the only button that can close the modal I think makes it most clear and prevents more accidental "misclicks".

AdamFipke (approved):
Looks good! Going to merge this, then work on one more PR, and then we can put this all on prod

PR #163, opened 2024-10-12
Anytime Question Comments, Guards, and Refactoring

### LINKED ISSUE #137: Anytime Question Comments
This description is outdated, see the title and the associated PR
**Description**

Current Anytime questions has received postive feedback and reviews. However, there is no room for follow up questions/discussions. 
This feature should allow discussions for each ticket. The AI answers should also contain links and sources. 

**Success Criterias**

1: Allow students to post follow-up questions or comments on existing Anytime Questions tickets. This will enable a discussion thread for each question, facilitating back-and-forth interactions between students and the course staff (TAs and professors).

2: Integrate the ability for the course chatbot to provide links and sources for automated responses like the chatbot.

karel-harjono:
Note that I (Adam) have taken over this PR

# Description

Adds anytime question comments, improves the student anonymous experience (and security), and a bunch of other things. Here's some photos

image
image
image
(note that Ramon's pfp is only broken on this browser because google keeps giving me 429 too many requests lol)
image

This also:
- Frontend:
	- Now utilizes Next.js's `<Image>` component where possible (or removed the unoptimized flag)
		- This is because Next.js will automatically re-size the images based on the viewport and how big the images will take up in the DOM, which should dramatically increase performance in some cases where there are lots of images
		- All that was really needed to do this was a loader function which is just turns relative URLs (e.g. `/image1`) into absolute ones (e.g. `https://coursehelp.ubc.ca/image1`)
		- The big incentive to finally figure this out and do this was getting the anonymous animal pictures and serving them on our system
	- Fixed a security issue where anyone could use our helpme server to optimize their own next.js images
	- Makes it so newlines are actually parsed by the CustomMarkdown component (previously it would just put a space instead of a newline). It also now parses Github Flavoured Markdown (needed for tables, strikeout, footnotes, and more)
	- Allows students to edit their questions even after they click "Satisfied" or "Needs staff help"
  - Allows TAs to subscribe/unsubscribe from the "AIAnsweredNeedsAttention" email
  - Makes it so when a student gets their AI answer, the AsyncQuestionCard will start as expanded
- Backend:
	- Added an AsyncQuestionROlesGuard that will take in a `:qid` and verify that the user making the call is in the same course as the question's course
	- Wrote a ton of new tests async questions (something like 20-30 + 37 new for the emails for a total of 60+ new tests)
	- Made various adjust and improvements to the email templates (e.g. unsubscribe link takes you to profile page, fixed null questionText when empty, adjusted information and wording on the emails to be more clear, etc.)
	- Refactors a lot in regards to async questions (and fixed a bunch of bugs and unintended behaviours)
	- Deleted some old Khoury College stuff I stumbled upon and knew for certain it was unused
	- Made emails default to writing to a log file instead of actually sending out emails if on dev environment
	- Removed redundant "content" field in mail-services.entity
  - Made users by default subscribed to all email notifications (rather than having new users opt-in)

There are 70 anonymous animals with 70 colours to pick from for a total of a 1 in 4900 chance of a collision (of having the same animal and colour) when two people comment on an anytime question

Closes #137 , #217

## Type of change

	- Though it's not as breaking as other breaking changes. Mostly the changes to Images and emails may be breaking in unexpected ways but probably not

Inside `.env.` on /frontend, add
```
NEXT_PUBLIC_HOST_PROTOCOL=https
NEXT_PUBLIC_HOSTNAME=coursehelp.ubc.ca
```

Inside `.env.` on /frontend, add
```
NEXT_PUBLIC_HOST_PROTOCOL=http
NEXT_PUBLIC_HOSTNAME=localhost
NEXT_PUBLIC_DEV_PORT=3000
```

```sql
INSERT INTO mail_services ("mailType","serviceType","name","content") VALUES 
('member','async_question_new_comment_on_my_post','Notify when someone comments on your anytime question','Someone commented on your anytime question'),
('member','async_question_new_comment_on_others_post','Notify when someone comments on an anytime question you commented on','Someone commented on an anytime question you commented on');
```

# How Has This Been Tested?

Changes were thoroughly manually tested.

20-30 new integration tests for async centre endpoints (for a total of 37)
image

37 new integration tests for making sure the emails are sent with correct behaviour (I even found a handful of bugs thanks to these)
image

# Checklist:


packages/common/index.ts
```diff
@@ -634,6 +635,30 @@ export class AsyncQuestionVotes {
...
+  @IsString()
+  commentText!: string
+
+  @IsString()
+  createdBy!: string
+
+  @IsDate()
+  createdAt!: number
+}
+
```
AdamFipke, line 661: As a tip for the future with types, have one type for creating comments (like AsyncQuestionCommentParams) which has all the decorators (which will get auto-validated with requests) as well as one type for the returned values (could just do AsyncQuestionComment) that has no decorators (as we don't need to validate the values coming from our own endpoint)

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx
```diff
@@ -159,7 +161,7 @@ const AsyncQuestionCard: React.FC<AsyncQuestionCardProps> = ({
             <div className="mb-1 flex justify-between">
               <div className="flex flex-grow">
                 {showUser && (
-                  <>
+                  <React.Fragment>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx:164]: why are you changing this shorthand syntax for a react fragment into the longhand one? (doing `<>` and `<React.Fragment>` is the same thing https://react.dev/reference/react/Fragment https://www.reddit.com/r/reactjs/comments/9yy3f4/or_reactfragment/

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
```diff
@@ -0,0 +1,55 @@
...
+const Comment: React.FC<CommentProps> = ({
+  author,
+  avatar,
+  content,
+  datetime,
+}) => {
+  return (
+    <div className="flex items-center border-b border-gray-200 py-4">
+      {/* Avatar */}
+      <React.Fragment>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:21]:
```suggestion
      <>
```

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
```diff
@@ -0,0 +1,55 @@
...
+          anonymous
+        />
+        <UserAvatar
+          size={34}
+          username={author}
+          photoURL={avatar}
+          className="mr-2 flex md:hidden"
+          anonymous
+        />
+      </React.Fragment>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:36]:
```suggestion
      </>
```

AdamFipke:
I have some general UI comments

image
- Don't show the "comments" text if there is no comments
- In the case there are no comments, you will need to look at antd's API for the list element and see how you can remove the big "No Data" thing
- Change the "Comment on this post" style to be a link style rather than primary (change type="primary" to type="link") and make it smaller
- Put the "Comment on this post" button to be on the same Row as the little arrow indicator for isExpanded (at the bottom of the card)
- When the user is creating a comment, lock the card so it's always expanded (and also remove the little isExpanded arrow)

image
- Add a loading state to the "Add Comment" button
- Change the cancel button to be a regular button, not a primary one
- If the text field has stuff in it when you click cancel, show a Popconfirm saying "Are you sure you want to cancel? Your comment will be discarded"

image
- I would reduce the margins between each comment, as well as the margins between the "Comments" text and the first comment


image
I somehow got this error after commenting, might want to update the types and the checks to reflect this since i guess this is possible somehow

AdamFipke:
Okay i also have an idea on how this use case could go

Instead of having it like this with the comments immediately visible and the create comment button immediately visible...
image

...instead, have everything shrunk with a comments button that will expand both the card *and* the comments section at the same time.
image
That way, the question cards can still be compact when not expanded, and can also make more sense imo since it may be confusing that the question text is truncated with "..." but then there's a bunch of comments just underneath (like people might thing the "..." is part of the question itself maybe). If there are no comments, you could just put "comments" instead of "comments (1)" (the brackets is just num of comments)

AdamFipke:
Also, as another idea, I realised that students may put "Thanks anonymous panda!" in the chat and users may be confused to who that is. My idea to fix that is when a student posts a comment, instead of showing the user's own pfp and name, it puts their anonymous animal with (You) in green beside it. Kinda like so:
image

Could also put (Author) in blue beside the name if the commentor is the original user as the poster (do this if question.creatorId is equal to comment.creatorId)
image

Also, idk if you already have this, but TAs and profs should not appear as anonymous probably

AdamFipke:
I am unable to test if this is the case in the student's view (due to the comment.creator is null error earlier), but i realise the endpoint is returning the student's name and photoURL in the browser even though they appear as anonymous

image

AdamFipke (changes requested):
See comments above. Good work so far. I have not reviewed all the code yet. You will also want to add tests for your replyToQuestion endpoint (and maybe some in the course controller with getAsyncQuestions. I would check to make sure the endpoint is not returning other students' name and photoURL in the comments ).

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx
Comment (karel-harjono) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx:164]: I'm against shorthands for react.fragment, but if it's the norm for this project I'll give in :(

karel-harjono:
image

packages/server/src/course/course.controller.ts
```diff
@@ -133,14 +134,21 @@ export class CourseController {
     );
     let all: AsyncQuestionModel[] = [];
 
-    if (Object.keys(asyncQuestionKeys).length === 0) {
+    if (true) {
```
AdamFipke, line 137: Might want to change this back to using redis

```suggestion
    if (Object.keys(asyncQuestionKeys).length === 0) {
```

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -44,6 +45,62 @@ export class asyncQuestionController {
...
+    const comment = await AsyncQuestionCommentsModel.create({
+      commentText,
+      creator: user,
+      question,
+      createdAt: new Date(),
+    }).save();
+
+    const updatedQuestion = await AsyncQuestionModel.findOne({
+      where: { id: questionId },
+      relations: ['creator', 'taHelped', 'votes', 'comments'],
```
AdamFipke, line 93: I can't leave a comment on the right lines but there's some updatedQuestion queries in the functions below that you will want to update to include 'comments'
relations: ['creator', 'taHelped', 'votes'], -> relations: ['creator', 'taHelped', 'votes', 'comments'],

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -44,6 +45,62 @@ export class asyncQuestionController {
...
+    });
+
+    if (!question) {
+      res
+        .status(HttpStatus.NOT_FOUND)
+        .send({ message: ERROR_MESSAGES.questionController.notFound });
+      return;
+    }
+
+    // check if user has commented within 2 minutes
```
AdamFipke, line 68: 2 minutes might be a bit overkill. Something like 30s should be fine.

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -44,6 +45,62 @@ export class asyncQuestionController {
...
+    const lastComment = await AsyncQuestionCommentsModel.findOne({
+      where: { creator: user, question },
+      order: { createdAt: 'DESC' },
+    });
+
+    if (lastComment) {
+      const timeDiff = new Date().getTime() - lastComment.createdAt.getTime();
+      if (timeDiff < 30000) {
+        res.status(HttpStatus.BAD_REQUEST).send({
+          message: ERROR_MESSAGES.questionController.createQuestion.tooFast,
```
AdamFipke, line 78: Also, I realised it would be pretty easy to change the error message to be custom. Something like:
```suggestion
          message: `You are posting too quickly. Please try again in ${timeDiff / 60000}s`,
```

packages/common/index.ts
```diff
@@ -131,6 +131,8 @@ export class UserPartial {
   @IsInt()
   @IsOptional()
   sid?: number
+
+  userRole?: string
```
AdamFipke, line 135: I think adding this `userRole` is what's causing the tests to fail?

You might need to make a seperate type for this? Maybe it's just missing decorators? Idk

AdamFipke:
So Karel is going in deep with cool AI things for his honours so I'll be taking this over from here

packages/common/index.ts
AdamFipke: UPDATE: so I was kinda right about this. `userRole` is already used inside user.entity, and I guess typeorm or whatever class converter thing saw the matching `userRole` on this type, it tried to convert it, but then complained due to the lack of decorators and that its of type string instead of type Role. The solution was just to rename it to courseRole lol

packages/server/tsconfig.json
```diff
@@ -1,17 +1,18 @@
...
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
-    "target": "es2017",
+    "target": "ES2023",
     "sourceMap": true,
     "baseUrl": "./src",
     "outDir": "./dist",
     "incremental": true,
     "skipLibCheck": true,
+    "lib": ["ES2023"],
```
AdamFipke, line 15:
```suggestion
    "lib": ["ES2023", "DOM"],
```
trying this, having issues locally rn

packages/frontend/app/(dashboard)/components/FooterBar.tsx
```diff
@@ -37,6 +37,8 @@ const FooterBar: React.FC = () => {
     >
       <div>
         <a
+          aria-hidden="true"
+          tabIndex={-1}
```
AdamFipke, line 41: this was to fix a warning that there was a targetable element under an aria-hidden element (I wanted the footerbar to be hidden to screen readers since the information it carries gives very little value)

packages/frontend/next.config.mjs
```diff
@@ -1,19 +1,31 @@
 import {withSentryConfig} from "@sentry/nextjs";
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  reactStrictMode: false,
+  // reactStrictMode: false,
```
AdamFipke, line 4: I turned on reactStrictMode and have not actually noticed any issues when developing so I think I'll leave it commented out (i.e. on). If you guys notice any problems while you're developing we can turn it off

packages/server/src/course/course.controller.ts
```diff
@@ -1077,7 +960,7 @@ export class CourseController {
 
     // have to do a manual query 'cause the current version of typeORM we're using is crunked and creates syntax errors in postgres queries
     const query = `
-    SELECT "user_model"."firstName" || ' ' || "user_model"."lastName" AS name, "user_model".id AS id
+    SELECT COALESCE("user_model"."firstName", '') || ' ' || COALESCE("user_model"."lastName", '') AS name, "user_model".id AS id
```
AdamFipke, line 963: Oh yeah interesting thing, so apparently doing `'John' || ' ' || null` in postgres will just return "null" which is kinda disappointing. This does now explain why sometimes userIds are shown instead of the user's names in some of the many Select dropdowns we have on our frontend.

I want to make lastName not nullable (default to empty string) to fix this from ever happening again but that's for another PR. 

Reason why I fixed this here was I was recording a demo video for Ramon and I had some users without lastnames  and it was messing up some dropdown in the demo.

packages/server/src/profile/user.entity.ts
```diff
@@ -149,4 +152,14 @@ export class UserModel extends BaseEntity {
 
   @Column({ type: 'boolean', default: false })
   readChangeLog: boolean;
+
+  @OneToMany((type) => QuestionModel, (q) => q.creator)
+  questions: QuestionModel[];
+
+  @OneToMany((type) => AsyncQuestionModel, (aq) => aq.creator)
+  asyncQuestions: AsyncQuestionModel[];
```
AdamFipke, line 160: These should have been there before (to specify the other side of the OneToMany relationship)

packages/frontend/next.config.mjs
AdamFipke: Nevermind it turns out it messes with antd's forms, causing all the default values on them to not work properly

AdamFipke (approved):

AdamFipke:
Approving to get rid of my old change request

packages/frontend/next.config.mjs
```diff
@@ -1,19 +1,31 @@
...
+    loaderFile: './loader.js',
+    // remotePatterns is a list of where all of our images can come from. It is needed otherwise malicious users can use our server to optimize their own images
     remotePatterns: [
       {
-        protocol: "https",
-        hostname: "**",
+        // allow images from the same host
+        protocol: process.env.NEXT_PUBLIC_HOST_PROTOCOL,
+        hostname: process.env.NEXT_PUBLIC_HOSTNAME,
+        port: process.env.NEXT_PUBLIC_HOSTNAME === 'localhost' ? process.env.NEXT_PUBLIC_DEV_PORT : '',
```
patrickma6199, line 20: Why is this set to an empty string if the hostname is not localhost? If we're expecting that it will always run on localhost, we don't need an env variable here (just hardcode it since its not sensitive data). If we won't always set it as localhost, there should be another value here.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx
```diff
@@ -33,30 +34,55 @@ const statusDisplayMap = {
...
-  isStaff,
+  userCourseRole,
   userId,
   courseId,
   mutateAsyncQuestions,
+  showStudents,
 }) => {
+  const [isLockedExpanded, setIsLockedExpanded] = useState(false)
   const [isExpanded, setIsExpanded] = useState(false)
+  const [isExpandable, setIsExpandable] = useState(true)
```
Comment (patrickma6199) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx:54]: Might be good to add comments here to explain what the differences of these three states are for future devs working on this component.

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -407,4 +339,407 @@ export class asyncQuestionController {
 
     return question;
   }
+
+  @Post('comment/:qid')
```
patrickma6199, line 343: Assuming that the bugfix for the unread counts is merged, you should add a call to update all UserCourseAsyncQuestion entities so that users know that a particular question changed before merging (assuming all is good with my implementation).

patrickma6199: I left it outside of the async question service since it was empty at the time. If you would like, since you changed the class name, it would have the lowest chance of conflicts if you move the db call to instantiate and to update the readLatest attribute to service functions for me after the merge rather than me moving the functionality to a service function myself.

packages/frontend/next.config.mjs
bhunt02: presumably because the next application on prod uses default port 80, although I'm curious to see if this has been verified to not cause any issues

packages/server/test/util/testUtils.ts
```diff
@@ -122,3 +124,41 @@ export async function clearAllCronJobs(
     schedulerRegistry.deleteCronJob(jobName);
   });
 }
+
+export const mockRedisQueueService = {
```
patrickma6199, line 128: Check the testing config I use in Queue Chats. Since there is an actual redis container in the Ubuntu VM, it is going to test more robustly since we arent merely mocking redis' behaviour. For local tests, I integrated redis-memory-server to run a lightweight instance of redis through node to complete the tests then stop.

packages/server/src/guards/async-question-roles.guard.ts
```diff
@@ -0,0 +1,39 @@
...
+      .select('"courseId"')
+      .where('id = :questionId', { questionId })
+      .getRawOne<{ courseId: number }>();
+
+    if (!aq) {
+      throw new NotFoundException('Async question not found');
+    }
+
+    return { courseId: aq.courseId, user };
+  }
```
patrickma6199, line 38: Whats the purpose of sending the user object back here as well? I'm assuming that this guard is to be used in all routes that handle async questions primarily. Querying and returning the user here adds nothing to any of the checks made to the async questions before returning the course id (since the aq object comes only from a query on the questioned). To follow the single responsibility principle, you should remove the user here entirely. We already have the two param decorators that fulfill this task. Imagine having to change all instances of querying for a user if something critical to the query changes. Better to keep these all in the same place (the param. decorator).

packages/frontend/next.config.mjs
AdamFipke: hmm now that I think about it more I'm not actually sure how this will work, but I unfortunately i'm not sure how to tell other than putting it on prod and seeing if things break and then adjusting it (would probably have been a good use for a staging server but oh well). Unless maybe @wskksw maybe knows something more that could help us?

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
AdamFipke: Yeah good idea, I can do that

packages/server/src/guards/async-question-roles.guard.ts
AdamFipke: Oh um so anything that extends the roles guard needs to return the courseId and the user (that way it can do the checks). These things do not get passed into the controller, they get passed into the RolesGuard (which the asyncQuestionRolesGuard extends) as a part of the setupData method. 

If you look at course roles guard, this pretty much works the same as that except instead of grabbing the courseId from the URL it grabs the questionId, finds the courseId for that question, and then uses that.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
```diff
@@ -0,0 +1,314 @@
...
+            }
+            size={40}
+            // the colour of the avatar is based on the username
+            // the name is authorId + questionId % length of ANIMAL_NAMES
+            // while the colour is just authorId + questionId
+            username={isUserShown ? author.name : 'Anonymous Student'}
+            colour={author.colour}
+            photoURL={
+              isUserShown
+                ? author.photoURL
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:97]: image
don't we have weird default PFPs with the initials for this elsewhere in the application?

bhunt02 (changes requested):
Mostly all good, some things are minor. Definitely make sure there's a null check on comments (if not just for making using the dummy data more comforting)

Also, definitely check out the profile picture-in-case-there's-not-one-and-the-user-isn't-a-student use case

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:97]: oh shoot we should! From my testing it was fine but I will re-check.

I will say I notice that what you got there sometimes happens if the image URL fails for some reason (e.g. google giving 429 too many requests)

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -407,4 +339,407 @@ export class asyncQuestionController {
...
+        'votes',
+        'comments',
+        'questionTypes',
+        'votesSum',
+        'isTaskQuestion',
+      ]);
+
+      const filteredComments = question.comments?.map((comment) => {
+        const temp = { ...comment };
+        // TODO: maybe find a more performant way of doing this (ideally in the query, and maybe try to include a SELECT to eliminate the pick() above)
```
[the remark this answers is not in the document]
AdamFipke: yep, though I realise that this might be more effort than its worth to try to figure out how to do (at least in this PR), mostly since some of this data processing stuff uses functions like nameToRGB, which might be hard to replicate in postgres (so either way a pass through of all the questions and comments would be required).

I adjusted this TODO comment though

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx
```diff
@@ -296,29 +390,57 @@ const AsyncQuestionCard: React.FC<AsyncQuestionCardProps> = ({
...
                   Still need faculty Help
                 </Button>
               </div>
             )}
         </Col>
       </Row>
-      <Row className="justify-center">
-        {isExpanded ? <UpOutlined /> : <DownOutlined />}
-      </Row>
+      {!isLockedExpanded && (
```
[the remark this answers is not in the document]
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx:421]: hmm, that is weird since it still did work for me even if there's no comments...

oh wait, it was probably redis actually lol. It probably still had old cached async questions where comments was undefined/null instead of an empty array.

I am going to adjust the endpoint so that if there are no question.comments, comments will become an empty array

docs/NEWDEVS_STARTHERE.md
```diff
@@ -109,19 +119,109 @@ Want to conditionally render something like an `elseif` statement? Do: `{conditi
...
+For example, you might want to check if the user is a TA or Prof in any course, so you do:
+```ts
+const user: UserModel = await UserModel.findOne(userId);
+const isStaff = user.courses.some((userCourse) => userCourse.role === Role.PROFESSOR || userCourse.role === Role.TA);
+```
+This will result in a *runtime* error (so you get no warning beforehand) because `user.courses` will be undefined because the database join was not made.
+
+You can perform the database join like so (though note that the syntax for this will change on typeorm version upgrade):
+```ts
+const user: UserModel = await UserModel.findOne(userId, {
```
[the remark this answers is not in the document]
AdamFipke: I have updated the docs now thanks!

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:97]: Yeah I think I might be right? Since it's working fine for me (and it seems to work fine on other pages too)
image

I think the issue you have there could be one of two things:
- it's a google URL that failed to fetch for whatever reason
- The image is missing from the uploads folder (this could happen if you're uploading a backup)

The behaviour you have there is how it would look currently on production I believe

bhunt02 (approved):
Everything looks good, thank you for addressing my comments!

PR #176, opened 2024-10-22
made question type name, color, cid not nullable

AdamFipke:
# Description

Small change.

Note that before this gets put on production, the following command should be ran:
`UPDATE question_type_model SET name = 'was null', "deletedAt" = NOW() WHERE name IS NULL`

Normally we could just set up a migration for this. However, we don't actually have migrations set up on prod.


bhunt02 (approved):
Really simple changes, thanks for handling it!

PR #179, opened 2024-10-26
Rate Throttler + Login Fixes

### LINKED ISSUE #177: Login Fixes
- Fix Login Unhandled Exceptions
- Improve error messages
- Add rate throttler

AdamFipke:
# Description

Fixes some issues with the login page (see #177)

The biggest thing here is this prevents brute-force attacks (since there was previously nothing stopping malicious users from just calling the login endpoint with every password, which is really bad).

Also I guess this gives some DoS protection now with the throttler. 

Note that the rate throttler does not apply to GET /profile since the middleware calls that endpoint a crap load (basically every time a link appears)

Closes #177 

image

## Type of change

# How Has This Been Tested?

I've tested lowering the rate throttler a lot to see what happens when certain things get rate throttled. Usually, some weird stuff happens, and it's kinda impossible to go through and have an edge case check for all of these, but I don't think it will be much of an issue for users since I don't think any of our UI does calls at 30/sec (and if it does, we kinda have a different problem).

The weirdest case to cover was if there was a "too many requests" error from the middleware since the middleware basically has to return a route, to which I added an error page for. However, I then just decided to not throttle GET /profile and that basically solved the issue. I am leaving the 429 error page though in case that becomes an issue in the future (e.g. someone makes middleware do a request to another endpoint).

Also as a side note, I tested this throttler by building the site and running it. (I added a new `yarn build:dev` command for when you want to build without sentry getting initialized)

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "dev:db:reset": "yarn workspace @koh/server run dev:db:reset",
     "migration:generate": "yarn workspace @koh/server run migration:generate",
     "build": "cross-env NODE_OPTIONS=--openssl-legacy-provider yarn workspace @koh/server run build && yarn workspace @koh/frontend run build",
+    "build:dev": "cross-env NODE_OPTIONS=--openssl-legacy-provider yarn workspace @koh/server run build:dev && yarn workspace @koh/frontend run build",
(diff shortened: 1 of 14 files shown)


bhunt02 (approved):
Wow, these should have probably been merged a while ago. Everything looks good, thanks for doing this and good job!

Issue #170, opened 2024-10-18
Server-Side TA auto-checkout


AdamFipke:
**Description**

TAs are auto-checked out in the frontend after 5 hours. (commonCourseFunctions.tsx) 
- this timeout will stop if the user switch pages or reloads or closes the page
- multiple tabs open will cause it to happen multiple times
- 5 hours is kinda long anyways

**possibly better solution**: Set up a cron job that automatically goes through all staff lists say every 10 minutes. When checking the staff member, check them out if: 
- it's been at least 3 hours since they have checked in (maybe make this customizable at the course-level)
- they have not helped a student in the past hour
- maybe some other intelligent ways logic to see when TAs stop interacting with the system?

AdamFipke:
should have been closed when #180 was merged

PR #180, opened 2024-10-27
TA Auto checkout

### LINKED ISSUE #58: Auto-Checkout staff after event ends
**Description**

After a queue session (e.g. a lab) ends, the staff are auto-checked out. Or, alternatively, they receive a notification maybe 10mins after the session ends asking if they would like to end the lab (options: "don't remind me", "5 more minutes", "end lab").

This sounds simple, but right now calendar events are unlinked from queues. This would require some way to connect calendar events with queues, which can then be used to control when a queue can be opened or closed.

UPDATE: ramon actually suggested a really simple way to do this, and that is to basically allow you to connect TAs with events (rather than queues). From there, we could simply have a cron check every 10mins or something that auto-checks out staff (and maybe even keep a record of when staff checkin late). Though this may still require some more advanced logic

AdamFipke:
# Description

Allows staff to be assigned to events. Once the event ends, assigned staff will be prompted to checkout. If they don't respond after 10 minutes, they will automatically be checked out. They also have an option for 10 more minutes, after which they will be prompted again if they want to check out. They can keep asking for 10 more minutes.

Works for both recurring events and one-time events.

Also fixes some things:
- some fixes around create/edit events
- Made it so that when creating an event, when you set the end date, that now gets set in localstorage so that if you are mass-creating events (e.g. for the start of a semester) they all already have the right end date
- Fixed the search and pagination of the organization users table
- queues now get properly cleaned at midnight (they weren't before because of some old logic issue)
	- by clean = all TAs checked out, all questions closed

Closes #58 and #170

## Type of change

# How Has This Been Tested?

- Added some calendar service tests for the cron jobs services (including auto-checkout tests).
	- unfortunately, I cannot use jest's useFakeTimers when the cron jobs run queries and stuff, so I can't do a direct test where a TA is checked in, the event ends, and check to see if they are checked out after 10mins. The next best thing I can do is to test each autocheckout function individually, and just trust that the cron job will run the function.

- Added a ton of new tests for the calendar integration tests (they were kinda a pain to setup but i'm glad they are working. Having good tests for something like the cron jobs is important). These are all of the calendar tests now (probably 2-3x as many as before:
Calendar Integration
    POST /calendar/:cid
      √ adds a new calendar event by TA should work (1940 ms)
      √ allows the user to pass in staffIds, creating calendar_staff entries (2018 ms)
      √ should return 403 if a student tries to create an event (1432 ms)
      √ should return 404 if the staffId does not exist (1559 ms)
      √ should return 404 if the course does not exist (1420 ms)
      √ should create a 1-time cron job for non-recurring events (1923 ms)
    PATCH /calendar/:calId/:cid
      √ does not allow non-logged in users to update an event (1336 ms)
      √ does not allow students to update an event (1618 ms)
      √ does not allow professors from other courses to update an event (1838 ms)
      √ updates an existing calendar event (1874 ms)
      √ updates the staff for an event (2129 ms)
      √ does not allow the endpoint to be called with an invalid date and staff list (1570 ms)
      √ updates the auto-checkout cron jobs with new startDates and endDates (2860 ms)
    GET /calendar/:cid
      √ gets all events for a course by a student should work (1487 ms)
    DELETE /calendar/:eventId/:cid/delete
      √ deletes a calendar event (1526 ms)
      √ should return 403 if a student tries to delete an event (1367 ms)
      √ should delete an auto-checkout job for each staff member (2436 ms)

- Also added some basic tests for the getcronjobs endpoint (basically to make sure only admins can call it)

# Checklist:


packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EventEndedCheckoutStaffModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+        />,
+      ]}
+      closable={false}
+    >
+      <div className="flex flex-col items-center">
+        <p className="text-lg">
+          The queue session has ended. Would you like to check out?
+        </p>
+        <p className="text-gray-500">
+          You will be auto-checked out in 10 minutes.
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/EventEndedCheckoutStaffModal.tsx:50]: maybe a bit informal, i'd change to "automatically checked out..."

packages/frontend/app/(dashboard)/organization/components/UsersTable.tsx
```diff
@@ -145,16 +146,16 @@ const UsersTable: React.FC<UsersTableProps> = ({
...
-            showSizeChanger={false}
-          />
-        )}
+        <Pagination
+          className="float-right"
+          current={page}
+          pageSize={50}
+          showQuickJumper
+          // set the total number of users very high so that it shows enough pages
+          total={5000}
```
bhunt02, line 155: was this not "users.total" because the pagination component would determine the number of page switcher options to show from the pageSize attribute given the value of the total attribute?

packages/server/src/profile/user.entity.ts
```diff
@@ -131,7 +132,8 @@ export class UserModel extends BaseEntity {
 
   @AfterLoad()
   setFullNames(): void {
-    this.name = this.firstName + ' ' + this.lastName;
+    // it is possible that lastname is null
+    this.name = this.firstName + ' ' + (this.lastName || '');
```
bhunt02, line 136: if we're using logical or, we're probably not going to get the desired result. would this not append true or false?

if this.lastName is null, it's falsy, and '' is falsy, returning false.
if this.lastName is not null, it's truthy , and '' is falsy, returning true.

maybe replace with a null coalescing operator (??) to prevent that?

bhunt02 (approved):
there's a lot here that goes over my head (maybe just the sheer number of changes and how broadly spread they are), but I remember your demo working fine, so I did light review of the code overall. not going to request changes because I'm sure you'll check my comments either way

packages/server/src/profile/user.entity.ts
AdamFipke: ah i think you're right, good catch!

packages/frontend/app/(dashboard)/organization/components/UsersTable.tsx
AdamFipke: no so actually users.total is not actually an attribute. When dima was making this component, he left `users` as `any` type so it didn't error when he did users.total. Also, doing users.length does not work since users will always be of length 50. We would need to change the backend to also send a count of the total number of users to the frontend, which i guess is something to be done, but i opted for this quick-n-dirty approach for now since this only affects us admins.

Previously, it would only show 50 users total (and i don't think the search worked either) so I just wanted to get it working. I will add this as a TODO though

bhunt02: oh, i see. there's an impl that works for it on the course side so I suppose I was mistaken

PR #175, opened 2024-10-21
Pause Questions

bhunt02:
# Description

Added ability for queue staff to 'pause' questions after they've been helped, in order to help more high-priority questions.
The question itself remains considered to be being helped - it is not moved back into the queue. While paused, a timer indicating how much time has elapsed since the question was paused is visible.
Additionally, other queue staff can pick up paused questions in case they clear up. This provides a method for queue staff to delay responding to certain questions in case of higher priority ones entering the queue, while also keeping the student on the top of the heap of questions to handle.
Changes to the 'helpedAt' column are made. Question entity now has an attribute 'pausedAt' that is essentially a temporary time-tracker and removed after the question is unpaused.

Closes #161 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This requires a run of `yarn install`
- [x] _This requires a database migration update_

# How Has This Been Tested?

Please describe how you tested this PR (both manually and with tests)
Provide instructions so we can reproduce. 

- [x] Integration test for the pausing process.
- [x] Manual testing with clients on different browser windows to ensure that relationships are retained and no unintended errors come about:
  - 1 queue staff, 1 student: pausing/unpausing
  - 1 queue staff, 2 students: ensuring that the 'help all' button is no longer present on pause
  - 1 queue staff, 3 students: ensuring that the 'help all' button is present when only 1 question is paused
  - 2 queue staff, 1 student: ensuring that the 'paused' question becomes visible to the other queue staff

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code where needed 
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any work that this PR is dependent on has been merged into the main branch
- [x] Any UI changes have been checked to work on desktop, tablet, and mobile

Issue #87, opened 2024-09-05
[Data] Question Duration incalculable with requeueing


AdamFipke:
**Description**

I realised that with questions, there are 4 datetimes: createdAt, firstHelpedAt, helpedAt, and closedAt
![image](https://github.com/user-attachments/assets/d2dd6850-574e-4ae1-bdf5-7b7bee8c19a1)

Normally, if you want to figure out how long the question took to get answered, you could just do closedAt - helpedAt.
There is a problem with this though, as this does NOT take into account the time taken where the student was helped *before* the re-queueing. Example timeline:

00:00 - question is first helped, firstHelpedAt and helpedAt set to 00:00
00:02 - question is requeued, nothing is set except the status
00:04 - question rejoins queue and is helped again, which sets helped at to 00:04
00:05 - question is marked as done, setting closedAt to 00:05

Problem: duration = closedAt - helpedAt = 0:01 when it should really be 0:03 since there is no way to figure out the first two minutes.

Solution: Maybe just add a "total duration" column that will accumulate time whenever the question is requeues (`totalduration += now() - helpedAt`) and whenever the question is closed (again, ``totalduration += now() - helpedAt`)

PR #185, opened 2024-11-09
More Accurate waitTimes and helpTimes

### LINKED ISSUE #174: Smarter wait times
**Description**

Right now, wait times are purely just based on when the question was created.

This means the wait time will go up while being helped (I think), but it also goes up if the question is in a limbo state (like being requeued, or drafting) 

i noticed in my class, if a student has a demo that needs to be marked and they are not ready so i requeue them, and then the join the queue 15mins later, it looks as though they have been waiting in the queue for much longer than they actually have

And wait time is basically defined as time when the question's status is either queued or paused.

maybe could do like a lastReadyAt date, with a waitTime
- waitTime would not always be up-to-date
- Whenever a question gets set from drafting/requeueing to queueing/paused, set lastReadyAt to now
- Whenever a question gets set from queueing/paused to any other status, set waitTime = waitTime + time since lastReadyAt
- On the frontend, we can display the wait time as waitTime + time since lastReadyAt

AdamFipke:
# Description

This is a really important PR that should be merged before we update prod with pause questions (#175 ) and the requeueing stuff in the queue locations one (#148). About 5% of questions have a helpedAt != firstHelpedAt, and these two PRs are going to increase that dramatically. And previously, any time that a question is helped more than once, it is impossible to properly calculate the waitTIme and helpTime.

Fixes
- You can now repeatedly pause/unpause a question and the waitTime and helpTime will be properly updated
- the wait time on question cards will no longer increase while the student is being helped or drafting or being requeued (because previously it wasn't the wait time, it was just the question's lifespan)
- Paused questions are now included in queueSize
- centralised the changeStatus service so that both students and TAs call it, making it more maintainable for the future (I didn't even know they had separate logic, when there really was no reason to)
- Removed paused time (and instead opted to just increase waitTime while paused, since they are fundamentally the same thing)
- Changed the pause timer on the questioncards to just be the helpTime but frozen (kinda like how you can pause a video, the timer stops until you click play again)
image
image

Closes #174 and #87 

## Type of change

- [x] This change requires a database query to update old data on production. These queries are below:

```sql
UPDATE question_model
SET "lastReadyAt" = "createdAt"
```
```sql
UPDATE question_model
SET "waitTime" = CEIL(EXTRACT(EPOCH FROM question_model."firstHelpedAt"::TIMESTAMP) - EXTRACT(EPOCH FROM question_model."createdAt"::TIMESTAMP))
WHERE "firstHelpedAt" IS NOT NULL
```
```sql
UPDATE question_model
SET "helpTime" = CEIL(EXTRACT(EPOCH FROM question_model."closedAt"::TIMESTAMP) - EXTRACT(EPOCH FROM question_model."helpedAt"::TIMESTAMP))
WHERE "helpedAt" IS NOT NULL AND "closedAt" IS NOT NULL
```

@bhunt02 please double check to make sure these queries to update the old data looks correct (I tried to basically copy the same calculation as the insights, and it looks right to me, but I want to double check with you in case it looks like there should be something else). I tested these queries and they work, just looking for any logic issues

**I also made some queries that we could run if we wanted maybe more accurate wait/help times during this migration?**

This one will subtract 1min from the waitTime if they have been helped more than once (helpedAt != firstHelpedAt) with a minimum waitTime of 1min,
```
UPDATE question_model
SET "waitTime" = GREATEST(CEIL(
  EXTRACT(EPOCH FROM "firstHelpedAt"::TIMESTAMP) - EXTRACT(EPOCH FROM "createdAt"::TIMESTAMP) -
  CASE
    WHEN "firstHelpedAt" != "helpedAt" THEN 60
    ELSE 0
  END
), 60)
WHERE "firstHelpedAt" IS NOT NULL;
```

This one will add 1 min to helpTime if they have been helped more than once (helpedAt != firstHelpedAt) with a **maximum helpTime of 1h.** -> This will also fix all the multi-day long helpTime issues.
```sql
UPDATE question_model
SET "helpTime" = LEAST(
	  CEIL(
	  EXTRACT(EPOCH FROM question_model."closedAt"::TIMESTAMP) - EXTRACT(EPOCH FROM question_model."helpedAt"::TIMESTAMP) +
	  CASE
	    WHEN "firstHelpedAt" != "helpedAt" THEN 60
	    ELSE 0
	  END
	), 3600
 )
WHERE "helpedAt" IS NOT NULL AND "closedAt" IS NOT NULL
```

# How Has This Been Tested?

- Added tests to see if the waitTime and helpTime will correctly update as the status of a question changes (from  Drafting -> Queued -> Helping -> Paused -> Helping -> Requeueing -> Queued -> Helping -> Resolved)
- Edited tests for insights
- All queries have been tested on a copy of the production database and behave how you would expect.

# Checklist:

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -445,11 +445,13 @@ export class Question {
   @Type(() => Date)
   helpedAt?: Date
 
+  // in seconds
   helpTime!: number
 
   @Type(() => Date)
   lastReadyAt?: Date
 
+  // in seconds
   waitTime!: number
 
   @Type(() => Date)
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -445,8 +445,12 @@ export class Question {
   @Type(() => Date)
   helpedAt?: Date
 
+  helpTime!: number
+
   @Type(() => Date)
-  pausedAt?: Date
+  lastReadyAt?: Date
+
+  waitTime!: number
 
   @Type(() => Date)
   closedAt?: Date
@@ -494,6 +498,13 @@ export enum ClosedQuestionStatus {
   ConfirmedDeleted = 'ConfirmedDeleted',
   Stale = 'Stale',
 }
+
+/** waitingStatuses are statuses where the student waiting to be helped */
+export const waitingStatuses: ReadonlyArray<OpenQuestionStatus> = [
+  OpenQuestionStatus.Paused,
+  OpenQuestionStatus.Queued,
+  OpenQuestionStatus.PriorityQueued,
+]
 
 export enum asyncQuestionStatus {
   AIAnsweredNeedsAttention = 'AIAnsweredNeedsAttention', // AI has answered, but the answer is unsatisfactory.
--- a/packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx
@@ -12,12 +12,7 @@ import { useState, useEffect } from 'react'
 import UserAvatar from '@/app/components/UserAvatar'
 import TaskMarkingSelector from './TaskMarkingSelector'
 import { QuestionTagElement } from '../../../components/QuestionTagElement'
-import {
-  getOriginalPausedTime,
-  getPausedTime,
-  getServedTime,
-  getWaitTime,
-} from '@/app/utils/timeFormatUtils'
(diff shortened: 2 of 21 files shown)


bhunt02:
Popping in for the equations: the standard ones look about right. The queries for adjusting helpTime when they are multiple days long looks okay as well.

bhunt02 (approved):
Changes look very thorough, I have no real concerns here and I think it'll make it a lot easier for future insight developments and data transparency.

PR #186, opened 2024-11-15
LMS API Integration

### LINKED ISSUE #182: Canvas API Integration
**Description**

Integration of the Canvas API to the HelpMe system so that assignment and student data can be imported for validation and chatbot response alignment.

**Success Criterias**

- [ ] Adapter pattern implementation for LMS integration
  - [ ] Implement the adapter for Canvas API which allows for all features listed below.
- [ ] Organization-based configuration of Canvas API
  - [ ] Organization administrators can enable Canvas integration and specify the API access URL.
  - [ ] Organization administrators will have to provide an API key to verify that their organization can access this Canvas root.
- [ ] Instructor Configuration of Canvas Course Integration
  - [ ] Instructors can provide a course ID from Canvas to link with the HelpMe system
  - [ ] Instructors will have to provide their own API key (??? or the organization API key could be universal, but this feels less secure)
- [ ] Student Validation
  - [ ] Using the Canvas API, the students of a course will be validated to determine if they are enrolled in the Canvas course or not.
    - [ ] This may require updates to the UserCourseModel in case Canvas is used within the Course
  - [ ] Automatic removal of students who are not part of the Canvas course?
- [ ] Chatbot Alignment
  - [ ] Canvas API will be used to retrieve assignment details from Canvas itself so that the Chatbot can ingest assignment information so it can be better aligned with the course information and more useful to students.
- [ ] All Canvas API interactions, sans inputting the configuration details, will be run through the backend so that responses can be strongly typed and filtered for data we don't want to send to just any client. This also removes the problem of sending the API key to the frontend from the DB (via the backend) so it may query the Canvas API itself, endpoints for Canvas interactions will need to be set up.

**Context**

- N/A

bhunt02:
# Description

Created a way for HelpMe courses to integrate with courses from other LMS platforms, allowing verification of students and (hopefully eventually) importing assignment data for the chatbots.

Closes #182 

## Type of change

# How Has This Been Tested?

- [ ] Unit test, avoided integration tests as they're not exactly important due to being the external platform

# Checklist:


packages/server/test/lmsIntegration.integration.ts
```diff
@@ -0,0 +1,5 @@
+describe('Lms Integration Integrations', () => {
+  it('should equal 2', async () => {
+    expect(1 + 1).toBe(2);
+  });
```
AdamFipke, line 4: > "Did you write tests for your feature?"

> "Yeah I wrote tests"

> The tests:

Kidding I read the PR description and yeah I agree we probably don't need many integration tests for this. Though, it may be a good idea to add some really simple integration tests for the new endpoints that are in the organization controller (though if it's hard to really make tests for these endpoints without mocking a bunch of stuff, maybe just some simple tests like "does it return 403 if the user is not an admin" or "does it return 401 if the user is not logged in")

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -0,0 +1,398 @@
...
+          <div
+            className={
+              'flex flex-col items-center justify-start gap-2 text-center text-lg'
+            }
+          >
+            <div className={'flex flex-col'}>
+              <p>
+                This course is not integrated with a learning management system.
+              </p>
+              {lmsIntegrations.length == 0 ? (
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:202]: So when I first go to this page, I get "TypeError: Cannot read properties of undefined (reading 'length')"
I guess there are multiple ways to fix this, but I suppose the quickest one is to change the endpoint to make sure it's returning an empty array

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -0,0 +1,398 @@
...
+  const [course, setCourse] = useState<LMSCourseAPIResponse>(
+    {} as LMSCourseAPIResponse,
+  )
+  const [lmsStudents, setLmsStudents] = useState<string[]>([])
+  const [assignments, setAssignments] = useState<LMSAssignmentAPIResponse[]>([])
+
+  const fetchOrgIntegrationsAsync = useCallback(async () => {
+    const response = await API.organizations.getIntegrations(organizationId)
+    setLmsIntegrations(response)
+  }, [organizationId])
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:67]: might want to add an error check, something like

```
await API.organizations.getIntegrations(organizationId)
      .then((response) => {
        setLmsIntegrations(response)
      })
      .catch((error) => {
        const errorMessage = getErrorMessage(error)
        message.error(errorMessage)
      })
```

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -0,0 +1,398 @@
...
+
+  const fetchDataAsync = useCallback(async () => {
+    const response = await API.course.getIntegration(courseId)
+    setLmsIntegration(response)
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:71]: same goes for this

AdamFipke:
okay so I guess it's a little hard for me to manually test the things, so I'll just kinda look over the code

packages/frontend/app/api/index.ts
```diff
@@ -114,6 +119,7 @@ class APIClient {
   ): Promise<T> {
     const res = (await this.axios.request({ method, url, data: body, params }))
       .data
+    if (res == '') return undefined as T
```
AdamFipke, line 122: may I ask why this change was made? Like was there a case with an endpoint that needed to return undefined but instead was giving an empty string?

Also I am a little bit concerned that there may be some pieces of frontend that assume that the response is an empty string and not undefined

bhunt02: cannot exactly remember? but i believe it was because there was strange behavior where 'undefined' responses were being returned as empty strings rather than undefined, and I don't believe this would have impacted other parts of the frontend but if you have an alternative/would like to check i'd make changes as needed

AdamFipke (changes requested):
okay yeah just a few small things (see comments). I don't really have a way of testing to make sure all the new things work, but I don't see much that would break existing stuff (and it won't be *critical* if this breaks) so that should be fine. 

Overall nice work, code looks pretty slick

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx]: should be addressed in latest commit

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx]: should be addressed in latest commit

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:219]: should be addressed in latest commit

packages/server/test/lmsIntegration.integration.ts
bhunt02: should be addressed in latest commit

packages/frontend/app/api/index.ts
AdamFipke: Hmm I see, I guess one thought is maybe why does the endpoint need to return undefined and could it instead return something else? Or maybe another thought is adding a new optional parameter to `req` called something like `turnEmptyResponseToUndefined`. It's not really an *ideal* solution but it's probably the best way to make it work without possibly accidentally introducing new bugs

bhunt02: reversed... i think maybe a review of the API requests to see if the behavior of expecting an empty string is ever utilized, but I altered my implementation by using the property of empty strings, undefined, null etc. all being 'falsy' in ECMA script to accommodate this

AdamFipke (approved):
sweet, looks good!

patrickma6199 (approved):
Looks good! I like how you followed the paradigm for the adapter as well.

PR #189, opened 2024-11-25
Queue Chats

### LINKED ISSUE #122: Queue Chat
**Description**

Add a 2-way messaging method inside the queues.

Why 2-way? It provides the most flexibility. For example:
```
TA: Hey can you join the zoom meeting?
Student: I can't find the link
TA: Check the top left of the page
```

These messages probably don't need to be stored on the backend, and could probably just be kept in the browser's local storage. These messages could probably just clear once they leave the page.

My suggestion is using the MQTT protocol and set up our server as a simple message broker (though it depends if there are any good javascript packages for this). MQTT is basically just a really simple pub/sub model (you can pub/sub to a 'room'). You would just need to make sure that only verified users can pub/sub to these chats. Using websockets with Server Side Events could also be an option, though this will likely require more work. 

Sounds of getting a new message is also a good idea. Though students should only receive sounds from TA's messages.

For desktop, this chat window could be just under the staff list (probably just flex-grow it so it takes up all remaining space, but no less than say 10rem).
A challenge is going to be figuring out how this looks on mobile (you will probably need to hide it somewhere, and maybe use antd's Alert component + sound to try to bring their attention there).

patrickma6199:
# Note: This will break current integration tests due to a change in setupIntegrationTest()'s return type. It is a mere one line change, but any integration tests that are merged after this will need the slight refactor.

# Description

This feature was mainly motivated by the use cases where staff and students wish to contact each other but cannot prior to joining the Zoom call or meeting. This could range from being able to communicate about technical issues to why one might be late to a meeting. Queue chats also enable minor issues such as assignment detail questions to be handled via the chat instead of a Zoom call or physical meetings adding to the means of resource retrieval for students.

In the queue page, the chatbot button has been removed and replaced with the queue chat (regardless of queue location type). When a question is set to the "Helping" status, the queue chat automatically opens for both parties and closes automatically when the student's question is resolved. If two or more students are being helped at the same time, each student will get a private chat window with the staff member. When chats end, dated summary statistics and chat metadata are saved in the database for potential future investigation.

Closes #122 

New Dependencies: redis-memory-server (to mock redis behaviour for integration tests by mocking Redis module)

## Type of change

- [X] This change requires a database query to update old data on production. This is easily handled with included migrations.

# How Has This Been Tested?

- [x] Manual
  - [x] Open two browser windows to emulate the staff member and a student. Test that when the student's question is being helped:
    - [x] Queue chat appears opened automatically
    - [x] Queue chat is visible and interactable even when Join Zoom Modal is opened
    - [x] Chat updates with messages even when either window is not focused by the OS
    - [x] Opening a new browser window (might have to be another browser entirely) to log into another student account and check that two private chats are opened automatically when the new student is also helped (before the first student is finished).
- [x] Automated
  - [x] Unit Test for Queue Chats Service in backend
  - [x] Integration test for Queue Chats Controller routes in backend

# Checklist:


packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueQuestions.tsx
```diff
@@ -22,6 +22,7 @@ import { QuestionTagElement } from '../../../components/QuestionTagElement'
 import QuestionCard from './QuestionCard'
 import JoinTagGroupButton from './JoinTagGroupButton'
 import { useCallback } from 'react'
+import { DeliveredProcedureOutlined } from '@ant-design/icons'
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueQuestions.tsx:25]: unused import

packages/frontend/app/(dashboard)/layout.tsx
```diff
@@ -62,27 +62,33 @@ const Layout: React.FC<LayoutProps> = ({ children }) => {
...
-            <StandardPageContainer className="flex-grow">
-              {children}
-            </StandardPageContainer>
-          )}
-        </ChatbotContextProvider>
+        {URLSegments[3] === 'queue' ? (
+          <StandardPageContainer className="flex-grow">
+            {children}
+          </StandardPageContainer>
+        ) : (
```
AdamFipke, line 69: I really think the chatbot should still be visible inside the queue page since the usecase is "while you're waiting, you can use the chatbot without needing to switch pages" and I've also seen a few of my students use it here. 

It would only probably be fine to hide the chatbot while the student is *currently being helped* (though still not ideal, since maybe the student asks the chatbot something while waiting and wanted to bring it up during discussion with TA. I would say only hide the chatbot unless it's really hard to make the UI work with it

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const [input, setInput] = useState<string>('')
+  const [isLoading, setIsLoading] = useState<boolean>(false)
+  const { queueChatData, mutateQueueChat, hasNewMessages } = useQueueChat(
+    queueId,
+    studentId,
+  )
+
+  useEffect(() => {
+    if (hasNewMessages) {
+      setIsOpen(true)
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:34]: I think having it auto-open is a good idea, just don't have it auto-open for TAs on mobile since this chat modal can take up a big part of the screen.

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
```diff
@@ -0,0 +1,220 @@
...
+import { CloseOutlined } from '@ant-design/icons'
+
+interface QueueChatProps {
+  role: Role
+  queueId: number
+  studentId: number
+  fixed?: boolean
+}
+
+const QueueChat: React.FC<QueueChatProps> = ({
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:18]: image

I noticed that when you're helping multiple students on mobile it looks a little cutoff.

I guess you would need to maybe do a flex-row-wrap for the buttons. And then for the chat itself, you would need to restrict the user so that only 1 chat is open at a time on mobile.

Also to save a little bit of space with the button, maybe just have the first name of the student on mobile.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -275,238 +275,251 @@ const Chatbot: React.FC<ChatbotProps> = ({
         )}
       >
         {isOpen ? (
-          <Card
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:278]: also idk if there was just a git moment (like with merges or something) or what but it appears as though there's a lot of changed lines to the chatbot here and I was curious about what changes were needed here and why

packages/server/src/queue/queue-sse.service.ts
```diff
@@ -44,20 +44,27 @@ export class QueueSSEService {
...
+    );
   }
 
-  private throttleUpdate(updateFunction: (queueId: number) => Promise<void>) {
+  private throttleUpdate(
+    updateFunction: (queueId: number, studentId?: number) => Promise<void>,
+  ) {
     return throttle(
-      async (queueId: number) => {
+      async (queueId: number, studentId?: number) => {
```
AdamFipke, line 67: was just curious on why the changes here were made? Since some of it looks like it may be leftover changes?

packages/server/src/queueChats/queue-chats.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  @ManyToOne(() => UserModel, (student) => student.studentChats, {
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({ name: 'studentId' })
+  @Exclude()
+  student: UserModel;
+
+  @Column({ nullable: true })
+  studentId: number;
```
AdamFipke, line 47: I saw this and got confused that this was like a university studentId. Maybe a better variable name could be studentUserId?

packages/server/src/queueChats/queue-chats.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  @ManyToOne(() => UserModel, (staff) => staff.staffChats, {
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({ name: 'staffId' })
+  @Exclude()
+  staff: UserModel;
+
+  @Column({ nullable: true })
+  staffId: number;
```
AdamFipke, line 37: Also, are we sure we want to leave this nullable? Since I'm unsure if there's a case where it would be null unless a user was deleted, but in that case it would probably be wise to delete the queue chats associated with it (which it looks like you have already done)

And I suppose that also applies to queueId and studentId

packages/server/test/queueChats.integration.unfinished.ts
```diff
@@ -0,0 +1,205 @@
...
+import {
+  CourseFactory,
+  QueueFactory,
+  UserCourseFactory,
+  UserFactory,
+} from './util/factories';
+import { Role } from '@koh/common';
+import MockRedisClient from './util/mockedRedisClient';
+
+// PAT TODO: Figure out how to mock redis properly
```
AdamFipke, line 12: yeah this is kinda a hard problem

maybe it would be worth looking into creating a "test" redis container that gets used for the tests? It might require some effort but at least in the future if we would ever add new features that utilized redis and want to write tests for it it would be a lot easier.

And yeah I know about the whole "we're not supposed to test other technologies and are just supposed to assume they work" thing but tbh I think that's kinda dumb since one of the main benefits to tests is regression testing (make changes and see if any tests fail) and being able to say update a framework or redis package or something and see what things broke right away is really nice and mocking the technology ruins that potential (and this was just things I thought about while I was trying to write integration tests for things that utilize the cron package).

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:18]: Also it's a little hard to notice in the screenshot, but the little outlines and numbers for the TaskMarkingSelector is showing on top of the queue chats. The old chatbot window did not have this problem.

It *might* be just as simple as increasing the z-index, but I imagine instead it will likely involve you needing to reorder the DOM so the queue chat is higher in the DOM

AdamFipke (changes requested):
see comments

packages/frontend/app/(dashboard)/layout.tsx
patrickma6199: We discussed this in the meeting already before the break if you remember and we all agreed that this was too specific a use case to spend any time on. If a user really wants to use the chat bot while waiting in a queue, they can simply duplicate the tab and do so in a couple clicks.

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:34]: Oh, I'll check this out, thanks for pointing it out.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:278]: There was a redundant import and some other styling issues that I found (the bottom margin of the chat button was on top of the footer slightly). I hope I didn't mess anything up but when I was testing it before, nothing was wrong as far as I could tell. Could you maybe point out what's happening when you load up the chatbot locally for you?

packages/server/src/queueChats/queue-chats.entity.ts
patrickma6199: I followed the paradigms present in almost all of our entities which all seem to have nullable columns where it might not actually be nullable in practice. I assumed there was some reason for this configuration so this is what I did as well for this entity. I could go back and change it but it doesn't introduce any vulnerabilities by being a nullable column just the same as the other nullable columns in other entities don't break the system either so it seems a bit redundant to spend time on by this point, unless we want to also change this for all of our other entities.

packages/server/src/queueChats/queue-chats.entity.ts
patrickma6199: I'm not sure if this is necessary as I would have to delete the old migration and regenerate a new one. I think it might be easier to simply add a comment here for clarity. I feel like intuitively, after seeing the staffId attribute, it should be fairly clear that both user ids are foreign keys to the user table.

packages/frontend/app/(dashboard)/layout.tsx
AdamFipke: I think maybe we should have another discussion about this then, since I don't recall that being the outcome of that discussion. 

Also, some students will never see any part of the system outside of the queue page (for example, they click on a queue invite and get sent to the queue page, and then they see that there's a course chatbot that they can use that might be able to just answer their question right away. I can also see this as a very common use case if profs share a queue invite to their students).

packages/server/src/queueChats/queue-chats.entity.ts
AdamFipke: It's just a maintenance/future-proofing thing so that future developers don't write a bad SQL query that leaves some of those fields null and causes the system to break because we were lacking nullish checks. It's a small thing

packages/server/src/queue/queue-sse.service.ts
patrickma6199: Yea, i migrated the queue chat sse into its own service but since it was originally in this file, I left some changes there. I'll clean it up since you're right, nothing there really needs to change.

packages/server/src/queueChats/queue-chats.entity.ts
patrickma6199: Yea, but maybe we should add a general maintenance issue to change this for all entities because it is in a lot of entities that didn't need it.

packages/frontend/app/(dashboard)/layout.tsx
patrickma6199: Yea, we can discuss it tomorrow with the team since we're meeting anyways!

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
```diff
@@ -210,9 +220,11 @@ const QueueChat: React.FC<QueueChatProps> = ({
         icon={<MessageCircleMore />}
         onClick={() => setIsOpen(true)}
       >
-        {isStaff
-          ? `${queueChatData!.student.firstName} ${queueChatData!.student.lastName}`
-          : `Queue Chat`}
+        {queueChatData && queueChatData.staff && queueChatData.student
+          ? isStaff
+            ? `${queueChatData!.student.firstName} ${queueChatData!.student.lastName}`
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:225]: oh I didn't notice this before but I should probably mention that not all of our users have lastNames (some have null, somehow (I think it can happen after login with google)). You may have already handled this somewhere else in the code but I thought I would mention this just in case

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/TaskMarkingSelector.tsx
```diff
@@ -131,7 +131,7 @@ const TaskMarkingSelector: React.FC<TaskMarkingSelectorProps> = ({
       aria-label={ariaLabel}
       aria-labelledby={ariaLabelledBy}
       style={{
-        zIndex: 1,
+        zIndex: 0,
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/TaskMarkingSelector.tsx:134]: huh I thought this would break the green rotating thing behind the task but it actually seems fine

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:248]: Oh, I wasn't aware. If it is sent as null in the response, I can easily fix that but I will test for this with a sample user.

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
```diff
@@ -0,0 +1,296 @@
...
+      <Alert
+        style={{ zIndex: 1050 }}
+        className={`${fixed ? 'fixed ' : ''} bottom-0 right-0 box-border overflow-y-auto md:bottom-8 md:right-2`}
+        message={`Chat data is not available.`}
+        description="Please try again later or contact support if the issue persists."
+        type="warning"
+        showIcon
+      />
+    )
+  }
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:102]: visual issue on desktop:
image

packages/server/src/question/question.service.ts
```diff
@@ -89,7 +79,58 @@ export class QuestionService {
...
+          user,
+        );
+      }
+    } catch (err) {
+      if (err instanceof Error) {
+        console.error(err.stack);
+      } else {
+        console.error(err);
+      }
+      throw new HttpException(
```
bhunt02, line 116: Also, why are these all 404s?

packages/server/src/question/question.service.ts
```diff
@@ -89,7 +79,58 @@ export class QuestionService {
     const isBecomingWaiting =
       !waitingStatuses.includes(oldStatus) &&
       waitingStatuses.includes(newStatus);
+    const isResolving = newStatus === ClosedQuestionStatus.Resolved;
+    const isFirstHelped = isBecomingHelped && question.firstHelpedAt;
```
bhunt02, line 83:
```suggestion
    const isFirstHelped = isBecomingHelped && !question.firstHelpedAt;
```

packages/server/src/queueChats/queue-chats.service.ts
```diff
@@ -0,0 +1,223 @@
...
+    staff: UserModel,
+    student: UserModel,
+
+    startedAt?: Date,
+  ): Promise<void> {
+    const key = `${ChatMetadataRedisKey}:${queueId}:${student.id}`;
+
+    // Remove any existing chat metadata and messages (in case of mismanagement; to protect previous chat history)
+    await this.redis.del(key);
+    await this.redis.del(`${ChatMessageRedisKey}:${queueId}:${student.id}`);
```
bhunt02, line 43: is there a reason this is present twice?

bhunt02 (commented):
can't say much, getting consistent 404's from redis even after running `yarn install`, I can try to give a better review later but I've left a few suggestions/changes nonetheless

packages/server/src/queueChats/queue-chats.service.ts
patrickma6199: Yea, because I separated the messages from the metadata.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
```diff
@@ -154,6 +170,20 @@ export default function QueuePage({ params }: QueuePageProps): ReactElement {
     }
   }, [clickedZoomModal, studentQuestion])
 
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:174]: this typeof window shouldn't be necessary since it's a client component?

bhunt02 (approved):
everything's working now! not having the initial issues I was having before and honestly works really good, glad this feature is being added!

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
```diff
@@ -0,0 +1,296 @@
...
+      if (questionId) {
+        API.queueChats.sendMessage(queueId, questionId, input).then(() => {
+          mutateQueueChat()
+          setIsLoading(false)
+          setInput('')
+        })
+      }
+    } catch (error) {
+      console.error(error)
+    }
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:88]: instead of just printing to console, maybe send a message to the user too. Also setIsLoading(false) should be in finally (or after catch), otherwise it will load forever if there's an error.
```suggestion
      if (questionId) {
        API.queueChats.sendMessage(queueId, questionId, input).then(() => {
          mutateQueueChat()
          setInput('')
        }).catch((e) => {
        message.error(getErrorMessage(e))
        }).finally(() => {
        setIsLoading(false)
        })
      }
```

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
```diff
@@ -0,0 +1,296 @@
...
+          <div className="no-scrollbar max-h-[50vh] overflow-y-auto">
+            <div className="mb-2 w-full pt-1 text-center text-xs italic text-gray-500">
+              Your chat messages will not be recorded for your privacy but
+              please remain respectful. Chat messages will remain until your
+              question is resolved.
+            </div>
+            {queueChatData!.messages &&
+              queueChatData!.messages.map((message, index) => {
+                return (
+                  <Fragment key={index}>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:144]: if the messages have an id, it is better to make the key this id.

This would matter if for some reason the messages' indexes were to change order (it would cause react to re-render all of the message components again rather than re-using them)

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -539,15 +539,19 @@ const Chatbot: React.FC<ChatbotProps> = ({
...
-          <Button
-            type="primary"
-            icon={<RobotOutlined />}
-            size="large"
-            className="mx-5 rounded-sm"
-            onClick={() => setIsOpen(true)}
-          >
-            Chat now!
-          </Button>
+          <>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:542]: unused fragment

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -297,7 +297,7 @@ const Chatbot: React.FC<ChatbotProps> = ({
       <div
         className={cn(
           variant === 'small'
-            ? 'fixed bottom-5 z-50 max-h-[90vh] w-screen md:right-5 md:max-w-[400px]'
+            ? 'fixed bottom-8 z-50 max-h-[90vh] w-screen md:right-1 md:max-w-[400px]'
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:300]: ~~bottom-8 is going to make this appear too high on mobile, probably better to do md:bottom-8 bottom-5~~ Actually I tested it out and bottom-8 looks far too high on pages with a scroll bar. Yes bottom-5 will overlap with the footer a bit, but trust me that's a far better tradeoff then having this button more likely to overlap page content.
image
image

Also this was on one of my TODOs, but i'll just add it here since it's easy:

`md:right-1 lg:right-3 xl:right-5 2xl:right-7`

idea being that for larger and larger screens it won't be too far right, but for smaller screens it won't overlap important info as much. If you want to toss this in that would be sweet (it would merge conflict otherwise)

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
```diff
@@ -912,6 +940,111 @@ export default function QueuePage({ params }: QueuePageProps): ReactElement {
...
+                        role={role}
+                        isMobile={isMobile}
+                        fixed={false}
+                        hidden={false}
+                      />
+                    )
+                  })}
+                </div>
+              </div>
+            )}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:1047]: ...and the queue page gets even more bloated and complex 😭 

kidding lol, i think it's just inevitable with a page so complex, but it is kinda comical of how massive it is haha

packages/frontend/app/(dashboard)/layout.tsx
```diff
@@ -62,27 +62,34 @@ const Layout: React.FC<LayoutProps> = ({ children }) => {
...
-          {URLSegments[4] === 'edit_questions' ||
-          URLSegments[4] === 'chatbot_questions' ? (
-            <div className="p-1">{children}</div>
-          ) : (
-            <StandardPageContainer className="flex-grow">
-              {children}
-            </StandardPageContainer>
-          )}
-        </ChatbotContextProvider>
+        {/* If on the queue page, do not display the chatbot */}
```
AdamFipke, line 65: Oh yeah so we already discussed this, but I guess just to state it here, I will probably be revisiting this soon to fit the chatbot in there non-intrusively (I have some ideas). I won't start any work yet in case you're still making some changes, but once you feel ready i'll come and give it a stab

packages/frontend/app/components/ChangeLogModal.tsx
```diff
@@ -66,9 +66,9 @@ const ChangeLogModal: React.FC<ChangeLogModalProps> = ({
       classNames={{
         header: 'flex items-center justify-center',
       }}
-      className="box-border flex flex-col items-center justify-center overflow-y-auto"
+      className="box-border flex flex-col items-center justify-center"
     >
-      <div className="childrenMarkdownFormatted box-border h-[75vh] w-full overflow-y-auto px-4">
+      <div className="childrenMarkdownFormatted box-border h-[60vh] w-full overflow-y-auto px-4">
```
AdamFipke, line 71: I think it's easier to read if it takes up more of the page
```suggestion
      <div className="childrenMarkdownFormatted box-border h-[75vh] w-full overflow-y-auto px-4">
```

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:143]: Yeah, I agree, but since the messages are temporary, they don't have ids. The good thing is messages can't be deleted (since its temporary anyways) so the order will never change.

packages/frontend/app/components/ChangeLogModal.tsx
patrickma6199: I made this change because this:

image

was how the mobile view looked. I feel like it being somewhat off the center didn't look great.

This is how the mobile view looks with 60vh.

image

I agree that it looks better being centered in the space below the header bar tho for desktop so ill just add a media query.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:300]: You might have some old css cached cause I also moved the button much closer to the side:

image

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:542]: whoopsies

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:1047]: Yeah, I was actually thinking of refactoring the page slightly just to get the column component defined at the top moved into its own file with its required states. I'm sure there are more changes to be made though to slim it out.

packages/server/src/question/question.controller.ts
```diff
@@ -60,6 +62,7 @@ export class QuestionController {
     private questionService: QuestionService,
     private queueService: QueueService,
     private redisQueueService: RedisQueueService,
+    private QueueChatService: QueueChatService,
```
AdamFipke, line 65: I notice this queuechatservice was added here but it seems to be unused?

packages/server/src/question/question.service.ts
```diff
@@ -137,6 +141,35 @@ export class QuestionService {
...
+        // Don't save chat metadata in database
+        await this.queueChatService.clearChat(question.queueId, question.id);
+      } else if (isFirstHelped) {
+        // Create chat metadata in Redis
+        await this.queueChatService.createChat(
+          question.queueId,
+          question.taHelped,
+          question,
+        );
+      }
```
AdamFipke, line 160:
```suggestion
      if (isResolving) {
        // Save chat metadata in database (if messages were exchanged)
        await this.queueChatService.endChat(question.queueId, question.id);
      } else if (newStatus in ClosedQuestionStatus) {
        // Don't save chat metadata in database
        await this.queueChatService.clearChat(question.queueId, question.id);
      } else if (isFirstHelped) {
        // Create chat metadata in Redis
        await this.queueChatService.createChat(
          question.queueId,
          question.taHelped,
          question,
        );
      }
```

Hmmm, I realise that there might be a case where a question gets deleted from limbo status or something and it ends up maybe not clearing the chat. So I'm thinking it might be better to just check if newStatus is in ClosedQuestionStatus

packages/server/src/queueChats/queue-chats-sse.service.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  private throttleUpdate(
+    updateFunction: (queueId: number, questionId: number) => Promise<void>,
+  ) {
+    return throttle(
+      async (queueId: number, questionId?: number) => {
+        try {
+          await updateFunction(queueId, questionId);
+        } catch (e) {}
```
AdamFipke, line 78: Just curious, so this catch statement pretty much catch any errors and then do nothing with them, would this catch any important errors like redis or http errors?

AdamFipke: Also I see that this limits it to 1 message per second i think? 
Will it throw any errors to the user if they post too quickly?

packages/server/src/queueChats/queue-chats.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+@Controller('queueChats')
+@UseGuards(JwtAuthGuard)
+export class QueueChatController {
+  constructor(
+    private queueChatService: QueueChatService,
+    private queueChatSSEService: QueueChatSSEService,
+  ) {}
+
+  @Get(':queueId/:questionId')
+  @UseGuards(JwtAuthGuard)
```
AdamFipke, line 31:
```suggestion
```

duplicate guard since the controller already has JwtAuthGuard

packages/server/src/queueChats/queue-chats.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+    );
+    if (!chatData) {
+      throw new HttpException(
+        ERROR_MESSAGES.queueChatsController.chatNotFound,
+        HttpStatus.NOT_FOUND,
+      );
+    }
+
+    await this.queueChatService
+      .checkPermissions(queueId, questionId, user.id)
```
AdamFipke, line 49:
```suggestion
    @UserId() userId: number,
  ) {
    const chatData = await this.queueChatService.getChatData(
      queueId,
      questionId,
    );
    if (!chatData) {
      throw new HttpException(
        ERROR_MESSAGES.queueChatsController.chatNotFound,
        HttpStatus.NOT_FOUND,
      );
    }

    await this.queueChatService
      .checkPermissions(queueId, questionId, userId)
```

packages/server/src/queueChats/queue-chats.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+      user.id,
+    );
+    if (!allowedToSend) {
+      throw new HttpException(
+        ERROR_MESSAGES.queueChatsController.sendNotAuthorized,
+        HttpStatus.FORBIDDEN,
+      );
+    }
+    try {
+      const isStaff = user.id === metadata.staff.id;
```
AdamFipke, line 122:
```suggestion
    @UserId() userId: number,
    @Body('message') message: string,
  ) {
    const metadata = await this.queueChatService.getChatMetadata(
      queueId,
      questionId,
    );
    if (!metadata) {
      throw new HttpException(
        ERROR_MESSAGES.queueChatsController.chatNotFound,
        HttpStatus.NOT_FOUND,
      );
    }

    const allowedToSend = await this.queueChatService.checkPermissions(
      queueId,
      questionId,
      userId,
    );
    if (!allowedToSend) {
      throw new HttpException(
        ERROR_MESSAGES.queueChatsController.sendNotAuthorized,
        HttpStatus.FORBIDDEN,
      );
    }
    try {
      const isStaff = userId === metadata.staff.id;
```

packages/server/src/queueChats/queue-chats.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+import { QueueChatService } from './queue-chats.service';
+import { User, UserId } from 'decorators/user.decorator';
+import { UserModel } from 'profile/user.entity';
+import { Response } from 'express-serve-static-core';
+import { QueueRole } from 'decorators/queue-role.decorator';
+import { ERROR_MESSAGES, Role } from '@koh/common';
+import { QueueChatSSEService } from './queue-chats-sse.service';
+
+@Controller('queueChats')
+@UseGuards(JwtAuthGuard)
```
AdamFipke, line 23: One thought was to add QueueRolesGuard and add some `Roles` decorators to some of these methods so that only people in the course can call them, but upon looking at your methods more closely that might not be necessary thanks to your .checkPermissions checks so maybe what you have should be good. Maybe one thought is to add a comment about it anyway that these methods don't use any roles guards since they use .checkPermissions

packages/server/src/queueChats/queue-chats.service.spec.ts
```diff
@@ -0,0 +1,313 @@
...
+import { RedisService } from 'nestjs-redis';
+import { QueueChatService } from './queue-chats.service';
+import { QueueChatsModel } from './queue-chats.entity';
+import { Connection } from 'typeorm';
+import { TestConfigModule, TestTypeOrmModule } from '../../test/util/testUtils';
+import { ApplicationTestingConfigModule } from '../config/application_config.module';
+import { QuestionFactory, UserFactory } from '../../test/util/factories';
+import { TypeOrmModule } from '@nestjs/typeorm';
+
+jest.mock('nestjs-redis');
```
AdamFipke, line 11: huh neat, i didn't know you could mock whole packages in jest, good to know

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
```diff
@@ -0,0 +1,299 @@
...
+        })
+    }
+  }
+
+  if (!queueChatData || queueChatError) {
+    return (
+      <Alert
+        style={{ zIndex: 1050 }}
+        className={`${fixed ? 'fixed ' : ''} bottom-0 right-0 box-border overflow-y-auto`}
+        message={`Chat data is not available.`}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx:99]:
```suggestion
        message={`Chat with ${isStaff ? 'student' : 'TA'} is not available.`}
```

AdamFipke (approved):
Changes look pretty good! I manually tested the simple case of starting to help someone, sending messages, and finish helping on both desktop and mobile for staff and student.

The suggestions I have for the backend are pretty minor. 

And I guess all that's needed is to re-add the chatbot (will need to see how it looks like on desktop and mobile for staff and student (so 4 cases)), and once that's done it should be good to merge

packages/server/src/queueChats/queue-chats-sse.service.ts
AdamFipke: okay well from some testing, it certainly seems like you can spam messages so 🤷‍♂️

packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/QueueChat.tsx]: I ended up removing this alert and instead opted to just return an empty fragment if there was an error since the error was kinda large and intrusive, which might get distracting for students if they start getting helped but then just see a big warning message

packages/server/src/question/question.service.ts
AdamFipke: i am going to implement these changes real quick

PR #191, opened 2024-11-26
Backup Uploads Directory

patrickma6199:
# Description

This PR adds to the database backup service by including a new backup service function that compresses the upload directory every four days via a Cron job (just the same as the other backup implementations for the database). The whole uploads directory of the backend gets zipped together with ```gzip``` to save as much space as possible and gets stored in a subdirectory of the `./backups` directory.

I also added a script to manually run all backups in case we needed that in production at any point, and also to easily manually test that all the service backup functions behave as expected using one line in the console: `yarn tsx backup-manual.ts`.

New Dependencies:
- `tsx` for running `.ts` files without transpiling into JavaScript

Closes # 165

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] This change requires a documentation update
- [X] This requires a run of `yarn install`
- [ ] This change requires a database query to update old data on production.

# How Has This Been Tested?

- [ ] Manually
  - [x] Added a manual testing file which can be run on dev machine (does not have to be a Linux machine; although I haven't tried it on MacOS yet, I doubt there would be issues)
- [x] Automated
  - [x] Added unit test to test service function functionality to `backup.service.spec.ts`

# Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code where needed 
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] New and existing unit tests pass locally with my changes
- [X] Any work that this PR is dependent on has been merged into the main branch
- [X] Any UI changes have been checked to work on desktop, tablet, and mobile


packages/server/src/backup/backup.service.ts
```diff
@@ -101,6 +105,41 @@ export class BackupService {
     }
   }
 
+  // Daily Uploads Backup Task - Keeps rolling backups for 5 days
+  @Cron(CronExpression.EVERY_DAY_AT_MIDNIGHT) //tentative
```
AdamFipke, line 109: I would do every 3 days

packages/server/src/backup/backup.service.ts
```diff
@@ -101,6 +105,41 @@ export class BackupService {
...
+      const compressCommand = `tar -czf ${backupDir}/${backupFile} -C ${uploadsDir} .`;
+
+      exec(compressCommand, (error, stdout, stderr) => {
+        if (error) {
+          console.error(`Uploads backup failed: ${stderr}`);
+          Sentry.captureMessage(`Uploads backup failed: ${stderr}`);
+        } else {
+          console.log(`Uploads backup saved: ${backupFile}`);
+          // Retain only the last 5 backups
+          this.deleteOldBackups(backupDir, 5); //tentative
```
AdamFipke, line 134: and then keep 14 days of backups

packages/server/src/backup/backup-manual-test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/*
+ * Manual tests for the backup service that bypass the cron jobs.
+ *
+ * Command: yarn tsx backup-manual-test.ts
+ *
+ * Note: when using tsx, the relative paths are resolved from the root of this node project (packages/server).
+ */
+
+async function runManualTests() {
```
AdamFipke, line 11: wait why is this here? You should be able to call the backup services and bypass the cron jobs just fine?

AdamFipke (changes requested):
see comments

AdamFipke (changes requested):
see comments

packages/server/src/backup/backup-manual-test.ts
patrickma6199: Tsx was added so that we could run the backup-manual-test.ts without transpiling it to JS. I wrote this file up mainly because it just makes manually testing the service a one-line process. I feel that in the future, there might be a use case where we might want to ensure that the backup services are still working in deployment or want to bypass cron in one line. If you really feel that this addition is unnecessary, I can remove it. I just think that it does no harm to have it just as an extra tool on our belt for interacting with the backup service.

packages/server/src/backup/backup.service.ts
patrickma6199: Neither 3, 4, or 5 day gaps between backups would be divisible if I kept 14 days of backups. I'll opt for every 4 days, and I'll keep the last 3 backups (12 days). Let me know if this sounds good to you.

AdamFipke: sure yeah 12 days should be fine

packages/server/src/backup/backup-manual-test.ts
AdamFipke: Ah I see, i guess the use case for wanting to manually run backups makes sense. 

Though just a question, when we run these manual tests, does it actually save backups? Since it may be wise to rename to not confuse it with our actual tests, maybe just call it runManualBackups and name the file backup-manual.ts kinda thing

patrickma6199: That makes sense. I'll change the filename in a second.

patrickma6199: I changed the wording on this filename and the code inside so it should be clearer now. I also updated the BACKUPS.md to reflect the new backup cycle for uploads. Everything should be good now.

AdamFipke (approved):
sweet looks good

PR #193, opened 2024-12-03
192 Prompt students to leave queue when last TA checks out

### LINKED ISSUE #192: Clear queue when last TA checks out
**Description**

So, problem right now is when TAs check out, questions can still sit in the queue, sometimes for days, which messes up the waitTime data.

The simple option is to just clear the queue if the last TA checks out. However, that will not work because of this example:
- TA1 is checked in from 9-11
- TA2 is supposed to be checked in from 11-12, but is 5mins late
Now, if TA1 checks out right at 11, it will delete all students questions, even though they should really carry over.

Thus, here's the idea:
"The last TA just checked out. Would you like to leave the queue or stay?" <- Later, we could add an option to convert to email #69 
Then, if the student does not respond in 10 minutes, mark the question as stale.
If the student chooses to stay, after say 20mins, check if there's any staff. If there isn't, re-prompt the student.

Additionally, if a TA checks in, all those alerts will need to be resolved and cron jobs deleted.

This is kinda a more advanced version of #180

AdamFipke:
# Description

Prompts students to leave the queue when the last TA checks out.

I also took this a step further to cover all of the following cases:
- student clicks "leave queue" on prompt
- student clicks "leave queue" on prompt while having questions in multiple queues
- student does not respond to prompt -> question's status becomes LeftDueToNoStaff after 10mins
- student clicks "i'll stay" 
	- check in 10mins, are there any staff?
		- no staff -> prompt student to leave queue again (basically creating a loop every 10-20mins as long as the there are no staff and the student keeps clicking "i'll stay")
- student doesn't click anything and the TA checks back in
- Tested so it only works when last TA checks out, and it also works when the same TA is checked into multiple queues
- Made sure to clear all cron jobs with the same name beforehand so there will be no errors from conflicts
- Also made sure all the cron jobs are expendable/transient and won't cause anything to break if they get destroyed (e.g. if the server shuts down)

Closes #192 

## Type of change

# How Has This Been Tested?

I have declared that cron jobs cannot really be tested (at least for integration tests). This is because any cron jobs that get activated using jest's fakeTimers will not run any code (or at least won't run anything like a DB query). The solution to fix this is to make the code inside the cron jobs just call specific functions, and then just test those functions.

Though, the unfortunate part with this strategy is it doesn't allow you to create elaborate integration tests to see how all the functions work together with the cron jobs. Also it requires a TON of mocking, which I'm never really a big fan of since mocking a function will mean that if that function changes critically (e.g. a major version change), the test will still pass.

Anyways rant over. I wrote a bunch of unit tests to test what I could basically. The rest had to be manual (and I did a LOT of manual testing).

# Checklist:


packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
```diff
@@ -213,12 +215,6 @@ export default function QueuePage({ params }: QueuePageProps): ReactElement {
     }
   }, [queueConfig, configTasks, isStaff, queueQuestions, questionTypes])
 
-  const staffCheckedIntoAnotherQueue = course?.queues?.some(
-    (q) =>
-      q.id !== qid &&
-      q.staffList.some((staffMember) => staffMember.id === userInfo.id),
-  )
-
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:221]: oh yeah as a note I'm removing the restriction that prevented staff from checking into multiple queues since nothing seems to break and it allows more flexibility with the system (in case for some reason some prof wants to set it up with multiple queues at the same time).

packages/server/src/queue/queue.module.ts
```diff
@@ -10,19 +9,36 @@ import { AlertsModule } from '../alerts/alerts.module';
...
+    AlertsService,
+    QuestionService,
+    QueueService,
+    RedisQueueService,
+  ],
+  imports: [
+    ApplicationConfigService,
+    SSEModule,
+    AlertsModule,
+    forwardRef(() => QuestionModule),
```
AdamFipke, line 41: oh yeah so there's some fun shit here.

As a result of this PR, there is an unavoidable nest.js circular dependency between the queue module and question module (I have tried everything I can think of to work around this to get rid of this circular dependency but I had no luck, all that ended up happening was making the circular dependency larger and across more files).

I'm not going to get into what caused it exactly since it was kinda inevitable (mostly since you need to import the queue service to call getQuestions to update redis, and that happens in the queue-clean service and question service, and then when the queue module depends on either queue-clean service or question service, bam circular dependency).

The good news is nest.js knows about this and has tools for fixing it (forwardRef), and so it actually doesn't turn out to be too much of an issue.

packages/server/src/queue/queue.module.ts
```diff
@@ -10,19 +9,36 @@ import { AlertsModule } from '../alerts/alerts.module';
...
+  ],
+  exports: [
+    QueueSSEService,
+    QueueCleanService,
+    ApplicationConfigService,
+    AlertsService,
+    QuestionService,
+    QueueService,
+    RedisQueueService,
+  ],
```
AdamFipke, line 36: oh yeah in case you're wondering what `exports` does, from my understanding it will export all of these services to whatever imports the queue module so that they don't have to *also* import the service (I was reading some nest.js documentation to fix the circular dependency issue and decided to also research what providers, exports, and imports all do).

For instance, the RedisQueueService is something the queue module only really needs. However, if you go to import the queue module into another module, like the question module, you will need to re-import the RedisQueueService into the question module too since the imported queue module depends on it. And simply by having the queue module re-export the RedisQueueService, you don't need to import it into question service.

It may also help with performance or something (something like sharing instances of a service or something) but I forget.

so:
- providers: an outside service that the module uses (i.e. an import in queue.service or queue.controller)
- imports: an outside module that contains the outside service that you want in your provider
- exports: services that can be accessed when an outside module imports this module

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:221]: there's freedom, and then there's unintended behavior; I think checking into multiple queues might be unintended behavior. did you test manually how the system behaves when checked into multiple queues?

packages/server/src/question/question.controller.ts
```diff
@@ -16,9 +16,11 @@ import {
   Body,
   ClassSerializerInterceptor,
   Controller,
+  forwardRef,
```
bhunt02, line 19: maybe remove this import and other references since it's commented out

packages/server/src/queue/queue-clean/queue-clean.service.ts
```diff
@@ -104,5 +117,246 @@ export class QueueCleanService {
...
+      this.schedulerRegistry.addCronJob(jobName, job);
+      job.start();
+    }
+  }
+
+  public deleteLeaveQueueCronJobForStudent(
+    queueId: number,
+    studentId: number,
+  ): void {
+    const jobs = this.schedulerRegistry.getCronJobs();
```
bhunt02, line 322: not a code review complaint but this forced O(n) is so nasty

packages/server/src/queue/queue.module.ts
bhunt02: language!

bhunt02 (approved):
logic looks pretty solid here, i have no real complaints just some stuff I noticed, good job Adam!

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:221]: yep, I tested everything I could think of and I couldn't find any issues.

packages/server/src/queue/queue-clean/queue-clean.service.ts
AdamFipke: yeah it's really unfortunate. Luckily this function runs very infrequently and also the number of cron jobs is not many (probably <100)

packages/server/src/queue/queue.module.ts
AdamFipke: ;P

PR #194, opened 2024-12-03
Display Unread Async Questions Count

patrickma6199:
# Description

This PR adds an "Unread Async Questions" count to the Async Questions card of the course dashboard to give some simple feedback to users and hopefully incentivize them to take a look at the async questions center more when there are any updates. Hopefully, this improves the reach of this feature for students that had never thought to interact with async questions center much.

image

Closes # 91

## Type of change

- [X] This change requires a database query to update old data on production. (but covered in migration)

# How Has This Been Tested?

- [x] Manual Testing
  - [x] Build the application in your development environment
  - [x] Use a demo student user to make a new async question
  - [x] The Async Questions Card component renders with a red unread counter on the top right for student accounts that didn't make the new async question
  - [x] Each user has their own unique unread questions count that updates to 0 when the async questions center is clicked into
- [x] Automated Testing
  - [x] Updated asyncQuestion integration test to check that counts incremented correctly for other users

# Checklist:


packages/server/src/course/course.controller.ts
```diff
@@ -1061,4 +1062,47 @@ export class CourseController {
...
+        courseId,
+      },
+    });
+
+    if (!userCourse) {
+      throw new NotFoundException('UserCourse not found');
+    }
+
+    return { count: userCourse.unreadAsyncQuestions };
+  }
```
AdamFipke, line 1084: instead of making it its own endpoint, have you considered going into profile.controller and change .get so that it also returns the unreadAsyncQuestions? It could save some API calls 

i.e.
```ts
 const courses = user.courses
      ? user.courses
          .filter((userCourse) => userCourse?.course?.enabled)
          .map((userCourse) => {
            return {
              course: {
                id: userCourse.courseId,
                name: userCourse.course.name,
		count: userCourse.unreadAsyncQuestions // add this line
              },
              role: userCourse.role,
            };
          })
      : [];
```

And then to get the undreadAsyncQuestions, you could do something similar to the getRoleInCourse function.

packages/server/src/profile/user-course.entity.ts
```diff
@@ -38,6 +38,9 @@ export class UserCourseModel extends BaseEntity {
   @Column({ nullable: true, default: false })
   expires: boolean;
 
+  @Column({ nullable: true, default: 0 })
```
AdamFipke, line 41: i'm not finished reviewing the code but i'm not sure if you want to leave this nullable, i think you might just want to leave it as a default of 0 with nullable false to ensure that all users have a number for it.

Also maybe you can try using a smaller number type to save some space (probably smallint)

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
```diff
@@ -24,7 +33,15 @@ const AsyncCentreCard: React.FC<AsyncCentreCardProps> = ({
...
-        title="Anytime Question Hub"
+        title={'Anytime Question Hub'}
+        extra={
+          unreadCount > 0 && (
+            <div className="mr-8 h-fit text-sm font-normal text-gray-200">
+              <span className="text-lg font-medium">{unreadCount}</span> Unread
+              Question{unreadCount > 1 ? 's' : ''}
+            </div>
+          )
+        }
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:44]: Huh neat. In my original vision for this, I saw it with some kinda colour (like green or maybe light blue) to grab the user's attention, but I realized how hard it would be to find a good colour for it and I actually am a big fan of what you did here. It's very visually consistent which helps since it's a busy page, but not *too* much so that it's still different enough from the queue cards. 

One small thing though is I think "X Unread Questions" is a little too many characters. Maybe something like "X New Questions" or even just "X New" or "X Unread"

Also, as another thought, maybe we can add a little green (1) indicator or something in the navbar for Anytime Qs

AdamFipke (changes requested):
see comments

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:44]: image

Oh yeah figured I should show a little screenshot here so if anyone else reviews this they can see what it looks like right now

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -173,6 +173,14 @@ export class asyncQuestionController {
         relations: ['creator', 'taHelped', 'votes'],
       });
 
+      // update read counts for all users in course
+      const query = await UserCourseModel.createQueryBuilder()
+        .update(UserCourseModel)
+        .set({ unreadAsyncQuestions: () => '"unreadAsyncQuestions" + 1' })
+        .where('courseId = :courseId', { courseId: cid })
+        .andWhere('userId != :userId', { userId: user.id }) // Exclude the question creator
+        .execute();
```
AdamFipke, line 182: Oh right, you might want to add a couple integration tests for async questions for this so that no one accidentally breaks it later

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:44]: Yeah, this was my original implementation a couple commits back but I switched it to try out the premade component from ant design. I agree tho that that looks more readable. I'm just wondering whether there's anything we can do maybe color-wise to make it stand out.

packages/server/src/profile/user-course.entity.ts
patrickma6199: Yeah, I just noticed that with many of our table definitions, attributes that shouldn't be nullable and have a default are still nullable, so I tried to follow what seemed to be the standard. Even the ***expires*** column above it's definition has nullable as an attribute despite that having only "true" or "false" is clearly sufficient for this case. I don't think it would do any harm but if you're worried, I can update the entity and migrations.

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:44]: Yeah, I'll shorten it so that the bolded numbers of all the cards are slightly more aligned.

packages/server/src/profile/user-course.entity.ts
AdamFipke: yeah there's a bunch of old bad database things that should probably be changed at some point. Like for example, user-course entity has a primary generated column and its courseId and userId are both nullable (meaning optional). This basically means that this user-course entity can exist on it's own when in reality its existence should depend on the user and the course (like if a user or course were to be deleted, the corresponding user-course entity will be deleted, and that any new user-course entities must have a real courseId and userId associated with them). Also I don't think `expires` is even used anymore lol (I think at one point the old khory system had a way to have temporary users and that's what that was for). But yeah it's just like a future maintainability thing so that some future developer doesn't misuse the field and make it null and cause the system to crash because there was a missing null check somewhere lol.

patrickma6199: Yeah no, I totally get you. I just think at this stage personally, its too hard to organize all these minute changes with all of the tasks I had hanging from the first term. I think we should instead just create a new urgent issue to refactor ALL of the table definitions we already have (including this one once the merge is done). That way, I can focus on closing these old PRs, and also, we can make one migration that alters all the tables instead of a bunch of migrations for each of the tables separately. I don't know if I'm explaining my thought process that well so let me know if you need more clarity from me haha.

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
patrickma6199: Will do, thanks for pointing this out!

packages/server/src/course/course.controller.ts
patrickma6199: I thought of this but thought that the response from that endpoint is already really big. Thinking about it now though, this would add hardly anything relative to the response and once I finish the profile caching issue I had, that would fix this fine. I'll refactor it, thanks for bringing it up.

packages/server/src/profile/user-course.entity.ts
bhunt02: > Yeah no, I totally get you. I just think at this stage personally, its too hard to organize all these minute changes with all of the tasks I had hanging from the first term. I think we should instead just create a new urgent issue to refactor ALL of the table definitions we already have (including this one once the merge is done). That way, I can focus on closing these old PRs, and also, we can make one migration that alters all the tables instead of a bunch of migrations for each of the tables separately. I don't know if I'm explaining my thought process that well so let me know if you need more clarity from me haha.

We haven't generally been making migrations regardless since it's set to update the database schema automatically to what is seen in the entities.

patrickma6199: Yeah, we should definitely not keep prod that way since rollbacks would be a pain. I'll just generate a new migration file then to avoid issues once/if prod switches out of "development mode".

patrickma6199: Which one was the second nullable you wanted me to remove? I don't think we should mess with the primary key acting attributed if they're nullable already since previous devs might have worked around it and I don't want to risk breaking anything. If you meant the expires field, I can remove it from there too.

bhunt02: > Which one was the second nullable you wanted me to remove? I don't think we should mess with the primary key acting attributed if they're nullable already since previous devs might have worked around it and I don't want to risk breaking anything. If you meant the expires field, I can remove it from there too.

There's no primary columns that are designated nullable, probably because that's impossible (relational integrity rules)
You can remove 'default' from expires because it has a defined default, but we can check the production database to see if there's any null values that currently exist (though I believe that may also be impossible)

Not sure what the other property listed here is

bhunt02 (commented):
I'm a bit uncomfortable with the implementation (there being no relation joining the async question + user course model with a boolean attribute to indicate if it's read), but I think this could work fine. I only worry that we'll have failed operations, but since it resets to zero rather than decrementing, it's unlikely it'll cause faults in the application.

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
```diff
@@ -13,6 +15,20 @@ const AsyncCentreCard: React.FC<AsyncCentreCardProps> = ({
...
+  useEffect(() => {
+    if (userInfo) {
+      const count = userInfo.courses.find(
+        (c) => c.course.id === cid,
+      )?.unreadCount
+      if (count) {
+        setUnreadCount(count)
+      }
+    }
+  }, [userInfo])
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:30]: this does not need to be in a useEffect since userInfo (and userInfo's count) does not change on this page.

Instead, I would just leave it as the following (not in any hooks)
```
const count = userInfo.courses.find(
        (c) => c.course.id === cid,
      )?.unreadCount
```

If you want, you can use a useMemo to cache the calculation, but seeing as how this is pretty much just a small for loop, doing a useMemo isn't going to improve performance much at all.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -101,6 +102,11 @@ export default function AsyncCentrePage({
     [sortBy],
   )
 
+  // This endpoint will be called to update unread count back to 0 when this page is entered
+  useEffect(() => {
+    API.course.updateUnreadAsyncCount(courseId)
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:107]: Good I like that there isn't an await here. 

One idea is to add a check so that it doesn't call the endpoint if it's already 0. Also, this should be updating the userInfo context otherwise it won't actually be updating the frontend when they go back to the course page (since userInfo context does not update automatically)

packages/server/src/profile/user-course.entity.ts
AdamFipke: > Which one was the second nullable you wanted me to remove? I don't think we should mess with the primary key acting attributed if they're nullable already since previous devs might have worked around it and I don't want to risk breaking anything. If you meant the expires field, I can remove it from there too.

Ah it was just the one field that needed to be changed for this PR, i think i was getting mixed up with another PR that had some nullable fields that shouldn't be nullable. But yeah looks like you made the change so this PR should be good in that regard

AdamFipke (changes requested):
Just a couple small frontend comments but other than that should be good

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:30]: > this does not need to be in a useEffect since userInfo (and userInfo's count) does not change on this page.
> 
> Instead, I would just leave it as the following (not in any hooks)
> 
> ```
> const unreadCount = userInfo.courses.find(
>         (c) => c.course.id === cid,
>       )?.unreadCount
> ```
> 
> If you want, you can use a useMemo to cache the calculation, but seeing as how this is pretty much just a small for loop, doing a useMemo isn't going to improve performance much at all.

useEffect will update when userInfo changes (never)
Declaring it as a const will cause it to happen on every frame (no caching)
I'd agree with useMemo to cache the value directly (removing the need for setState)

Changing from useEffect to a const will hinder performance (barely) and to useMemo not change performance much at all, the point is mostly that the const has the variable recalculated at each render

Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:30]: > Declaring it as a const will cause it to happen on every frame (no caching)

Just curious, is it every frame or is it every re-render?

packages/server/src/profile/user-course.entity.ts
patrickma6199: > > Which one was the second nullable you wanted me to remove? I don't think we should mess with the primary key acting attributed if they're nullable already since previous devs might have worked around it and I don't want to risk breaking anything. If you meant the expires field, I can remove it from there too.
> 
> There's no primary columns that are designated nullable, probably because that's impossible (relational integrity rules) You can remove 'default' from expires because it has a defined default, but we can check the production database to see if there's any null values that currently exist (though I believe that may also be impossible)
> 
> edit: Yeah I'd remove the nullable argument on unreadAsyncQuestions since a column with a default field can't be nullable. Since it'll be an alter table statement that adds it to user course model it can fill the existing rows with the default value

I mean the "primary key acting attributes" (with a typo my bad). I am aware that primary keys cannot be nullable but since we have a lot of primary-key-like columns that are not actually primary keys in the database, the backend relies on them as if they were primary keys for relationships so removing them would not be smart.

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:30]: > > > Declaring it as a const will cause it to happen on every frame (no caching)
> > 
> > 
> > Just curious, is it every frame or is it every re-render?
> 
> My bad for inconsistent phrasing - there's no 'frame' because it's just a DOM render that will re-render based on flags we set in our code. It's on every re-render, and there's far better places to optimize rerenders - since this useEffect would only execute once it actually lowers the complexity of each re-render since the value is cached in the React State object and doesn't have to be recalculated. Declaring it as a const will make it more complex by adding that additional computation on each render. Either way, it is a negligible effect.
> 
> Basically, the DOM will only ever re-render when a flag in a useCallback, useMemo, useEffect, etc. is altered, when component parameters are altered, or when a State object is altered, as far as I remember. What we should worry about more in optimization of renders is ensuring we aren't doing unnecessary amounts of large data retrieval from API routes, less so simple use[...] structures in pages to make them work.

I'm pretty sure that since it comes from the user context and isn't a calculated value, there's nothing to do when re-renders take place here so a const should be fine.

Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:30]: > I'm pretty sure that since it comes from the user context and isn't a calculated value, there's nothing to do when re-renders take place here so a const should be fine.

find() is O(n)

bhunt02 (approved):
Looks good just please fix the tests :)

AdamFipke (approved):
Looks good!

AdamFipke (approved):
Looks good!

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -117,6 +118,14 @@ export default function AsyncCentrePage({
...
   )
 
+  // This endpoint will be called to update unread count back to 0 when this page is entered
+  useEffect(() => {
+    if (
+      userInfo.courses.find((e) => e.course.id === courseId)?.unreadCount !== 0
+    )
+      API.course.updateUnreadAsyncCount(courseId)
+  }, [])
+
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:128]: wait no crap, this doesn't update the userinfo context so it won't actually update until the user refreshes the page

i mentioned this in one of my comments but forgot to double check to make sure the change was made

Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:128]: pushed a fix on prod, only took a min to do

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx]: Just saw this and wanted to add, the find() call is indeed in O(n) time but the dependency array will only allow this effect to run when userInfo is changed (never so long as you're the same user you were). In other words, const is fine. Using, for instance, a useMemo() or useCallback() here would just be adding another hook that runs when dependencies change to a hook that runs when dependencies change.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:128]: Sorry, I just saw this thread. Thanks for this addition on behalf of past Patrick!

PR #198, opened 2025-01-08
TA Notes + Add why-did-you-render (for dev)

### LINKED ISSUE #108: TA Notes
**Description**

This was a request from Ramon.

image
His thought is to have a way for students to click on a staff profile picture which would then open up a page that shows maybe a cut-down version of the profile page.

image

He would also like the ability for users to have a description field, which can really have anything written in it. 

One tricky part is going to be managing access to these semi-public profile pages, since we should probably only allow users that are in the course to look at them.

In the future, this public TA profile page could be expanded to have some extra details like how many students they have helped, their average help time (maybe).

As a future expandability thing (if we do decide to have a semi-public profile page), we could have a way for TAs to open students profiles (maybe if the TA wanted to know how many questions this particular student is asking or something)

Also, instead of a whole page for viewing someone's profile, another option is to just open a modal (which I think would be better since it doesn't take the user off the page)

AdamFipke:
# Description

Adds the ability for TAs to write notes on themselves (or for professors to do so). 

These notes are displayed whenever a user hovers over the staff card on the queue page.

These notes can be modified on the queue page or by going to the course roster page.

Note that if there are no notes set (and that you are not able to edit the notes), the little popover will not show up, which should mean that the student's experience will be unaffected unless the prof or TA opts into the feature.

This is to cover Ramon's use case where he wants a way to display what concepts TAs can help the students with (e.g. this TA can only help MATH101 or something).

This PR also:
- Does some optimizations to the course-role and queue-role guard
- Fixes and optimizes the getUserInfo service function.
	- Previously, it would perform multiple queries (one for selecting the user info, one for total) and more importantly would not actually perform the SELECT part of the query (so it would get everything and filter on JS side). And most importantly I am able to add on the TANotes easily (I couldn't just add it to the SELECT since that didn't work, and I would instead have to map through all of the userCourses for each user to find the matching course and get their TANotes on JS side, which scales much worse than literally just using a select in the query.
- Adds the why-did-you-render package (should help with finding future performance issues). It required some extra steps to get set up compared to other packages, but now you can add `NEXT_PUBLIC_WHY_DID_YOU_RENDER=true` to your frontend .env file and now you will have it enabled (by default it is not enabled since it slows performance a little bit).
- This also makes the mobile UI for the CourseRosterTable look a little nicer

Closes #108 

image
image
image

TA view - they can edit their own notes but not other TAs
image

Student view
image

Mobile course roster table before and after
image

## Type of change

# How Has This Been Tested?

For manual testing, as a professor I have tried editing other TA notes as well as my own on both the queue page and course roster page.
As a TA I have tried editing my own notes and viewed another TA's notes.
As a student I have tried viewing a TA's notes and hovering a staff card without TA notes (no popover shows up as expected),

Also added integration tests for the new updateTANotes endpoint

# Checklist:


packages/server/src/course/course.service.ts
```diff
@@ -227,57 +234,61 @@ export class CourseService {
...
   ): Promise<GetCourseUserInfoResponse> {
-    const query = await getRepository(UserModel)
-      .createQueryBuilder()
-      .leftJoin(
-        UserCourseModel,
-        'UserCourseModel',
-        '"UserModel".id = "UserCourseModel"."userId"',
-      )
-      .where('"UserCourseModel"."courseId" = :courseId', { courseId });
+    // these are the params the query will use
```
AdamFipke, line 237: if you're wondering what the changes here are about, it's because the SELECT (in which I wanted to select the TANotes), LIMIT (take), and OFFSET (skip) were all not working

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
```diff
@@ -550,203 +557,197 @@ export default function QueuePage({ params }: QueuePageProps): ReactElement {
...
+                  studentQuestion?.location === 'Online') ||
+                  queue.type === 'online') &&
+                  clickedZoomModal &&
+                  studentQuestion?.status === OpenQuestionStatus.Helping && (
+                    <JoinZoomButton
+                      zoomLink={queue.zoomLink ?? course?.zoomLink}
+                      textSize="sm"
+                    >
+                      Join Zoom
+                    </JoinZoomButton>
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:673]: this might be pedantic but if this is a specialized component for joining zoom why don't we just hard code 'join zoom' into it

bhunt02 (approved):
Nothing looks wrong here! I think this is good to merge

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/page.tsx:673]: I am confused on what you mean? Is it not already hard-coded?

Also funny thing, I had trouble finding this comment in the files view since I had Hide Whitespace Changes on, it wasn't until I turned it off that I could see it
image

Speaking of, you might want to also turn on Hide Whitespace Changes, it makes reviewing PRs a lot less of a headache

PR #206, opened 2025-01-10
Integrate LMS Data into Chatbot

### LINKED ISSUE #202: (LMS Integration) Integrate LMS Data into Chatbot Knowledge Base
**Description**

Use the existing LMS integration adapter to develop a way to give the course chatbot information directly from Canvas regarding the course announcements and the due dates of assignments.

**Success Criterias**

- Retrieve details for Canvas course announcements and persist them in HelpMe; track when they were both retrieved and when they were last modified (on Canvas!) so that the HelpMe API can know when it is necessary to update these records
- Retrieve details for Canvas course assignments and persist them in HelpMe; follow largely the same steps above but track the due date.
- Develop a conversion method to create documents for the Chatbot that will stress the major details; especially due dates for assignments. 

**Context**

-   N/A

bhunt02:
# Description

Formally integrated the LMSAssignmentModel table and introduced the LMSAnnouncementTable.
* Added/updated associated frontend user-interface elements.
* Added ability to upload the LMS data items to the Chatbot (and persist the related metadata in HelpMe for tracking)
* Ability to delete the LMS document chunks from the Chatbot with the persisted vector store identifiers
* Various visual things

Probably needs more fine-tuning with how the data is formatted, but it seems to work well enough for now! One thing I want to think about is the 'preventative measures' update thing I have coded in which prevents spammy updates from using a ton of $ on our API key, as it refuses to update/recycles the previously saved prompt/information if there's no indication from the LMS that the original document has been modified.

I also reorganized the functions from the Course module and the Organization module that had involvement with LMS functionality that violated SRP (oops that's my bad) and it's a LOT cleaner now. Also moved the tests over and changed how some things are phrased/the routes.

Closes #202 

## Type of change

  - Need to run the following because there's some changes to the foreign keys as they are set up and TypeORM does NOT like updating enum foreign keys that are shared with multiple tables!
```sql
DROP TABLE "lms_assignment_model";
DROP TABLE "lms_announcement_model";
DROP TABLE "lms_course_integration_model";
DROP TABLE "lms_organization_integration_model";
```

# How Has This Been Tested?

I manually tested this by connecting the COSC 304 course in the production data dump to Canvas LMS with an API key provided by Dr. Lawrence and used assignment information and announcement information to save document chunks to the Chatbot vector store. I also tested the prompts manually and they seem to work well enough.

I updated integration and specification (unit) tests as needed, I may be able to implement some minor tests but given how much the LMS integration side of the backend API relies on external API calls, it doesn't seem like a feasible idea to test.

# Checklist:


packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx
```diff
@@ -0,0 +1,326 @@
...
+    return (
+      <div className={'h-full w-full'}>
+        <Spin tip="Loading..." size="large" />
+      </div>
+    )
+  } else {
+    return (
+      <div>
+        <div className="bg-white">
+          <div
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx:298]: I guess this is an unnecessary `<div>`

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -191,13 +133,101 @@ export default function CourseLMSIntegrationPage({
...
+      if (!result) {
+        message.error(
+          `Unknown error occurred, could not enable synchronization with the LMS.`,
+        )
+      } else {
+        message.success(result)
+      }
+      setSyncing(false)
+      setUpdateFlag(!updateFlag)
+    })
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:158]: You might be able to disregard this comment since I haven't checked out the endpoint yet, but would it be possible to give more descriptive error messages if it fails?

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -191,13 +133,101 @@ export default function CourseLMSIntegrationPage({
...
+
+  if (isLoading) {
+    return (
+      <div className={'flex h-full w-full items-center justify-center'}>
+        <Spin className={'text-nowrap'} size="large">
+          <div className={'text-helpmeblue mt-16'}>Loading...</div>
+        </Spin>
+      </div>
+    )
+  }
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:228]: Small thing, but I just thought I would mention that I made a `<CenteredSpinner>` component that may interest you

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:158]: There's generally no error message to return unless it's failing to establish the CRON job (which is just a full-stop failure), and since there are so many different things to save when synchronizing the documents, otherwise there are no failures unless a server logic or database error occurs. I'll remember to add a 'catch' for any thrown backend exceptions before this is merged. 

More informative error messages and I'd be giving stack traces and indications of how our backend operates, which is a no-no with how to treat users. 

But next time do you think you could check against the endpoint just in case it _does_ end up in unnecessary comments?

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -356,33 +400,120 @@ export default function CourseLMSIntegrationPage({
...
+                    <Descriptions.Item label={'API Course ID'}>
+                      {integration.apiCourseId}
+                    </Descriptions.Item>
+                    <Descriptions.Item label={'Course Name (Course Code)'}>
+                      {course.name} ({course.code})
+                    </Descriptions.Item>
+                    <Descriptions.Item label={'Student Count'}>
+                      {course.studentCount}
+                    </Descriptions.Item>
+                  </Descriptions>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:421]: Oh I guess I missed this in the original PR, but I guess doing Descriptions.Item like this is depricated in antd and they want you to pass a `items` prop instead. They have an example in the docs https://ant.design/components/descriptions

Only really worth refactoring since I imagine antd might drop support for it in the next major version

packages/server/src/lmsIntegration/lmsIntegration.controller.ts
```diff
@@ -1,43 +1,315 @@
...
+  @UseGuards(
+    JwtAuthGuard,
+    OrganizationRolesGuard,
+    OrganizationGuard,
+    EmailVerifiedGuard,
+  )
+  @Roles(OrganizationRole.ADMIN)
+  async upsertOrganizationLMSIntegration(
+    @Param('oid', ParseIntPipe) oid: number,
+    @Body() props: any,
```
AdamFipke, line 49: is there a type that can be used instead of `any`? You can then put it as one of the common types and put it inside the API method
E.g. 

Inside `index.ts` in /common
```
 export interface/type/class/whatever UpsertLMSParams = {
        apiPlatform: LMSIntegrationPlatform
        apiKey: string
        apiKeyExpiry?: Date
        apiKeyExpiryDeleted?: boolean
        apiCourseId: string
      }
```
and then you can use that instead of `props: any` here.

and in  `index.ts` in /api
```
upsertCourseIntegration: async (
      courseId: number,
      props: UpsertLMSParams ,
    ): Promise<string> =>
      this.req(
        'POST',
        `/api/v1/lms/course/${courseId}/upsert`,
        undefined,
        props,
      ),
```

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:421]: I was thinking about not using this component anyway, so this gives me a reason to not to.

packages/server/src/lmsIntegration/lmsIntegration.service.ts
```diff
@@ -33,25 +150,593 @@ export class LMSIntegrationService {
...
+      this.syncDocuments(courseId).catch((err) => {
+        console.error(
+          `Failed to synchronize Canvas LMS for HelpMe course ${courseId}`,
+          err,
+        );
+        Sentry.captureException(err);
+      });
+    });
+    try {
+      this.schedulerRegistry.addCronJob(jobName, job);
```
AdamFipke, line 679: oh yeah so fun thing.
When the server restarts, all cron jobs created this way get deleted...

So, you will need a way to auto-restart all of these cron jobs when the server starts up. I have an example of how I did so in `calendar.service.ts`.

Basically, you first need a `onModuleInit()` function (which nestjs runs when the server boots up) and that needs to re-create all of the cron jobs. You can see how I did so with the `retroactivelyCreateAutoCheckoutJobs` function.

An alternative option is to leverage the fact that all of these cron jobs all run at the same time and instead use the `@Cron` decorator kind of like in `backup.service.ts`. You can then have this function just call syncDocuments for every course with an LMSCourseIntegrationModel or every course with a flag set in the DB that this course is syncing documents or however else you would like to do it (I haven't fully reviewed the code so you may already have this. Also you would need to implement this anyway if you wanted to make your own "retroactivelyCreateLMSSyncCronJobs" function). The point is is I think for this usecase using `@Cron` is going to be a much better and easier option (plus you won't have to mock redis in your tests!). I guess schedularRegistry is only good for dynamically creating cron jobs that work at different times.

AdamFipke (changes requested):
I didn't do a super thorough look through but nothin looks concerning. Only non-small thing worth changing is using `@Cron` instead of shedularRegistry

packages/server/src/lmsIntegration/lmsIntegration.service.ts
bhunt02: Yeahhhh I should've looked deeper because for the purposes of this using the @Cron decorator is going to be a hell of a lot easier

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx]: So after reviewing more code I guess there could be a role error or invalid configuration error or any error thrown by one of the called services. While none of these errors are likely to happen and pretty much shouldn't happen, when it *does* happen it can be really useful to know as a user, especially so that it can be fixed if the user decides to report it (which actually happened to me recently and I'm glad I was more verbose with it).

 > More informative error messages and I'd be giving stack traces and indications of how our backend operates, which is a no-no with how to treat users.

I don't mean stack traces (if the endpoints were throwing stack traces that would be ***really*** bad, and I've never seen any of our endpoints do that). I just mean like to print the error message that the backend gave, just with like a `getErrorMessage` function call, like `message.error("Error starting syncronization: " +getErrorMessage(e))`. And I guess the same would go for the other api calls on this page, like `forceSync`, `clearDocuments`, `deleteIntegration`, etc.

Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx]: Oh right as I posted this you already posted the fix lmaoo

AdamFipke (approved):
sweet, good changes!

I guess the only thing that would be left is to manually test it (especially to make sure the cron job is working).

Actually I guess maybe some simple tests for `resynchronizeCourseIntegrations` could be good but other than that it should be good

bhunt02:
image
Here's a visual in case reviewers haven't checked it out.

bhunt02:
Update on visuals:
image

AdamFipke:
Nice, new changes look pretty slick!

bhunt02:
https://github.com/user-attachments/assets/13c481c5-b64c-4ba0-84e8-738f052a7c6e

individual doc stuff

wskksw (approved):

PR #207, opened 2025-01-10
User Guide for chatbot

AdamFipke:
This is 1 md file


bhunt02 (approved):
Markdown file will NOT destroy helpme I think

PR #208, opened 2025-01-11
(Bug Fix) Case-Sensitive Email in Login Form

### LINKED ISSUE #113: [BUG] Case-sensitive login email
**Describe the bug**
For some reason the login email is case-sensitive, and I'm fairly certain emails are supposed to be case-insensitive.
This was also pointed out by @bhunt02

bhunt02:
# Description

Fixed an issue related to login-by-email being case-sensitive, while email addresses are well-known to be case-insensitive identifiers.

Closes #113

## Type of change

# How Has This Been Tested?

* Added integration test (login module has no specification tests) to demonstrate the functionality working as intended.

# Checklist:

### DIFF
--- a/packages/server/src/login/login.controller.ts
+++ b/packages/server/src/login/login.controller.ts
@@ -53,8 +53,16 @@ export class LoginController {
       }
     }
 
+    // An unfortunate double-query so that we can check the email in a case-insensitive way
+    const userId = await UserModel.createQueryBuilder('UserModel')
+      .select('id')
+      .where('LOWER("UserModel"."email") = :email', {
+        email: body.email.toLowerCase(),
+      })
+      .getRawOne<{ id: number }>();
+
     const user = await UserModel.findOne({
-      where: { email: body.email },
+      where: { id: userId?.id ?? -1 },
       relations: ['organizationUser', 'organizationUser.organization'],
     });
 
--- a/packages/server/src/auth/auth.service.ts
+++ b/packages/server/src/auth/auth.service.ts
@@ -16,6 +16,7 @@ import { MailServiceModel } from 'mail/mail-services.entity';
 import { ChatTokenModel } from 'chatbot/chat-token.entity';
 import { v4 } from 'uuid';
 import { UserSubscriptionModel } from 'mail/user-subscriptions.entity';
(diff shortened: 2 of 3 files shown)


packages/server/src/login/login.controller.ts
```diff
@@ -53,8 +53,16 @@ export class LoginController {
...
+    const userId = await UserModel.createQueryBuilder('UserModel')
+      .select('id')
+      .where('LOWER("UserModel"."email") = :email', {
+        email: body.email.toLowerCase(),
+      })
+      .getRawOne<{ id: number }>();
+
     const user = await UserModel.findOne({
-      where: { email: body.email },
+      where: { id: userId?.id ?? -1 },
```
AdamFipke, line 65: wait i might be missing something obvious but why weren't we able to do this in one query? I mean it's not the end of the world obviously since an extra query isn't going to make a noticeable difference to wait to login but i was just curious

bhunt02: unfortunately can't use query functions like LOWER in findOne (as far as i can tell), and it's not easy to get the necessary relations from querybuilder, i think

packages/server/src/login/login-course.service.ts
```diff
@@ -19,6 +19,9 @@ import { khourySemesterCodes } from './last-registration-model.entity';
 export class LoginCourseService {
   constructor(private connection: Connection) {}
 
+  /* TODO: Adam, this is deprecated from the Northeastern code base, correct?
+      We should remove it, yes or no?
+   */
```
AdamFipke, line 24: yes i think this can be deleted, i don't think it's used anywhere

why this is still here i'm not sure other than just no one has spotted to get rid of it. I have deleted a few of these old files as I came across them. There's a handful of these unused bits still scattered around but otherwise good find

packages/server/test/auth.integration.ts
```diff
@@ -20,6 +20,7 @@ import {
 } from 'profile/user-token.entity';
 import { JwtAuthGuard } from 'guards/jwt-auth.guard';
 import { ExecutionContext, Injectable } from '@nestjs/common';
+import { SelectQueryBuilder } from 'typeorm';
```
AdamFipke, line 23: i don't think this import is used?

PR #226, opened 2025-01-28
Caching User Info Data in Redis Cache

### LINKED ISSUE #92: Cache profile in redis
**Description**

The `get` in `profile.controller.ts` is probably the most-frequently called endpoint in the whole system, and the data with it is usually unchanged, making it a great target for redis!

patrickma6199:
# Description
This addresses the time it takes to check user credentials through the context and middleware which occurs at every request by caching the profile object in Redis when it is fetched. User data is rarely altered but in the case that it is, I have added calls to clear the redis entry whenever a user object changes (course changes, registration changes, account changes, etc). I also took the liberty of moving the profile functionality to another service for uniformity and maintainability.

Dependencies: redis-memory-server (I used the same implementation to mock Redis here as with Queue Chats)

Closes #92 

## Type of change

# How Has This Been Tested?

- [x] Updated integration test for profile routes
- [x] Created unit test for redis-profile service
- [x] Updated unit test for profile service

# Checklist:


packages/server/src/redisProfile/redis-profile.service.ts
```diff
@@ -0,0 +1,66 @@
...
+import { GetProfileResponse } from '@koh/common';
+
+@Injectable()
+export class RedisProfileService {
+  /**
+   * The redis client to use for the redis profiles
+   */
+  private readonly redis: Redis;
+
+  private readonly EXPIRATION_TIME = 30 * 60; // 30 min in seconds
```
AdamFipke, line 14: Starting it with 30mins is not a bad idea. In the off chance we missed an endpoint and user's data didn't get updated, they will only have to deal with that outdated info for 30mins.

Later on, when we get more confident that the cache is always getting busted correctly, we can increase the expiration time, possibly upwards to a day.

AdamFipke (approved):
Looks good, nice work

packages/server/src/course/course.controller.ts
bhunt02, line 1: At line 992, function updateUserRole(), I'd make sure to check and delete any cached profiles for this user.

bhunt02: I'd also add one for after 'addStudentToCourse()' in the course.service.ts file (re-used multiple times)

bhunt02: ~~And in the editCourse() function, because the courses are cached in the profile and the details might change~~
This is unnecessary as the useCourse() hook is used for course info

bhunt02: Same thing in removeUserFromCourse()

packages/server/src/organization/organization.controller.ts
bhunt02, line 1: I'd probably add a profile cache clear for 'createCourse()' because the professor themselves can create the course. Maybe even 'updateCourseAccess()'

bhunt02: And probably deleteUserCourses() as well

packages/server/src/course/course.controller.ts
bhunt02: There's also, (in the controller): 'withdrawCourse()'

bhunt02 (changes requested):
Things are mostly good, you asked me to look over to see any vulnerabilities where the cache should be cleared. Any I could think of are attached as comments

packages/server/src/course/course.controller.ts
patrickma6199: Thanks for this!! Just fairly worried that people will be confused that their courses and stuff aren't updating, so any route I miss is a welcomed catch :)

PR #229, opened 2025-02-04
Unread Count => Updates Count (Bugfix)

patrickma6199:
# MERGE SOON PLEASE

# Description

You heard it right, the counter now counts all updated verified anytime questions since the last time your account visited the page! No more double counting!

Keep in mind, when implementing user info caching, and async question comments, merge this first and integrate the updates accordingly.

P.S. also includes a fix for the course entity, and the unreadCount is no longer pulled from the userInfo context since profile caching will be much less meaningful if all records have to change as frequently as this value will for each user.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] Manual: make changes to an anytime question as a staff or student and see other user accounts in that course view the counter increase when pages re-render.
- [X] asyncQuestion.integration.ts rewritten to accommodate changes

# Checklist:


packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -119,20 +119,9 @@ export default function AsyncCentrePage({
...
-        setUserInfo({
-          ...userInfo,
-          courses: userInfo.courses.map((course) =>
-            course.course.id === courseId
-              ? { ...course, unreadCount: 0 }
-              : course,
-          ),
-        })
-      })
+    API.course.updateUnreadAsyncCount(courseId)
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:124]: not necessarily more inefficient, it's only called once per page since the dependencies array is [], just means an API call is made... imo it's negligent in terms of added complexity - still constant time!

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
```diff
@@ -36,8 +38,8 @@ const AsyncCentreCard: React.FC<AsyncCentreCardProps> = ({
           unreadCount > 0 && (
             <>
               <div className="mr-8 h-fit text-sm font-normal text-gray-200">
-                <span className="text-lg font-medium">{unreadCount}</span>{' '}
-                unread
+                <span className="text-lg font-medium">{unreadCount}</span>
+                {` update${unreadCount > 1 ? 's' : ''}`}
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:42]: is this supposed to say "update(s)" or "unread" cause i don't know about the word 'update(s)' for this context... then you wouldn't need the plural ternary either (if it's unread)

packages/server/ormconfig.ts
```diff
@@ -92,6 +93,7 @@ const typeorm = {
     LMSOrganizationIntegrationModel,
     LMSCourseIntegrationModel,
     LMSAssignmentModel,
+    UserCourseAsyncQuestionModel,
```
bhunt02, line 96: not to nitpick but this is triggering some sense of OCD in me that the LMS models are separated

bhunt02 (approved):
looks good overall, I think - i'd wait for a second review just in case since the initial issue wasn't caught before it was merged originally

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -83,6 +84,22 @@ export class asyncQuestionController {
...
+          .set({ readLatest: false })
+          .where('asyncQuestionId = :asyncQuestionId', {
+            asyncQuestionId: question.id,
+          })
+          .andWhere(
+            `userCourseId IN (SELECT id FROM user_course_model WHERE "userId" != :userId)`,
+            { userId: user.id },
+          )
+          .execute();
+      }
```
AdamFipke, line 102: I don't think we should notify all users when a question is voted on, since the voting feels more of a gimmick and not really something that any users would care about

packages/server/migration/1737567086073-addUnreadAsyncQuestionsCount.ts
```diff
@@ -1,19 +0,0 @@
-import { MigrationInterface, QueryRunner } from 'typeorm';
-
-export class addUnreadAsyncQuestionsCount1737567086073
-  implements MigrationInterface
-{
-  name = 'addUnreadAsyncQuestionsCount1737567086073';
-
-  public async up(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query(
-      `ALTER TABLE "user_course_model" ADD "unreadAsyncQuestions" integer NOT NULL DEFAULT '0'`,
```
AdamFipke, line 10: i'm fairly certain you're not actually supposed to delete migration files since it's like deleting history but in this instance it doesn't matter so this is fine

patrickma6199: Since we haven't been using the migrations in prod, we can clean up the history a bit.

packages/server/src/course/course.controller.ts
```diff
@@ -1136,25 +1138,52 @@ export class CourseController {
     return;
   }
 
+  // Moved from userInfo context endpoint as this updates too frequently to make sense caching it with userInfo data
+  @Get(':id/unread_async_count')
+  @UseGuards(JwtAuthGuard)
```
AdamFipke, line 1143: I was also going to make a comment about the lack of a roles guard but i realised that that won't really do anything in this case lol

packages/server/src/course/course.controller.ts
```diff
@@ -1136,25 +1138,52 @@ export class CourseController {
     return;
   }
 
+  // Moved from userInfo context endpoint as this updates too frequently to make sense caching it with userInfo data
```
AdamFipke, line 1141: Okay I was going to write this big long thing about how I disagree but it had me think more about the requirements of this feature and yeah i think i agree.

So we already have email notifications, but the idea behind this feature is that this sorta works to grab users' attention to look here even if they miss the email (as well as being able to soft-notify users of when new questions or comments are posted).

And while making it a separate endpoint will be worse for performance, as we add more features and make the system a thing that people spend more time on, the chances of there being a new unread question while they are actively using the system goes up.

packages/server/src/course/course.controller.ts
```diff
@@ -1136,25 +1138,52 @@ export class CourseController {
...
+      throw new ForbiddenException('User is not in the course');
+    }
+
+    const userCourseAsyncQuestions = await UserCourseAsyncQuestionModel.find({
+      where: {
+        userCourse,
+        readLatest: false,
+      },
+      relations: ['userCourse'],
+    });
```
AdamFipke, line 1162: why are we adding the 'userCourse' relations here?

packages/server/src/course/course.controller.ts
```diff
@@ -1136,25 +1138,52 @@ export class CourseController {
     return;
   }
 
+  // Moved from userInfo context endpoint as this updates too frequently to make sense caching it with userInfo data
+  @Get(':id/unread_async_count')
+  @UseGuards(JwtAuthGuard)
+  async getUnreadAsyncCount(
+    @Param('id', ParseIntPipe) courseId: number,
+    @User() user: UserModel,
```
AdamFipke, line 1146: instead of using `@User`, use `@UserId` since it performs one less query.

Only use `@User` if you need something from UserModel that isn't just userId

packages/server/src/course/course.controller.ts
```diff
@@ -1136,25 +1138,52 @@ export class CourseController {
...
-    if (!userCourse) {
-      throw new NotFoundException('UserCourse not found');
-    }
-
-    userCourse.unreadAsyncQuestions = 0;
-    await userCourse.save();
+    userCourseAsyncQuestions.forEach((uc) => {
+      uc.readLatest = true;
+      uc.save();
+    });
```
AdamFipke, line 1186: A lot of this is abstracted away so idk how bad this is in terms of performance (like if it ends up doing 10 seperate db queries one after the other that would be pretty slow). Idk if there's a way to prepare a bunch of updates in bulk or maybe even just write a manual `UPDATE user_course_async_question SET "readLatest" = true WHERE "userCourseId" = $1`

packages/server/src/course/course.controller.ts
```diff
@@ -1136,25 +1138,52 @@ export class CourseController {
...
+    const userCourseAsyncQuestions = await UserCourseAsyncQuestionModel.find({
       where: {
-        user,
-        courseId,
+        userCourse: {
+          courseId,
+          user,
+        },
       },
+      relations: ['userCourse'],
```
AdamFipke, line 1180: again, not sure why userCourse relations is added here

packages/server/src/profile/user-course-asyncQuestion.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  )
+  @JoinColumn({ name: 'asyncQuestionId' })
+  asyncQuestion: AsyncQuestionModel;
+
+  @Column({ nullable: true })
+  asyncQuestionId: number;
+
+  @Column({ type: 'boolean', default: true })
+  readLatest: boolean;
+}
```
AdamFipke, line 39: By giving this a PrimaryGeneratedColumn, having userCourseId and asyncQuestionId nullable, and without having any ON DELETE CASCADE, if any userCourse is deleted or any async question is deleted, these entities will still persist, which seems unintentional.

I propose the following changes:
- Don't make asyncQuestionId or userCourseId nullable
- Make it so both asyncQuestion and userCourse have onDelete: 'CASCADE'

Having it be a PrimaryGeneratedColumn instead of two primary columns actually does make sense though since that way you can have multiple of these entities, which will be useful since we can create a new one of these entities every time a question is posted on.

Also, I'm really debating whether or not to change it from userCourseId to userId. 
Advantages of doing so:
- At minimum one less query for your get async_unread_count since you can immediately query instead of needing to get usercourse first (and this endpoint gets called often so it's not too negligible)
- UserCourse shouldn't actually have a PrimaryGeneratedColumn since that means users can "be" in the same course multiple times, and by making another table dependent on this id (which is currently unused absolutely everywhere), we will be committing to this and it will be even more annoying to change it in the future

Disadvantages of doing so:
- Having userCourse be there with an onDelete CASCADE means that when a user leaves a course, all of their unread counts also automatically get deleted, which is neat

packages/server/src/profile/user-course-asyncQuestion.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  Column,
+  Entity,
+  JoinColumn,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+import { UserCourseModel } from './user-course.entity';
+import { AsyncQuestionModel } from '../asyncQuestion/asyncQuestion.entity';
+
+@Entity('user_course_async_model')
```
AdamFipke, line 12: one thing i just wanted to note that i like about making this its own model is that in the future we could add the type of notification (e.g. newComment or newQuestionPosted) and then add some kind of visual to indicate that, so that's cool

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -379,14 +433,32 @@ export class asyncQuestionController {
...
+        .where('asyncQuestionId = :asyncQuestionId', {
+          asyncQuestionId: questionId,
+        })
+        .andWhere(
+          `userCourseId IN (SELECT "id" FROM user_course_model WHERE "userId" != :userId)`,
+          { userId: user.id },
+        )
+        .execute();
+    }
+
```
AdamFipke, line 445: Instead of notifying everyone (except the TA) when the question is updated, I see two cases where we should notify:
- When the question creator gets their question human verified (when body.status === asyncQuestionStatus.HumanAnswered or when body.verififed === true) *only the question creator* should get notified
- When a question is marked visible (body.visible === true) *all* **students** (except question creator since it won't be new to them) should get notified

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -271,15 +301,35 @@ export class asyncQuestionController {
...
+        .where('asyncQuestionId = :asyncQuestionId', {
+          asyncQuestionId: questionId,
+        })
+        .andWhere(
+          `userCourseId IN (SELECT id FROM user_course_model WHERE "userId" != :userId)`,
+          { userId: user.id },
+        )
+        .execute();
+    }
+
```
AdamFipke, line 316: Instead of notifying everyone except the creator, I would say notify *all* **staff** in the course only when body.status === AIAnsweredNeedsAttention and *only if they haven't already been notified about that question*
- The idea being that if a student creates a new question, all staff will get +1 unread
	- From there, if a student immediately updates with satisfied or AIAnsweredNeedsAttention (which they *usually* do), all staff don't get an *additional* +1 unread
	- If a student for some reason does not update the status right away and some TA comes and reads it, only they will get +1 unread again

packages/server/src/profile/user-course-asyncQuestion.entity.ts
AdamFipke: > Having it be a PrimaryGeneratedColumn instead of two primary columns actually does make sense though since that way you can have multiple of these entities, which will be useful since we can create a new one of these entities every time a question is posted on.

Actually I'm having second thoughts about this.

So there are two strategies for doing this unread count. There is either automatically creating new user-course-asnycQuestion entities whenever a new question is posted and then **updating** readLatest (which you have done), and then there is creating a new user-course-asyncQuestion whenever we want the user to get notified and then **deleting** all of these entities when they read them.

Doing the update method will be annoying if we want to allow multiple unread entities per question. For example, if a question has 10 comments and therefore has 10 of these entities and then someone posts an 11th comment, you would *need* to create another entity since someone may have never read the other 10 (like you can't just update the existing entities with readLatest = false otherwise the person who hasn't read anything will still only have 10 notifications).

So, given your current implementation, you will want to delete PrimaryGeneratedColumn since there is not a case where you would want multiple of these entities per userCourse-asyncQuestion pair (instead use two Primary Columns).

ALTERNATIVELY, you re-write some code:
- Delete the readLatest column
- Rename this entity to unreadAsyncQuestionUpdate
- Any time you would update readLatest to false, instead insert a new unreadAsyncQuestionUpdate
- Any time you would update readLatest to true, delete all unreadAsyncQuestionUpdate

So now, basically just the count of these entities becomes the number of notifications for a user. New notification -> insert new entity. Read notifications -> delete all entities

AdamFipke (changes requested):
See comments

Summary for important parts:
- I'm guessing you copied your userCourse-asnycQuestion entity off of our flawed user-course entity (not that you may have known that there were issues with it). When I saw your entity, I saw the PrimaryGeneratedColumn (which by having allows there to be multiple of the same userCourse-asnycQuestion relationships) and I got confused and thought that this was done so that you could have multiple notifications per question which is not a bad idea. But, upon looking at the implementation more closely, it seems like it was still built for only one notification per question (by creating all of the userCourse-asyncQuestion entities when the question is created and then *updating* them as need be). But this did give me the idea that it probably should be changed so that each new notification inserts a new userCourse-asnycQuestion and then each "read" action deletes all userCourse-asyncQuestions for that user.
- Some of the logic for when notifications should be sent should probably be changed

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:124]: i think he's just saying that it will be more inefficient if the user goes on this page often since it will now always make the API call rather than only when it needs to. 

Though actually this just had me realise something, Since next.js does some pretty heavy caching, I think that if you click on this page, click to another page, receive a notification, and click back it won't actually make another API call
@patrickma6199 this may need some testing.
Try building this branch with `yarn build:dev`, run `yarn dev:proxy` in another terminal, then run `yarn prod:start` when it's done building. Try this example and see if the unread count is actually updating. If it isn't, then making it a seperate endpoint may have been a fruitless endeavour and it may be better to move it back to the userInfo context since it's not going to update anyways. Unless there's maybe some hack you can do to get the API to call anyways even if the page is cached but I don't know of one that isn't just re-rendering the whole page.

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:42]: I changed the word to update(s) here since it now tracks when anything changes with a visible question (and you were not the one that made the change). I didn't change the variable name because there is a function called updateUnreadCount which I had trouble renaming. I'll think on it, but this is, after all, the hardest part of programming.

packages/server/src/course/course.controller.ts
patrickma6199: I was having issues with it being recognized as a join unless I also populated the relation. If you think it isn't necessary, I can try without once more.

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
patrickma6199: They won't get an additional 1 regardless because, as I said, it no longer double counts. That is the purpose of the new table. The counter will never be higher than the number of questions present (assuming all were still active and being "updated").

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
patrickma6199: Cases relating to the specifics of implementation I would much rather address in another PR when this one is already merged and the counter on the course page currently deployed in production actually means something. Until then, I think my implementation will suffice.

packages/server/src/course/course.controller.ts
AdamFipke: hmm, if it's requiring you to add that relation join, you could try doing userCourse.id inside the where clause instead of just userCourse (just a guess)

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:42]: oh I think the variable name is fine, I'm just thinking that the text would be more user-friendly if it said "unread" instead of "updates" since it's more universally understood what that means in this context, even if it's less "accurate"

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
AdamFipke: Wait I'm still kinda confused the previous code was double counting? It only makes the update once, when the question is created, does it not?

packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx
Comment (patrickma6199) [reply packages/frontend/app/(dashboard)/course/[cid]/components/AsyncCentreCard.tsx:42]: I don't see an instance where people would be confused when seeing something like "1 update" in what that means for their experience when they click into the async centre. This is too pedantic for such a bug-fix PR. We could debate many of the frontend word uses but so long as its fairly clear what is meant, its a waste of time going back and forth to resolve a literal single word change.

packages/server/src/course/course.controller.ts
patrickma6199: My bad, I was referring to the code in the patch route (which was the old implementation here). Since I was using a custom object as the userCourse in the where clause, it was having some error. Regardless, I removed the relation for both queries. Also, when an entity is passed into the where clause, it is the id that is compared to fetch from the database regardless, so I don't think hardcoding it helps

packages/server/src/profile/user-course-asyncQuestion.entity.ts
patrickma6199: Creating and deleting entities is definitely going to be more inefficient. I think you're misunderstanding the meaning of the number. I'll add a too;tip so that users get a run-down, but the number represents the number of async questions that have been updated since you last visited the page. it will not "tick up" as more events occur on a single question. Again, that is the purpose of having a join table between async questions and user course entities (so that there is some unique relationship of many to many) to store the relationship attribute, "readLatest". 

Notes:
- No, it will not be more efficient to change the backend implementation from the data storage level from a development perspective especially to fix a **broken** feature currently in production
- No, it will not be more efficient from an implementation perspective because constantly insert and delete database entires will require the database server to perform more low-level operations (checking for free memory space) when writing and deleting than simply updating all rows with a given foreign key (a very common database query that I am sure the DBMS is optimized to handle)
- There is no benefit to having a single generated primary column: if you think about any given situation, there will never be a case where we would need double entries (where userCourseId and asyncQuestionId clash), hence why in normalization, they would be sufficient fk's. I thought we discussed this before but we (perhaps you given your experience with the system) is going to have to remap all the backend routes and fks within entities as having pointless nullable fks that are not real fks is common practice in this repository. It kinda defeats the purpose of your change if I go through all of that to have true formal fks to UserCourse and AsyncQuestion just for UserCourse's key (any every other table for that matter) to still be improper. To summarize this point, there is little purpose to making big entity changes right now in the pursuit of data safety when our core tables are set up wrong, making all fks to them wrong, making backend relations wrong in they way they are joined, If the goal is ultimately to fix all the tables, we should address this soon in a **separate PR**. Please don't bring up this point about primary keys again regarding new tables, as I hope its clear, I see no justification in being strict about fixing it for new entities (again, we will have to go back into every entity to make major changes if we mean to actually fix this issue).

packages/server/src/profile/user-course-asyncQuestion.entity.ts
patrickma6199: Yeah, and I was thinking of animating questions on the async centre as well (similar to how queue cards look when checked in) since we can pinpoint which questions have been updated for each userCourse.

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
patrickma6199: It would double count if it counts an update once and counts the questions creation before a given user has entered the page. So, instead of counting the number of overall updates (eg. 20 new comments on a single question => 20 updates), I am now only tracking the number of updated questions (eg. 20 new comments on a single question => 1 update).

AdamFipke: Er wait I might still be confused but wasn't the query to add +1 only in the createQuestion endpoint before (and not any other endpoint)? I think my brain is just having trouble processing "double count if it counts an update once and counts the questions creation"

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
AdamFipke: I guess I think this current implementation would introduce some new bugs (like if a TA edits a response or marks a question as visible, all other TAs get notified even though they probably don't care (which kinda defeats the purpose of the feature)).

The same goes for in the StudentEditQuestion endpoint, since everyone will be notified when a student edits their question text a bit even if no one will really care (or if they DELETE a question, everyone will be notified).

AdamFipke: Oh yeah and speaking of. Since deleting anytime questions only do a soft delete, maybe it might be worthwhile to manually delete any created userCourse-AsyncQuestion entities for that asyncquestion whenever one is deleted?

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:124]: Okay I tested this and it turns out to be fine (no caching), so that's good

AdamFipke (approved):
Okay Patrick and I have talked and we agreed that this will be okay to merge (I'll put on prod tonight), and that I will be making a seperate PR for some of the other changes that we can continue to discuss

PR #232, opened 2025-02-05
System Chatbot

### LINKED ISSUE #222: A way to share the userguide to all course chatbots
**Description**

Apparently this might be harder than it first seems, and that there isn't really a "global" document store and implementing one may be hard (since the RAG system would then need to look through two different stores of chunks to figure out which one is relevant and according to Kevin that may be hard).

One idea is for some sort of interface or tooling that allows us to mass upload a userguide document into all courses' chatbots (including when they first create a course). We would also need a way to mass delete the userguide documents in case we wanted to update it. Not a pretty solution but it may be something. (note: maybe with a promise.all on frontend to mass upload to all courses and then have a progress bar to show how many are done)

Also, we could set a flag inside the course on whether they have deleted the userguide so that they don't get a new one.

AdamFipke:
# Description

Adds a "System" chatbot.

So there's a lot of different ways to get the userguide to the chatbot, but this is by far the easiest.
Basically, you must first create a "HelpMe" **course** where you upload the chatbot. Then, give the user a toggle that lets them ask "System" questions, which will then send the question to the "HelpMe" course.

Not only is this insanely simple to do, it adds a lot of other advantages:
- Can send what page the user is on if the user is on "System" to get better help from the userguide
- Can be used to collect feedback (and what kinds of system questions users ask)
	- much better than having them email me (which no one has done btw)
- Can easily customize the chatbot prompt or give suggested questions

Collaborated ideas with @wskksw 

image
image
image
image

NOTE that there is a new .env variable that you need to set in order for this to work (`NEXT_PUBLIC_HELPME_COURSE_ID`)

Closes #222 

## Type of change

# How Has This Been Tested?

It's all frontend so a bunch of manual testing

# Checklist:

### DIFF
--- a/packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
@@ -11,6 +11,7 @@ import {
   Space,
   Segmented,
   Alert,
+  Popconfirm,
 } from 'antd'
 import {
   CheckCircleOutlined,
@@ -91,6 +92,10 @@ const Chatbot: React.FC<ChatbotProps> = ({
   const hasAskedQuestion = useRef(false) // to track if the user has asked a question
   const pathname = usePathname()
   const currentPageTitle = convertPathnameToPageName(pathname)
+  const [popResetOpen, setPopResetOpen] = useState(false)
+  // used to temporarily store what question type the user is trying to change to
+  const [tempChatbotQuestionType, setTempChatbotQuestionType] =
+    useState<ChatbotQuestionType | null>(null)
 
   const courseIdToUse =
     chatbotQuestionType === 'System'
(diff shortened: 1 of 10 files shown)


wskksw (approved):

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -411,6 +455,16 @@ const Chatbot: React.FC<ChatbotProps> = ({
                             </div>
                             <div className="flex flex-col gap-1">
                               {item.sourceDocuments &&
+                              chatbotQuestionType === 'System' ? (
+                                <div className="align-items-start flex h-fit w-fit max-w-[280px] flex-wrap justify-start gap-x-2 rounded-xl bg-slate-100 p-1 font-semibold">
+                                  <p className="px-2 py-1">User Guide</p>
+                                  {getSourceLinkButton(
+                                    'User Guide',
+                                    'https://github.com/ubco-db/helpme/blob/main/packages/frontend/public/userguide.md',
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:463]: isnt our repo private?

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -324,6 +350,24 @@ const Chatbot: React.FC<ChatbotProps> = ({
             )}
             extra={
               <>
+                <Segmented<ChatbotQuestionType>
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:353]: i wonder if maybe adding a check that the system chatbot's course (your example has -1 for id) exists somewhere in the code and disabling this option if not might be helpful in preventing unintended behaviour

Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:353]: like maybe just adding a route that checks if that course exists and returns a boolean and we can store it in an app-wide context or something

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -535,6 +590,13 @@ const Chatbot: React.FC<ChatbotProps> = ({
                     className="mt-3"
                   />
                 )}
+                {courseIdToUse === -1 && (
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:593]: i'm not sure this is a good thing to show to all users - exposing what our application structure looks like behind the scenes is a no-no imo

bhunt02 (changes requested):
some fleeting thoughts here, overall this looks great and is a great feature to have for the user base! just had some thoughts about preventing unintended and/or error states

i'll approve as soon as addressed or justification otherwise is made

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:463]: nope, it's public. It's the chatbot repo that's private

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:629]: Uhg I mean the idea behind this is it will never be shown to users, only us developers, but I see what you mean in case it becomes erroneous for some reason. I'll make a change i guess

PR #234, opened 2025-02-07
Unread Async Bug Fixes Bug Fixes

AdamFipke:
# Description

Does the following:
- renames user-course-asyncQuestion into unread-async-question since it's easier to say/find and is more self-explanatory on why its there. 
- It no longer uses PrimaryGeneratedColumn and instead uses PrimaryColumn to prevent multiple notifications for a question to ever be created
- It now references user and course rather than UserCourse's id since it makes some queries easier and eliminates a dependency for a primarygeneratedcolumn that shouldn't exist
- It now deletes notifications when a question is deleted (useful if a student creates a question (notifying TAs) and then deletes the question before any TA can see it)
- It now deletes all notifications if a user withdraws from a course
- Made the getUnreadAsyncCount and updateUnreadAsyncCount both only use 1 query and also properly made them use UserId instead of User decorators.
- Changed code to only give notifications in certain situations rather than ANY update
	- no more notifying everyone that a student/ta deleted a question
	- no more notifying everyone (even TAs) when a student or TA edits a visible question (or even just opens the edit modal and clicks save, since the frontend re-sends all fields not just changed ones and there's no checks to see if anything has actually changed)
		- like if a TA marks a questions visible or edits a response, all other TAs would get notified and I don't think they would care
		- if a student edits their visible question to fix a typo, all students and staff would be notified of said change

## Type of change

# How Has This Been Tested?

I have manually tested to make sure each query/change runs and works as expected (this includes withdrawing from a course to see if the notifications get deleted, they do).

I have also tried modifying a visible question and it does not re-trigger unread updates.

# Checklist:


patrickma6199 (approved):
Everything looks good to me. I like the name of the table just because I hated typing the other one, and making the table's primary key include the course and user ids instead of the user course directly. There's one suggestion I wanted to add though.

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -181,14 +180,15 @@ export class asyncQuestionController {
       });
 
       if (usersInCourse?.length) {
-        await getRepository(UserCourseAsyncQuestionModel)
+        await getRepository(UnreadAsyncQuestionModel)
           .createQueryBuilder()
           .useTransaction(true)
```
patrickma6199, line 185:
```suggestion
```

patrickma6199: I'm not sure what I was thinking here since this is a single query (wouldn't need a transaction). Also, I have found no documentation for this function and probably shouldn't have used it in the first place. should probably get rid of it before merging.

AdamFipke: sure! I have no problems with making this change

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -319,6 +345,8 @@ export class asyncQuestionController {
         `c:${question.course.id}:aq`,
         updatedQuestion,
       );
+      // delete all unread notifications for this question
+      await UnreadAsyncQuestionModel.delete({ asyncQuestionId: questionId });
```
bhunt02, line 349: good catch, but is the FK not cascade delete?

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -426,19 +455,42 @@ export class asyncQuestionController {
...
 
-    // if a question is updated by staff and is verified, mark all user's questions as unread
-    if (updated && updatedQuestion.visible) {
-      await getRepository(UserCourseAsyncQuestionModel)
-        .createQueryBuilder()
-        .useTransaction(true)
-        .update(UserCourseAsyncQuestionModel)
+    // Mark as new unread for all students if the question is marked as visible
+    if (body.visible && !oldQuestion.visible) {
+      await createQueryBuilder(UnreadAsyncQuestionModel)
```
bhunt02, line 460:
```suggestion
      await UnreadAsyncQuestionModel.createQueryBuilder()
```

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -426,19 +455,42 @@ export class asyncQuestionController {
...
+        .execute();
+    }
+    // When the question creator gets their question human verified, notify them
+    if (
+      oldQuestion.status !== asyncQuestionStatus.HumanAnswered &&
+      !oldQuestion.verified &&
+      (body.status === asyncQuestionStatus.HumanAnswered ||
+        body.verified === true)
+    ) {
+      await createQueryBuilder(UnreadAsyncQuestionModel)
```
bhunt02, line 485:
```suggestion
      await UnreadAsyncQuestionModel.createQueryBuilder()
```

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -285,31 +288,54 @@ export class asyncQuestionController {
...
+        .execute();
+    }
+    // if the question is visible and they rewrote their question and got a new answer text, mark it as unread for everyone
+    if (
+      updatedQuestion.visible &&
+      body.aiAnswerText !== oldQuestion.aiAnswerText &&
+      body.questionText !== oldQuestion.questionText
+    ) {
+      await getRepository(UnreadAsyncQuestionModel)
         .createQueryBuilder()
```
bhunt02, line 330:
```suggestion
      await UnreadAsyncQuestionModel.createQueryBuilder()
```

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -285,31 +288,54 @@ export class asyncQuestionController {
...
 
-    // if a question is updated by student and is verified, mark all course users' entries as unread
-    if (updated && updatedQuestion.visible) {
-      await getRepository(UserCourseAsyncQuestionModel)
+    // Mark as new unread for all staff if the question needs attention
+    if (
+      body.status === asyncQuestionStatus.AIAnsweredNeedsAttention &&
+      oldQuestion.status !== asyncQuestionStatus.AIAnsweredNeedsAttention
+    ) {
+      await createQueryBuilder(UnreadAsyncQuestionModel)
```
bhunt02, line 305:
```suggestion
      await UnreadAsyncQuestionModel.createQueryBuilder()
```
current syntax (as well as getRepository()) is deprecated in later versions of typeorm

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -181,14 +180,14 @@ export class asyncQuestionController {
       });
 
       if (usersInCourse?.length) {
-        await getRepository(UserCourseAsyncQuestionModel)
+        await getRepository(UnreadAsyncQuestionModel)
           .createQueryBuilder()
```
bhunt02, line 184:
```suggestion
      await UnreadAsyncQuestionModel.createQueryBuilder()
```
deprecated syntax, getRepository(x) is deprecated in later versions of typeorm

bhunt02 (changes requested):
just some stuff that will make my life easier mostly, looks mostly good so I'll approve once these are addressed (if possible)

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
patrickma6199: Yeah, but I found out during my patch that the questions are archived so these entities have to be manually deleted.

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
patrickma6199: That's good to know.

bhunt02 (approved):
Don't see any more problems here, looks good

PR #236, opened 2025-02-09
Upgrade TypeORM to most recent version (0.3.20)

### LINKED ISSUE #57: Upgrade typeorm to v0.3.x or greater
**Description**

- I'm tired of getting critical vulnerability emails every week
- Also, it is really annoying that you basically can't do joins and select statements at the same time in the current version of typeORM, we could get a pretty nice performance boost by just adding some select statements to these
- A lot of endpoints have been filtering on the server-side instead of the database-side. Replace any lodash `pick` methods or other object assignments with select statements instead.
- This is probably gonna take some work. Both Dima and I have tried upgrading it, saw a bunch of errors, and then reverted lol. Though, seeing as how we'd be going from v0.2.x to v0.3.x, this is a major version change so I imagine a lot of endpoints will need to be changed.

bhunt02:
# Description

Upgrade TypeORM to most recent version; forced to upgrade a bunch of different parts of Nest to accommodate this. What happens as a result? Really bad side effects!
Side effects:
* keepConnectionAlive attribute from the test module for TypeORM was actually deprecated at some point; it's now a default parameter, but NestJS doesn't recognize it as such. The default time for the TCP connection to keep-alive (or something) is 30000 ms, so 30 seconds.
* Original implementation of Redis module is no longer recognized or viable.
* Changes to NestJS syntax which meant that services could not be solely imported, have to be imported by module name.
* RedisMemoryServer seems to have some connection issues
* You can no-longer query with models in 'where' clause sometimes for some reason, and instead must use Ids. Tests caught a lot of these, but there's bound to be areas that broke 
image
image
* And probably more I have yet to find.

This is basically like a refactor upgrade of the system where I have to change a bunch of stuff and hope it works fine. Avoiding upgrading all the way to Nest 11.0.0+ as that will force us to upgrade the Node version up to 20+ (plus break even more packages)

Closes #57 

## Type of change

  - Very likely (mostly from how querying with models in the 'where' clause doesn't seem to work)

# How Has This Been Tested?

Will require manual tests for basically every endpoint as the syntax has changed - but most things should still be supported.

# Checklist:


packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx
```diff
@@ -160,7 +160,7 @@ const QuestionCard: React.FC<QuestionCardProps> = ({
                             isMyQuestion &&
                             studentAssignmentProgress &&
                             studentAssignmentProgress[taskKey] &&
-                            studentAssignmentProgress[taskKey].isDone
+                            studentAssignmentProgress[taskKey]?.isDone
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QuestionCard.tsx:163]: wild

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/TaskDisplay.tsx
```diff
@@ -82,7 +82,7 @@ const TaskDisplay: React.FC<TaskDisplayProps> = ({
       for (const tempTask in taskTree) {
         if (
           taskTree[tempTask].precondition &&
-          taskTree[tempTask].precondition.taskId === taskID
+          taskTree[tempTask].precondition?.taskId === taskID
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/TaskDisplay.tsx:85]: also wild, linter doing weird things i guess even though the checks for precondition not being falsy are already there

packages/server/src/calendar/calendar.controller.ts
```diff
@@ -105,11 +109,18 @@ export class CalendarController {
   @UseGuards(CourseRolesGuard)
   @Roles(Role.TA, Role.PROFESSOR)
   async updateEvent(
-    @Param('calId', ParseIntPipe) calId: string,
+    @Param('calId', ParseIntPipe) calId: number,
```
AdamFipke, line 112: nice catch with this

packages/server/src/insights/insight-objects.ts
```diff
@@ -959,13 +965,12 @@ export const StaffWorkload: InsightObject = {
...
-      UserModel,
-    )
-      .select('UserModel.id', 'id')
-      .addSelect("UserModel.firstName || ' ' || UserModel.lastName", 'name')
-      .where('UserModel.id IN (:...ids)', { ids })
-      .getRawMany<{ id: number; name: string }>();
+    const staffNames: { id: number; name: string }[] =
+      await UserModel.createQueryBuilder()
+        .select('UserModel.id', 'id')
+        .addSelect("UserModel.firstName || ' ' || UserModel.lastName", 'name')
```
AdamFipke, line 971: oh yeah so doing concatenating the firstname and lastname like this will make the full thing become null if lastName is null (which some people do) so you need to COALESCE for lastName and/or firstName

.eslintrc.js
```diff
@@ -10,7 +10,6 @@ module.exports = {
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
     "prettier",
-    "prettier/@typescript-eslint",
```
AdamFipke, line 13: i updated the prettier and eslint versions (i forgot why exactly), but this plugin in particular doesn't exist anymore (i believe was integrated into prettier)

.gitignore
```diff
@@ -50,4 +50,8 @@ husky.sh
 /backups/*
 
 # location of emails sent on dev environment
-/packages/server/src/mail/sent_dev_emails.log
\ No newline at end of file
+/packages/server/src/mail/sent_dev_emails.log
+
+# location of desktop notifications sent on dev environment
+/packages/server/src/notification/desktop_notifications.log
+
```
AdamFipke, line 57: this might be noteworthy if you are testing these things

docs/DEVELOPING.md
```diff
@@ -102,13 +101,12 @@ docker-compose build && docker-compose up
...
 
 If you're creating a new `table_name.module.ts` file, you must also add it to the `app.module.ts` file 
-If you're creating a new `table_name.controller.ts` file, you must also add it to the `ormconfig.ts` file 
 
 Also, you must update the `seed.controller.ts` file to reflect the new database changes. This seed file is used to populate the database with dummy data on http://localhost:3000/dev.
 
 #### Migrations
 
-If you change an entity, you MUST run `yarn migration:generate -n [migration-name]`, to make the migration file, then `yarn migration:run` will automatically run on deployment to staging/production. Commit the migration file to Git!
+If you change an entity, you MUST run `yarn migration:generate ./migration/your-migration-name -d ./typeORMCLI.config.ts`, to make the migration file, then `yarn migration:run` will automatically run on deployment to staging/production. Commit the migration file to Git!
```
AdamFipke, line 109: this was an unfortunate change that needed to be made. 

So typeorm changed it so you no longer pass `-n my-migration-name` but rather you need to specify the path to the file as the **first** parameter.

Thus, the command needs to be in the order of `migration:generate ./migration/your-migration-name -d ./typeORMCLI.config.ts` and not `migration:generate -d ./typeORMCLI.config.ts ./migration/your-migration-name`.

And for some reason yarn does not allow you to pass arguments into the *middle* of a command, only at the end.

For example, I wanted to make it so you could just do `yarn migration:generate migration-name` but there doesn't exist a way I can pass migration-name into the middle of the full command without maybe upgrading yarn or with a bash script

docs/NEWDEVS_STARTHERE.md
```diff
@@ -147,7 +147,7 @@ Want to conditionally render something like an `elseif` statement? Do: `{conditi
 
 `service` - These define methods that make calls to the database. Unit tests test these.
 
-`entity` - These define the database schema. If you make any changes to these, be sure to make a migration (see `DEVELOPING.md`). If you are making a new entity, be sure to add it inside `ormconfig.ts`!
```
AdamFipke, line 150: I have good news! you don't need to do this anymore. I just added a blob pattern to ormconfig.ts that will automatically add any .entity.ts files to our orm config

packages/common/index.ts
```diff
@@ -70,7 +70,6 @@ export class User {
   desktopNotifsEnabled!: boolean
   @Type(() => DesktopNotifPartial)
   desktopNotifs!: DesktopNotifPartial[]
-  insights!: string[]
```
AdamFipke, line 73: was unused, same with queue startTime and endTime

packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx
```diff
@@ -79,7 +79,7 @@ const QueueCard: React.FC<QueueCardProps> = ({
         classNames={{
           header: cn(
             'text-white rounded-t-lg',
-            queue.isOpen ? 'bg-[#25426C]' : 'bg-[#1e3659]',
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/QueueCard.tsx:82]: so .isOpen was some kinda calculated attribute that was mostly pointless since I think it was basically always true unless the queue was disabled.

With the typeorm version upgrade, I could no longer just have attributes on entity objects without also making them a column. Thus, I had to decide whether to remove .isOpen as an attribute or make it a full column. And since this attribute was already kinda pointless this was an easy decision.

A harder decision was the queueSize attribute. It would of been a royal pain to refactor a bunch of the backend to just turn queueSize into a regular function that you can call when you want to get the queueSize and say add it onto the queue return DTO.

Thus, I made queueSize a pointless column that you can modify to be updated when you call myQueueModel.addQueueSize(). Another option would be to make queueSize an actual calculated column that will automatically update when the question table is updated, but that felt like a little bit more work (plus maybe less performant)

packages/server/.gitignore
```diff
@@ -43,3 +43,6 @@ uploads/*
 
 # postgres.env (place in packages/server)
 postgres.env
+
+# Sentry Config File (Also auto-generated...)
+.env.sentry-build-plugin
```
AdamFipke, line 48: oh yeah i tried to fix sentry sourcemaps not being accurate in this in case you're curious on some changes (such as the ones in webpack.config.js)

packages/server/package.json
```diff
@@ -131,7 +133,7 @@
     "ts-loader": "^8.0.10",
     "ts-node": "^8.6.2",
     "tsconfig-paths": "^3.9.0",
-    "typeorm-factory": "^0.0.14",
+    "typeorm-factory": "github:AdamFipke/typeorm-factory",
```
AdamFipke, line 136: hehe

For recap, the old typeorm-factory package is incompatible with typeorm v0.3.x. I found a fork that updated it, but it was not built. So I built it and that's this. 

Also as a result if you want to add a new factory, you need to add it to a couple files (factories.ts and factory.service.ts)

packages/server/src/alerts/alerts.controller.ts
```diff
@@ -32,13 +32,13 @@ export class AlertsController {
...
   ): Promise<GetAlertsResponse> {
     const alerts = await AlertModel.find({
       where: {
         courseId,
-        user,
-        resolved: null,
+        userId,
+        resolved: IsNull(),
       },
     });
```
AdamFipke, line 43: This is an important change to note.

A. You can no longer query with `null`, and must use `IsNull()`
B. For some reason you can no longer query with a Model (in this case it was a UserModel). You must query with an id.


I'm going to add this to the docs

docs/NEWDEVS_STARTHERE.md
AdamFipke: okay nevermind. The tests pass with this but when I try to run `yarn dev` it doesn't work (I get "SyntaxError: Cannot use import statement outside a module" for some reason, probably some commonJS vs esm or something reason)

packages/server/src/decorators/queue-role.decorator.ts
```diff
@@ -9,10 +9,10 @@ import { UserCourseModel } from 'profile/user-course.entity';
 export const QueueRole = createParamDecorator(
   async (data: unknown, ctx: ExecutionContext) => {
     const request = ctx.switchToHttp().getRequest();
-    const queue = await QueueModel.findOne(request.params.queueId);
+    const queue = await QueueModel.findOneBy({ id: request.params.queueId });
```
AdamFipke, line 12: oh yeah findOneBy can be a bit nicer syntax instead of needing to do `.findOne({ where: { id: ... } })` (one less layer of brackets, as it just lets you pass in the where clause directly)

packages/server/test/util/testUtils.ts
```diff
@@ -59,27 +65,28 @@ export function setupIntegrationTest(
...
+    //   console.log('Starting Redis in-memory server');
+    //   try {
+    //     redisTestServer = new RedisMemoryServer();
+    //     redisHost = await redisTestServer.getHost();
+    //     redisPort = await redisTestServer.getPort();
+    //   } catch (err) {
+    //     console.error('Error initializing redis test container:', err);
+    //     throw err;
+    //   }
+    // } else {
```
AdamFipke, line 80: this was another unfortunate change. I really don't know why it was failing (it gave rather unhelpful error messages). Given more time I could probably figure out why but it's not really a priority

PR #239, opened 2025-02-11
OrgOrCourseRolesGuard + Allow Course Profs to Edit Their Courses

### LINKED ISSUE #237: OrgOrCourseGuard
**Description**

A guard that allows anyone with the specified course role OR organisation role to access the endpoint.

Will be useful for #200

AdamFipke:
# Description

Creates a new guard that allows you to specify an endpoint where the user needs either the right course role or the right org role
- it does two queries simultaneously to check if they either have the right org role or the right course role (for the given course)

**This also fixes some access problems where some endpoints would error for course profs**:
- deleteUserCourses (used to remove students from your course)
- getOrganizationCourse (used when the staff goes on any page to edit a course)
- updateCourse (used when prof wants to update their course details)

Another piece of motivation for creating this guard is it will make it easier to extend admin permissions so that they can call more endpoints without being in a course (for the future)

Here is an example of a course professor updating the course page (which would previously be impossible)
image

Closes #237 

## Type of change

# How Has This Been Tested?

Added a bunch of new tests to whatever endpoint is utilizing the new guard.

I have also manually tested:
- viewing course page as org admin
- dropping a user from a course as an org admin
- viewing and updating a course as course prof
- dropping a user from a course as a course prof

I didn't manually test with an org prof (who would also end up needing to be a course prof due to how the system is set up) since the same logic is checked

# Checklist:

### DIFF
--- a/packages/server/src/guards/org-or-course-roles.guard.ts
+++ b/packages/server/src/guards/org-or-course-roles.guard.ts
@@ -75,12 +75,15 @@ export class OrgOrCourseRolesGuard implements CanActivate {
     return true;
   }
 
+  /* Returns false if the course does not exist, if they are not in the course, or if they have the wrong role.
+  This is because ultimately I opted for having better performance with having less queries with the tradeoff of less-granular error messages.
+  Same mantra goes for matchOrgRoles
+  */
   async matchCourseRoles(
     roles: string[],
     userId: number,
     courseId: number,
   ): Promise<boolean> {
-    // first, use query builder to check if they have the right course roles
     try {
       await UserCourseModel.findOneOrFail({
         where: {
--- a/packages/server/src/guards/org-or-course-roles.guard.ts
+++ b/packages/server/src/guards/org-or-course-roles.guard.ts
@@ -123,8 +123,6 @@ export class OrgOrCourseRolesGuard implements CanActivate {
         ou_role: string;
       }>();
 
-    console.log(orgCourseUser);
-
     if (!orgCourseUser) {
       throw new NotFoundException(
         'This course does not exist in your organization',
--- a/packages/server/src/guards/org-or-course-roles.guard.ts
+++ b/packages/server/src/guards/org-or-course-roles.guard.ts
@@ -10,7 +10,9 @@ import {
 } from '@nestjs/common';
 import { Reflector } from '@nestjs/core';
 import { OrganizationCourseModel } from 'organization/organization-course.entity';
-import { UserModel } from 'profile/user.entity';
+import { OrganizationUserModel } from 'organization/organization-user.entity';
+import { UserCourseModel } from 'profile/user-course.entity';
+import { In } from 'typeorm';
 
 /* The user must have one of the specified CourseRoles OR OrgRoles in order to access this endpoint.
     This could be used if there was an endpoint you want the course professor to be able to use or the org admin.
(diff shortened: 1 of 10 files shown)


patrickma6199 (approved):
LGTM! Just to clean up the implementation, I think you should also change the existing role guards' implementation of matchRoles(). I'm not sure why it filters through the array and checks a number instead of just using includes(). This is me being nitpicky though, doesn't really matter or make it more efficient. Also, I really need this feature for some of my routes for semesters. Let's get this closed!

packages/server/src/guards/org-or-course-roles.guard.ts
```diff
@@ -0,0 +1,141 @@
...
+        oc_id: number;
+        oc_organizationId: number;
+        oc_courseId: number;
+        ou_id: number;
+        ou_role: string;
+      }>();
+
+    if (!orgCourseUser) {
+      throw new NotFoundException(
+        'This course does not exist in your organization',
```
bhunt02, line 130: couldn't this be not found because the user's not part of the course? seems like a better message to leave regardless - it's the more likely scenario regardless

packages/server/src/guards/role.guard.ts
```diff
@@ -29,9 +29,14 @@ export abstract class RolesGuard implements CanActivate {
...
 
   async canActivate(context: ExecutionContext): Promise<boolean> {
-    const roles = this.reflector.get<string[]>('roles', context.getHandler());
+    let roles = this.reflector.get<string[]>('roles', context.getHandler());
     if (!roles) {
-      return true;
+      // if it doesn't have the roles decorator, maybe it has the CourseRoles decorator?
+      roles = this.reflector.get<string[]>('CourseRoles', context.getHandler());
+      if (!roles) {
+        // if it lacks any role decorator, then this guard won't do anything (TODO: maybe throw a NotImplementedException instead? Though doing this may break some endpoints)
```
bhunt02, line 37: it makes sense to just return true if no roles are specified, however I'm pretty sure this is just a really worst-case fallback in case the roles decorator doesn't properly assign to 'roles', because it's a mandatory parameter

packages/server/src/organization/organization.controller.ts
```diff
@@ -580,51 +581,54 @@ export class OrganizationController {
...
-      await UserCourseModel.delete({
-        courseId: cid,
-        role: Role.PROFESSOR,
-      });
-
-      for (const profId of courseDetails.profIds) {
-        const chosenProfessor = await UserModel.findOne({
-          where: { id: profId },
+      // update profs (only if admin since only admins can adjust profs in courses)
+      if (user.organizationUser.role === OrganizationRole.ADMIN) {
```
bhunt02, line 585: smart

bhunt02 (approved):
I don't have any complaints, the code here looks good. Left a comment on one of yours that specified a TODO I don't think is necessary.

packages/server/src/guards/role.guard.ts
AdamFipke: yeah I guess the idea with throwing a NotImplementedException is that it would only get thrown if the developer using the guard didn't set it up properly (and for example forgot to add a Roles decorator). And it would be fine to return NotImplemented since the Roles decorator is static at runtime and thus couldn't be caused by the user.

Idea being that newer devs probably aren't going to dig into the guards and other decorators to really see how they work (I for one didn't look for a while) and to warn them if they're using the CourseRoles guard without a Roles decorator (which would in case make the CourseRoles guard return true i.e. do nothing)

It's a small thing, anyway

packages/server/src/guards/org-or-course-roles.guard.ts
AdamFipke: Yeah you're right about this

So I've actually put way too much thought into the error messages and there's like this tradeoff where if I want to give better or more accurate error messages, I would need to perform more database queries.

For example, in matchCourseRoles, since it only does one query that finds the corresponding UserCourseModel with the correct role, it will give the same error message regardless if the course does not exist, if they are not in the course, or if they have the wrong role.

And there's even more of this with the matchOrgRoles

Ultimately I opted for having better performance with having less queries with the tradeoff of less-granular error messages.

But anyways yeah you have the right idea with that

PR #240, opened 2025-02-11
Courses Page Table View and Semester Formalization

### LINKED ISSUE #233: Table View for Courses Page (Home of Organization Page) and Semester Determination
**Description**

Add a way for the frontend interface to render a table view of the courses sorted by name and semester. Also, standardize the semester attribute to some degree to allow for more behaviours that hinge on a course's semester. Perhaps allow for the organization to set standard semesters which can then be used by all courses.

patrickma6199:
# Description

This PR contains a formalized semester configuration that gives organization admin and professors the ability to create, edit, and delete semesters (it used to be that, so long as the format was followed from the legacy implementation, submitting a change to the semester setting of courses would create a new one if it couldn't find one matching its season and year in the database). Now, semesters track start and end dates and allow for semester names and descriptions (optional) to be set as well. The course settings page now uses a dropdown that lists the course organization's semesters.

In the courses page, courses can now be displayed in a table format as well where each semester's courses are grouped and the tables are paginated. The course cards now also display their respective semesters. Mousing over the semester divider in table view or the semester tag in the card view displays the semester details in a popover component.

Closes #233 

## Type of change

```
// Run this after the migration
// This will reset all semester settings leading to all production courses having no attached semesters. In order to update any course information from this point, one must set a semester for the course which will steadily force adoption of the new semester layout
UPDATE "course_model"
SET "semesterId" = NULL
WHERE "semesterId" IS NOT NULL;

DELETE FROM "semester_model";
```

# How Has This Been Tested?

- [x] Updated implementations of organization and profile service unit tests and integration tests
- [x] Added integration test for semester endpoint (it is simple CRUD so the integration tests checking database changes is sufficient)

# Checklist:


packages/frontend/app/(dashboard)/components/ArchiveCourse.tsx
```diff
@@ -17,11 +18,28 @@ const ArchiveCourse: React.FC<ArchiveCourseProps> = ({
...
+                course: {
+                  ...userCourse.course,
+                  enabled: !courseData.course?.enabled,
+                },
+              }
+            }
+            return userCourse
+          }),
+        })
         fetchCourseData()
```
AdamFipke, line 43: Good catch with this!

packages/frontend/app/(dashboard)/components/ArchivedCoursesSection.tsx
```diff
@@ -0,0 +1,130 @@
...
+          dataSource={archivedCourses
+            .sort((a, b) => a.course.name.localeCompare(b.course.name))
+            .map((userCourse) => {
+              const semester = semesters.find(
+                (s) => s.id === userCourse.course.semesterId,
+              )
+              return { ...userCourse, semester }
+            })}
+          rowKey={(userCourse: UserCourse) => userCourse.course.id}
+          pagination={archivedCourses.length > 5 ? { pageSize: 5 } : false}
```
AdamFipke, line 122:
```suggestion
          pagination={archivedCourses.length > 10 ? { pageSize: 10 } : false}
```

5 is too small imo

packages/frontend/app/(dashboard)/organization/settings/page.tsx
```diff
@@ -124,215 +152,447 @@ export default function SettingsPage(): ReactElement {
...
+                          <EditOutlined />
+                          <span>Edit Logo</span>
+                        </button>
+                      </Row>
+                    </Col>
+                  </Form.Item>
+                </Form.Item>
+
+                <Form.Item label="Banner">
+                  <Form.Item name="organizationBanner" noStyle>
```
AdamFipke, line 421: I realise that it was like this before but ew nested form items gross haha

packages/frontend/app/(dashboard)/organization/settings/page.tsx
```diff
@@ -124,215 +152,447 @@ export default function SettingsPage(): ReactElement {
       })
   }
 
+  // for semester management
+  const [semesterForm] = Form.useForm()
```
AdamFipke, line 156: I noticed that this semester form, handleAddSemester, and handleEditSemester are here in org page rather than inside the SemesterModal, and I was just curious on why?

packages/frontend/app/(dashboard)/components/coursesSection.tsx
```diff
@@ -28,83 +45,282 @@ const CoursesSection: React.FC<CoursesSectionProps> = ({ courses }) => {
...
+                const diff =
+                  semesterB.endDate.valueOf() - semesterA.endDate.valueOf()
+                if (diff == 0) {
+                  return a.course.name.localeCompare(b.course.name)
+                } else {
+                  return diff
+                }
+              }
+              return 0
+            })
```
AdamFipke, line 207: yikes I'm noticing there are quite a few .finds, .filters, .somes, and more on the semester/courses pages.

If courses and semesters didn't have such a small `n` and thus performance doesn't matter too much, I would say I feel like some performance optimizations should be made. For example by doing just one pass and joining up semesters with usercourse just once. Or, better yet, do the join on the database. Could maybe even do some sorts on the database.

And the only reason why performance would really matter on this page is that this will be the first page that user's get dropped into from a fresh load (like if they go to `localhost:3000` while logged in they get redirected to `localhost:3000/courses`), meaning that Next.js wouldn't have had a chance yet to pre-fetch and load the page.

Though, like I said since `n` is very small this probably won't end making much of an impact until many years down the line so this should be fine

packages/server/src/organization/organization.controller.ts
```diff
@@ -450,182 +433,146 @@ export class OrganizationController {
...
-    if (!courseInfo) {
-      return res.status(HttpStatus.NOT_FOUND).send({
-        message: ERROR_MESSAGES.courseController.courseNotFound,
-      });
-    }
-
-    if (!courseDetails.name || courseDetails.name.trim().length < 1) {
-      return res.status(HttpStatus.BAD_REQUEST).send({
-        message: ERROR_MESSAGES.courseController.courseNameTooShort,
+    await getManager().transaction(async (manager) => {
```
AdamFipke, line 436: I like that you put these db calls in these endpoints into transactions, i gotta start doing that more

packages/server/src/semester/semester.controller.ts
```diff
@@ -1,16 +1,138 @@
...
+      where: { id: organizationId },
+    });
+
+    if (!organization) {
+      throw new BadRequestException('Organization not found');
+    }
+
+    const semesters = await SemesterModel.find({
+      where: { organizationId: organization.id },
+    });
```
AdamFipke, line 40:
```suggestion
   try {
    await OrganizationModel.findOneOrFail({
      where: { id: organizationId },
    });
    } catch {
       throw new BadRequestException('Organization not found');
    }

    const semesters = await SemesterModel.find({
      where: { organizationId: organizationId },
    });
```

I like to use findOneOrFail since I like to think it helps performance since typeORM doesn't need to convert it into a returned object but in reality it probably still does but eh take of this comment what you will

packages/server/src/semester/semester.controller.ts
```diff
@@ -1,16 +1,138 @@
...
+
+    return semesters;
+  }
 
-  @Get()
-  async get(): Promise<SemesterPartial[]> {
-    return SemesterModel.find({
-      take: this.appConfig.get('max_semesters'),
```
AdamFipke, line 13: should we keep this around? Dima added them a while back to help the use case where someone is malicious and tries to break the app by making many different entities (I think he used postman for it)

AdamFipke: wait i added it to my previous suggested changes

AdamFipke (approved):
Code looks good, just some minor comments.

I haven't manually tested it, but seeing as how it's not going on prod right away anyway it should be fine to merge and then fix any small things we may find after

packages/frontend/app/(dashboard)/components/ArchivedCoursesSection.tsx
patrickma6199: The reason I left it at 5 is because again, the UI will make a new one of these tables for each semester present in the courses you are registered for/instruct. I'm just imagining that overtime, having 10 instead of 5 with, say, three or four semesters, would start to require scrolling, kind of defeating the purpose of the pagination.

packages/frontend/app/(dashboard)/organization/settings/page.tsx
patrickma6199: I actually looked up some example code for antd's forms and saw this style quite a lot so I continued to adopt it. I could change it if it bothers you but it won't do anything to the UI regardless. I do agree tho that its a bit hard to read lmao.

packages/frontend/app/(dashboard)/components/coursesSection.tsx
patrickma6199: That's a fair point, ill look into precomputing this outside of the render, the reason I want more of this to be computed in the frontend though, is to avoid any issues with info being cached for the profiles. By now, its so interconnected that I am still worried about causing some unforeseen effects.

packages/frontend/app/(dashboard)/organization/settings/page.tsx
patrickma6199: Oh, I remember why I did this now that I'm looking at the code again. I wanted the modal to only handle the form so that I could reuse the same form for adding and editing semesters by simply passing a different function for onSubmit.

It also means I won't have to add ugly conditionals to the modal itself to change the behaviour from adding to editing since one also needs to populate the fields with the current values. Since I need to display the semester data here, I would still have to also pass in the semester array and return the modified one on success to update the actual semester cards.

Although this may not seem the cleanest, it was the most intuitive way to generalize the form for both use cases without extra conditional checks or making the modal component "do too much".

packages/frontend/app/(dashboard)/components/coursesSection.tsx
patrickma6199: I've moved this sort call to outside the component with a useMemo to memoize the result. Besides this though, this is not going to be a heavy computation until n is VERY large. There is also no other more efficient means of sorting the courses than this due to our database structure and I am not sure how much more efficient I can make it other than precomputing these sorted arrays to be mapped. Also, functional programming via chaining in JS is usually more efficient than implementing any of this by hand.

packages/server/src/organization/organization.controller.ts
patrickma6199: Yeah, its only for routes that make a lot of changes at once. Its actually really neat since it throws an error which by default causes a 500 internal error if not handled. Not that I haven't handled it but it definitely makes the code more resistant to unexpected behaviour.

packages/server/src/semester/semester.controller.ts
patrickma6199: Oh, I was not aware there was an upper limit in the org config. Do you still think its relevant? I'm sure we would ideally want to keep history for potentially longer.

packages/server/src/semester/semester.controller.ts
patrickma6199: Yea I see what you mean. I'm just thinking that because organization professors and admin are the only ones that can access this route, it would be safe. And admin can manually delete semesters now if they aren't necessary (something that wasn't present in the original implementation; only additions were supported). I think giving more of this control to the admin and org. professors instead of adding an upper bound is more applicable since semesters won't get created that often, and don't store much information per entity.

patrickma6199:
> Code looks good, just some minor comments.
> 
> I haven't manually tested it, but seeing as how it's not going on prod right away anyway it should be fine to merge and then fix any small things we may find after

Yeah, I'm still thinking about a dev branch so that main is always whats most up to date on prod for organization. Let me know what you think. It wouldn't be hard to make, just need to remember to merge to that instead of main.

packages/server/src/semester/semester.controller.ts
patrickma6199: I noticed this applies to some of my other routes in this controller too so I switched them up too. This is useful!

packages/frontend/app/(dashboard)/organization/settings/page.tsx
AdamFipke: Ah fair enough. Yeah react lets you do it different ways, and it's unclear if there's a strictly better way.

Usually I like to put the associated state and functions for a component inside the component itself, even if it means there would be some unclear logic inside the modal, just since I think the slight organisation increase may increase maintainability more than what is lost in weird logic. BUT tbh that's probably like a 1% difference so who cares haha

packages/server/src/organization/organization.controller.ts
AdamFipke: Ah yeah Nest.js automatically catches any uncaught errors. And if it's not like an HTTPException than it defaults to 500 level. It's pretty dang nice actually

PR #242, opened 2025-02-15
Event Details Tooltip

### LINKED ISSUE #212: Calendar event details popup
**Description**

So the calendar library we are using can make some very crowded calendars that are impossible to read. I would like to have a little popup of some kind whenever the user clicks on a calendar event to show all the details of it (maybe Popover, ideally not a full modal since it's very attention-grabbing)

Also the calendar library we are using isn't very flexible and I'm not a fan of its docs/api so this may seem harder than it looks.

Also it would probably be good to show any assigned TAs in this popup

AdamFipke:
# Description

Adds some pretty tooltips when you hover an event. This also adjusts the content body of the event to show more details like the location and location type.

image
image
This popconfirm was added to prevent accidentally hovering and clicking the link unintentionally.
image
It also shows assigned staff (if there are any)
image

This also adds tooltips to the TA Checkin Out Page
image
image

And you bet it works on mobile (it will just be on-tap rather than on-hover)

Closes #212 

## Type of change

# How Has This Been Tested?

This is mostly just a frontend change so a lot of manual testing. This includes seeing how the tooltips behave on desktop and mobile for both staff and students.

# Checklist:

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -799,6 +799,20 @@ export enum calendarEventLocationType {
   online = 'online',
   hybrid = 'hybrid',
 }
+export function getCalendarEventLocationTypeFormatted(
+  locationType: calendarEventLocationType,
+): string {
+  switch (locationType) {
+    case calendarEventLocationType.inPerson:
+      return 'In-Person'
+    case calendarEventLocationType.online:
+      return 'Online'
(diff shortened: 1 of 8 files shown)


bhunt02 (approved):
Changes are pretty simple - looks good

PR #249, opened 2025-02-28
1.4 changelog

AdamFipke:
# Description

Just putting it in a PR in advance 'cause I can

### DIFF
--- a/packages/frontend/public/changelog.md
+++ b/packages/frontend/public/changelog.md
@@ -1,3 +1,55 @@
+## Version 1.4.0 - Mar 2nd, 2025
+
+#### New Features
+- ***General***
+  - You can now leave comments on Anytime questions!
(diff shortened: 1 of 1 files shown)


patrickma6199 (commented):
Should we tack on queue chats and merge it before the prod push as well? Or is there still some tweaks to make?

patrickma6199 (commented):
Oh, and we could also briefly explain the changes to the async centre "update count" since it tracks differently now.

AdamFipke:
> Should we tack on queue chats and merge it before the prod push as well? Or is there still some tweaks to make?

That PR is ready to merge

> Oh, and we could also briefly explain the changes to the async centre "update count" since it tracks differently now.

Ah yeah fair point, i'll adjust it

PR #253, opened 2025-03-16
Fix Security on Chatbot Routes

### LINKED ISSUE #252: Fix route security on chatbot routes
**Description**

Since basically as long as your logged in, you can ask questions or upload documents into any course.

We probably need to restrict it so that only the helpme backend can call the chatbot routes (probably with some shared API key). This will come with some performance overhead but it's the only way to do it since the chatbot repo does not have access to user or course info

AdamFipke:
# Description

So the problem with the current token system is you just need to be logged in to get a token (and thus call any chatbot endpoint).
So, for example, i could just create an account (and never join any course) and then call the endpoints that delete all chatbot data for all of the courses (because that's an endpoint on the chatbot apparently lol).
So there are no role guards basically, nor any course guards (so i can access documents in questions in courses that i am not in).
So by adding a new chatbot api key (that only the backend of helpme has), users can't just call the chatbot endpoints directly, they have to go through the helpme system first (in which i can use CourseRolesGuard to protect the endpoints)

This also:
- Adjusts/adds more tooltips for helping users around the chatbot
- Cleaned up and consolidated a lot of chatbot types that should be a lot more accurate

Closes #252 

## Type of change

In `/server/.env`:
```
# Needed to make requests to the chatbot service in chatbot repo (needs to match the CHATBOT_API_KEY in the .env in the chatbot repo)
CHATBOT_API_KEY=changemepls
# Change this to the courseId of your "HelpMe" course (used by chatbot when asking system questions)
HELPME_COURSE_ID = 34
```

In Chatbot Repo `.env`
```
# All requests made to the chatbot repo will need this API key (make sure it matches the CHATBOT_API_KEY in the .env in the helpme repo)
CHATBOT_API_KEY=changemepls
```

In `/server/.env`:
```
# Needed to make requests to the chatbot service in chatbot repo (needs to match the CHATBOT_API_KEY in the .env in the chatbot repo)
CHATBOT_API_KEY=ForIBecomeBrokeFinancerOfOpenAICalls
# Change this to the courseId of your "HelpMe" course (used by chatbot when asking system questions)
HELPME_COURSE_ID = 
```

In Chatbot Repo `.env`
```
# All requests made to the chatbot repo will need this API key (make sure it matches the CHATBOT_API_KEY in the .env in the helpme repo)
CHATBOT_API_KEY=ForIBecomeBrokeFinancerOfOpenAICalls
```

# How Has This Been Tested?

So I didn't add many integration tests for this since I basically would have to mock the chatbot api, which means the only things I would really be testing for is whether the guards are working. And tbh I want to gamble that spending the time to make a bunch of tests just to check the roles guard isn't going to save much maintenance time (if any time) in the future. And I'm very confident that I implemented the guard correctly on each of the endpoints since I have made various courseRolesGuards. Or in short, I guess ~~I'm just lazy~~ there's just a lot of work to be done and this is not a priority right now.

I did manually test each route though, and also tested if the CourseRolesGuard worked.

# Checklist:

### DIFF
--- a/packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
@@ -17,7 +17,6 @@ import {
   RobotOutlined,
   CloseOutlined,
 } from '@ant-design/icons'
-import axios from 'axios'
 import { useCourseFeatures } from '@/app/hooks/useCourseFeatures'
 import { useUserInfo } from '@/app/contexts/userContext'
 import {
@@ -28,9 +27,6 @@ import {
 } from '@/app/utils/generalUtils'
 import { Feedback } from './Feedback'
 import {
-  PreDeterminedQuestion,
-  Message,
-  ChatbotAskResponse,
   chatbotStartingMessageSystem,
   chatbotStartingMessageCourse,
   ChatbotQuestionType,
@@ -39,7 +35,7 @@ import { API } from '@/app/api'
 import MarkdownCustom from '@/app/components/Markdown'
 import Link from 'next/link'
 import { usePathname } from 'next/navigation'
-import { Role } from '@koh/common'
+import { PreDeterminedQuestion, Role, Message } from '@koh/common'
 import { Bot } from 'lucide-react'
 
 const { TextArea } = Input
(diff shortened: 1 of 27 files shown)


packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -281,32 +243,6 @@ const Chatbot: React.FC<ChatbotProps> = ({
     setInput('')
   }
 
-  const getSourceLinkButton = (
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:284]: oh yeah so "getSourceLinkButton" was basically a react component so I renamed it for clarity.

But that's small. A bad thing here was that this component was defined inside the body of another component (nested component definitions), which will mean that all getSourceLinkButtons will always do a full unmount and re-mount every time there's a re-render of the outer Chatbot component (for example, it would completely unmount and remount every time the user types something in the text box).
It's mentioned here in the React docs: https://react.dev/learn/preserving-and-resetting-state

PR #255, opened 2025-03-24
Chatbot Document Display + Fix hardcoded db credentials

### LINKED ISSUE #254: Chatbot document display
instead of requiring the user to provide a source link, we handle it

AdamFipke:
# Description

Main changes
- Makes it so all uploaded chatbot documents will first get converted to pdfs, stores the pdf in the database, generated a pdf URL based on the id of the doc (for GET performance) and courseId, and then sends the pdf and generated URL to the chatbot server. 
- Added a new endpoint that allows you to GET one of these course documents, which then becomes the `href` for the source link, which will then show the pdf in the browser
  - I'm fairly certain this should work nicely in all browsers? I'll need to manually test how this performs (especially on (real) mobile). If it ends up like asking to download the pdf n whatnot, I might just switch to a frontend package to display the pdf
- For these pdf links, I added the `#page=` part to the link which will send the user to the referenced page of the chunk

Other Changes
- This also removes the option for users to give their own source link, since like I cannot imagine a use case where someone would rather give their own source link to a document rather than just have us handle it automatically. (also there's the weird thing where the sourceLink would be the same for all uploaded documents if you are bulk uploading)
- You can now upload images, .xlsx, .md, and some other formats for the chatbot
  - image formats have the `parseAsPng`/`isSlideDeck` option as true

This also fixes how our db credentials were hardcoded in the docker-compose and now use a `postgres.env` instead.
- instead of just adding the new variables to our pre-existing `.env`, I opted to create a new one since otherwise the postgres container will have all of our .env variables loaded, which could be bad if that container ever gets compromised.

Closes #254

## Type of change

In `postgres.env` in `/server`:
```env
POSTGRES_USER=changemne
POSTGRES_PASSWORD=changemne
POSTGRES_NONROOT_USER=changemne
POSTGRES_NONROOT_PASSWORD=changemne
```

In `postgres.env` in `/server`:
```env
POSTGRES_USER=postgres
POSTGRES_PASSWORD=mysecretpassword
POSTGRES_NONROOT_USER=helpme
POSTGRES_NONROOT_PASSWORD=mysecretpassword
```

- [x] This change requires you to re-run `yarn dev:db:up` (do after creating the new `postgres.env`)

# How Has This Been Tested?

Just a *lot* of manual testing since testing the upload and getDocument endpoints are going to be really awkward (especially since a big part of it depends on the chatbot api running).

I mean I guess adding some tests for it eventually could be good but man that's gonna take me at least a day or two.

# Checklist:


docker-compose.yml
```diff
@@ -1,23 +1,21 @@
-version: '2'
```
AdamFipke, line 1: i think there was a warning in the console about `version` being redundant and should be removed to reduce confusion so I did that

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/chatbot_settings/components/AddChatbotDocumentModal.tsx
```diff
@@ -253,31 +261,14 @@ const AddChatbotDocumentModal: React.FC<AddChatbotDocumentModalProps> = ({
               <Form.Item
                 name="isSlideDeck"
                 label="Parse document as slides"
-                tooltip="This will generate descriptions for images, which is particularly useful if the document is a slide deck or another image-heavy component. Only works for .pdf or .pptx"
+                tooltip="By default images/graphics embedded in your uploaded files will not be detected by the chatbot. Ticking this will transform pages of the document into images and automatically generate AI summaries of said images (using OpenAI's ChatGPT). This is useful for any document that isn't just text. Warning that it will take a lot longer to process."
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/chatbot_settings/components/AddChatbotDocumentModal.tsx:264]: Oh yeah I read through the chatbot repo and learned what was actually going on so I changed the tooltip.

packages/server/package.json
```diff
@@ -82,7 +83,8 @@
     "passport": "^0.7.0",
     "passport-jwt": "^4.0.0",
     "passport-saml": "^3.2.4",
-    "pg": "^8.2.1",
+    "pg": "^8",
+    "pg-query-stream": "^4.8.1",
```
AdamFipke, line 87: pg-query-stream is used for streaming data out of the database. Typeorm actually requires it but didn't warn us about not having it until I actually tried to use .stream() so that was neat

PR #274, opened 2025-05-14
fix(invite): add option to decline course invitation (#273)

### LINKED ISSUE #273: [Small] Invite page -> No thanks (takes to /courses)
This would just be a simple addition to add a button on the invite page that takes you to /courses.

*Usually* if you are on the invite page, you almost always want to accept the invite. But sometimes cookies will be set and you will just be taken to it by accident and so you might want an option to just say "no thanks"

Akshat-Kalra:
# Description

Fixes Issue #273
 
Adds a "Decline Invitation Button" to the invite page taking the user back to the `/courses` route.

Achieved by adding 2 optional (`secondaryButtonLabel` and `secondaryButtonAction`) props to the  `InviteCard` component which only renders if both the label and the action is provided.

No new dependencies required

Closes #273 

## Type of change

# How Has This Been Tested?

I logged in with an account with no courses and with the UBCO Organization, then went to `/invite?cid=37&code=invite-code` route on localhost, which is the invite link for CS 304 in the UBCO Org, then manually tested it.

# Checklist:

### DIFF
--- a/packages/frontend/app/invite/page.tsx
+++ b/packages/frontend/app/invite/page.tsx
@@ -150,6 +150,10 @@ export default function CourseInvitePage(): ReactElement {
                 }
                 await addStudent(userData)
               }}
+              secondaryButtonLabel="Decline Invitation"
+              secondaryButtonAction={() => {
+                router.push('/courses')
+              }}
               cover={
                 <Image
                   alt="generic course image"
--- a/packages/frontend/app/invite/components/InviteCard.tsx
+++ b/packages/frontend/app/invite/components/InviteCard.tsx
@@ -8,6 +8,8 @@ type InviteCardProps = {
   buttonAction: () => void
   description?: { title?: string; text?: string }
   cover?: ReactElement
+  secondaryButtonLabel?: string
+  secondaryButtonAction?: () => void
 }
 
 const InviteCard: React.FC<InviteCardProps> = ({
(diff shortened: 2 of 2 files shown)


packages/frontend/app/invite/page.tsx
```diff
@@ -150,6 +150,10 @@ export default function CourseInvitePage(): ReactElement {
                 }
                 await addStudent(userData)
               }}
+              secondaryButtonLabel="Decline Invitation"
+              secondaryButtonAction={() => {
+                router.push('/courses')
+              }}
```
AdamFipke, line 156: *technically* speaking putting routing actions inside an onClick like this isn't recommended (for webcrawlers like for SEO, and I believe it might impact accessibility) and instead the button should have an `href` attribute, but meh I don't think it would matter that much in this case (since we don't need SEO and since it's still a Button and not like a div it should still be fairly obvious for screen readers)

AdamFipke (approved):

packages/frontend/app/invite/page.tsx
AdamFipke: Oh and I guess one more thing that's related that I should mention. Next.js's `<Link>` component could also be used to wrap an antd Button, that way it takes advantage of the fancy Next.js things (most notably being prefetching the /courses page before the user clicks on it). 

Not that I think that change really needs to be made since it's rare that this would be clicked on, but something i'd just thought might be a useful thing for future :)

PR #275, opened 2025-05-22
Feature#270: User chatbot history

### LINKED ISSUE #270: User chatbot history
**Description**

Right now, students don't have a way to see what chatbot questions they asked in the past. 

To add this, you will probably just want to create a new endpoint to get all the interactions for their user ids (each interaction has 1 to many chatbot questions) and then send those to the frontend. 

For the place to put this on the frontend, you will probably want to make it a tab under the profile page by these:
Image

For the page itself, you could choose to copy parts from the chatbot_questions page.tsx. However, you likely don't need to make it as complicated as I made that page (e.g. you don't need to group conversations that had the same question together, you don't need to show verified, suggested, times asked, or the user score. You only really the question text, the answer, the source documents, and when it was asked). You can also choose to do something else instead of a table, but the key part is that it needs to both show the information and should also probably be expandable so they need to click something to see the entire conversation).

Also, technically speaking chatbot questions are stored in both the chatbot repo and helpme repo, and each contains different information for a chatbot question (which can get a little messy when trying to join), though luckily for this I don't think you will need the chatbot questions from the chatbot repo. You should only need to query the interactions and chatbot questions from the helpme repo side and display those.

Also, I don't think there is any other actions that the user needs to be able to do (e.g. they don't need to edit or delete chatbot questions), only displaying them is needed. Though, if you wanted to be extra, you could add a feature to allow them to retroactively thumbs up or thumbs down their  question (see Feedback.tsx for how it's done in the chatbot modal).

Akshat-Kalra:
# Description
Added a feature for users to look at their past chatbot history

Resolves Issue #270

Looks like this:
Screenshot 2025-05-22 at 11 48 04 AM

Closes #270

## Type of change

- [X] This requires a run of `yarn install`

# How Has This Been Tested?

 Integration Test:
- Wrote an integration test in the `server/test/chatbot.integration.ts` file and tested for correct behaviour and also expected errors if the endpoint is called with a different userId than that of the user.

Manual Testing: 
- Just interacted with the chatbot, then reset the chat and interact more to have multiple conversations. Then go to the profile section and check this tab to see your chatbot history.

# Checklist:


packages/frontend/app/(dashboard)/profile/components/UserChatbotHistory.tsx
```diff
@@ -0,0 +1,116 @@
...
+        }
+      })
+      .finally(() => setLoading(false))
+  }, [userInfo?.id])
+
+  return (
+    <div className="mx-auto max-w-3xl">
+      <h2 className="mb-4 text-2xl font-bold">Your Chatbot Conversations</h2>
+      {loading ? (
+        <div className="py-8 text-gray-500">Loading...</div>
```
AdamFipke, line 51: For this i'd recommend using the CenteredSpinner component

packages/frontend/app/(dashboard)/profile/components/UserChatbotHistory.tsx
```diff
@@ -0,0 +1,116 @@
...
+    API.chatbot.studentsOrStaff
+      .getChatHistory(userInfo.id)
+      .then((res: GetChatbotHistoryResponse) => {
+        setHistory(res.history)
+      })
+      .catch((err) => {
+        if (err?.response?.status === 403) {
+          setError("You are not allowed to access this user's history.")
+        } else {
+          setError('An error occurred while fetching your chatbot history.')
```
AdamFipke, line 41: for this I would recommend using the getErrorMessage function (from /app/utils/generalUtils)

packages/frontend/app/(dashboard)/profile/components/UserChatbotHistory.tsx
```diff
@@ -0,0 +1,116 @@
...
+      ) : (
+        <Collapse accordion>
+          {history.length === 0 && (
+            <div className="text-gray-500">No chatbot history found.</div>
+          )}
+          {history.map((interaction) => (
+            <Panel
+              header={
+                <Tooltip
+                  title={formatDateAndTimeForExcel(
```
AdamFipke, line 63: I love that you found this function haha. I made it a while back for export course data to csv and then ended up using it in a few other areas afterwards anyways. Just a fun little thing haha

packages/frontend/package.json
```diff
@@ -29,6 +29,7 @@
     "class-transformer": "^0.5.1",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.1",
+    "dayjs": "^1.11.13",
```
AdamFipke, line 32: huh interesting, we've actually been using dayjs in a handful of files but I'm surprised to see that we didn't actually have it as part of our package.json dependencies. I guess we have some packages that already use dayjs as a subpackage so it didn't throw an error when it was imported before. Probably a good thing that it's included here now

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -169,6 +170,25 @@ export class ChatbotController {
     );
   }
 
+  @Get('history/:userId')
+  @Roles(Role.PROFESSOR, Role.TA, Role.STUDENT)
+  async getChatbotHistory(
+    @Param('userId', ParseIntPipe) userId: number,
+    @User() user: UserModel,
```
AdamFipke, line 177: Hmm, I was going to make a comment that instead of passing a userId as a param to instead just use the UserId of the user calling the endpoint (thus, calling /history will just return the logged in user's chatbot history). Though it doesn't really make too much of a difference either way.

However, I would suggest using the `@UserId` decorator instead of the `@User` decorator since it will be 1 less database query.

AdamFipke: Oh right, also the `@Roles(Role.PROFESSOR, Role.TA, Role.STUDENT)` decorator does not do anything unless there is a guard to consume those roles (usually either CourseRolesGuard or OrganizationRolesGuard). 

In this case, I would just remove the `Roles` decorator altogether since you can only get your own chatbot history in its current implementation.

In the future, there may be potential to make an admin version of this endpoint that lets admins see other users' chatbot history (in which case you would do `@Roles(OrganizationRole.ADMIN)` with the `OrganizationRolesGuard`), but yeah hopefully that clarifies things with the roles

packages/server/test/chatbot.integration.ts
```diff
@@ -103,4 +103,86 @@ describe('ChatbotController Integration', () => {
...
+        'responseText',
+        questionData.responseText,
+      );
+      expect(firstQuestion).toHaveProperty(
+        'vectorStoreId',
+        questionData.vectorStoreId,
+      );
+    });
+
+    it('should not return the chatbot history for another user if the user is not a TA or Professor', async () => {
```
AdamFipke, line 157:
```suggestion
    it('should not return the chatbot history for another user', async () => {
```

packages/frontend/app/(dashboard)/profile/components/UserChatbotHistory.tsx
Akshat-Kalra: Resolved

packages/frontend/app/(dashboard)/profile/components/UserChatbotHistory.tsx
Akshat-Kalra: Resolved

packages/frontend/app/(dashboard)/profile/components/UserChatbotHistory.tsx
Akshat-Kalra: :)

packages/frontend/package.json
Akshat-Kalra: 👍

AdamFipke (changes requested):
Overall, this is surprisingly good for a first full-stack PR! Nice job! Only a handful of small comments really

packages/server/test/chatbot.integration.ts
Akshat-Kalra: Oh yes, at first I was trying to make the endpoint such that a professor or a TA could call it with any userId but not a student, hence the `userId` param in the endpoint. But that didn't really work out probably because of the Role decorator.

I'll fix this text, but maybe I'll leave out the endpoint as is with the `userId` param, in hopes of fixing it in the future?

packages/server/src/chatbot/chatbot.controller.ts
Akshat-Kalra: > Hmm, I was going to make a comment that instead of passing a userId as a param to instead just use the UserId of the user calling the endpoint (thus, calling /history will just return the logged in user's chatbot history). Though it doesn't really make too much of a difference either way.
> 
> However, I would suggest using the `@UserId` decorator instead of the `@User` decorator since it will be 1 less database query.

Well I was trying to make it more generalisable so that a prof or a ta could call the endpoint for another userId, in theory it really only needs a check of some sort to very the Role, but it wasn't really working out when I tried today so I just removed it for now.

But I think we should keep the `userId` param of the endpoint as is, in hopes of maybe one day improving it. This would also remove the need for the admin version of this endpoint. What do you think about that?

Akshat-Kalra: > However, I would suggest using the `@UserId` decorator instead of the `@User` decorator since it will be 1 less database query.

Resolved

Akshat-Kalra: > In this case, I would just remove the `Roles` decorator altogether since you can only get your own chatbot history in its current implementation.

Resolved

Akshat-Kalra:
> Overall, this is surprisingly good for a first full-stack PR! Nice job! Only a handful of small comments really

Thanks :)

packages/server/src/chatbot/chatbot.controller.ts
AdamFipke: > Well I was trying to make it more generalisable so that a prof or a ta could call the endpoint for another userId, in theory it really only needs a check of some sort to very the Role, but it wasn't really working out when I tried today so I just removed it for now.

Yeah that makes sense, though from my experience developing on this system, I would say it is easier to have different endpoints each with different roles rather than having one mega endpoint that has different logic based on the role of who's calling it (we've had quite a few of the latter in this system and some end up being huge and hard to follow).

So for this example, I would have the following endpoints:
- `GET /chatbot/history` named getMyChatbotHistory (no role protections)
- `GET /chatbot/history/:userId` named getUserChatbotHistory (only admin role)
- `GET /chatbot/history/:userId/:courseId` named getUserChatbotHistoryForCourse (only profs and TAs inside said course can call this endpoint)

There's probably some design principle going on at play that makes me want to make the endpoints this way

AdamFipke (approved):

packages/server/src/chatbot/chatbot.controller.ts
Akshat-Kalra: Got it!

I went ahead and changed the endpoint to be just `/history` now :)

AdamFipke:
oh you got cursed with the like 1/20 chance that github actions fails due to a deadlock error lol (this error in particular has been plaguing this project for years haha, no one has quite dug in and found what was causing it)

Akshat-Kalra:
> oh you got cursed with the like 1/20 chance that github actions fails due to a deadlock error lol (this error in particular has been plaguing this project for years haha, no one has quite dug in and found what was causing it)

Lol, I was wondering what caused that

Akshat-Kalra:
> I re-ran the job so you get the little green check mark :)

Thanks, hopefully its good to merge now :)

PR #293, opened 2025-06-04
Improve RedisProfile Cache Busting

AdamFipke:
# Description

So I found a bug where after creating an account and verifying my email, it doesn't clear the redis profile cache so I am just stuck there waiting for the 30mins or whatever for the cache to bust itself (also for more context, the redis profile caching is NOT on prod yet, so no actual users were affected by this bug).

And so rather than just adding another cache bust to that endpoint, I remembered that we have these typeorm EventSubscribers that we can use to connect database events with external systems. 

So, I opted to remove all manual cache busting and instead added EventSubscribers for all relevant updates so that they bust the caches automatically whenever they update.

Not only does this fix any cases we might've missed with the original redis profile cache, but this will also make it more maintainable for the future so that you don't have to remember to bust the cache for like any course update (though if someone were to add a new relation to the profile get() request, then they will need to add a new subscriber for events of that relation)

## Type of change

# How Has This Been Tested?

I only manually tested updating the user profile (changing name and then refreshing page) and it seems to work. I also tried updating and deleting semesters and that seems to work.

There's technically more to test but eh this will be tested as I make the guides for the system.

# Checklist:

### DIFF
--- a/packages/server/src/profile/profile.service.ts
+++ b/packages/server/src/profile/profile.service.ts
@@ -18,10 +18,14 @@ import {
 import { UserModel } from './user.entity';
 import { pick } from 'lodash';
 import { OrganizationService } from '../organization/organization.service';
-import * as checkDiskSpace from 'check-disk-space';
+import * as checkDiskSpaceModule from 'check-disk-space';
 import * as path from 'path';
 import * as fs from 'fs';
-import * as sharp from 'sharp';
+import * as sharpModule from 'sharp';
+
+const checkDiskSpace =
+  (checkDiskSpaceModule as any).default || checkDiskSpaceModule;
+const sharp = (sharpModule as any).default || sharpModule;
 
 @Injectable()
 export class ProfileService {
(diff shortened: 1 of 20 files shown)


packages/server/src/profile/profile.service.ts
```diff
@@ -16,20 +16,20 @@ import {
   ServiceUnavailableException,
 } from '@nestjs/common';
 import { UserModel } from './user.entity';
-import { RedisProfileService } from '../redisProfile/redis-profile.service';
 import { pick } from 'lodash';
 import { OrganizationService } from '../organization/organization.service';
-import checkDiskSpace from 'check-disk-space';
```
AdamFipke, line 22: I was having an issue where I was getting checkDiskSpace.default was not a function/undefined when trying to upload a pfp, but weirdly enough this caused the tests to fail since the tests kinda want it the other way (I guess due to mocking)

bhunt02 (approved):
Haven't tested manually, but the additions look well thought out, the code is clean, and I don't see any flaws in the logic.
Also. THANK YOU for using an Observer pattern implementation for the Redis profile. The related functions were being called in far too many places with the same args.

Akshat-Kalra (approved):
Looks good!!🙌

Issue #285, opened 2025-05-27
Anytime questions - Would you like your post to be public


AdamFipke:
**Description**

Right now, faculty have the option of making an anytime question private or public (to other students)

Instead, when creating a question, students should have the option there on whether to post private or public (default to public). It should also state that you will appear anonymous to other students.

Not sure whether to keep allowing faculty to mark a question as public or private. Probably best to keep allowing this, but to add a tooltip or warning that "this student chose to make their question private, are you sure you want to make it public? They will still appear anonymous to other students". Though, perhaps some legacy flag is needed since that tooltip would be inaccurate for all current anytime questions as they all start as private.

Note that work here may conflict with #149

PR #295, opened 2025-06-05
Anytime Questions: Require Author Permission for Public AQ, Anonymity Settings

### LINKED ISSUE #284: Anytime Questions - Would you like to appear anonymous
**Description**

Rather than forcing all students to appear anonymous, maybe make it optional (default is anonymous).

This would also need to be done for anytime question comments too.

Note that work here may conflict with #149

bhunt02:
# Description

Added Anytime Question support for requiring both Staff to respond to an AQ and Author permission to make an Anytime Question public. Additionally added AQ anonymity toggle.

Rundown:
* Comes with two new Course Features: 
  * asyncCentreDefaultAnonymous: whether questions are anonymous by default (true) or not (false)
  * asyncCentreAuthorPublic: whether authors can set their anytime question to be public (default = false)
* Alterations to logic in creating, updating and retrieving Anytime Questions and their comments made
* asyncCentreAuthorPublic = true:
  * When staff goes to post response and sets it to be public, if the author didn't indicate they want it to be public, they'll be served a popconfirm warning (really just flavor)
* Anytime Question comments can also be toggled to be anonymous or not, except when written by question author (set to anonymity of question)
* When Anytime Question comments are updated (anonymity setting), all previous comments on the same question by the same user have their anonymity setting changed to match
* Small bug fixes and code errors where I noticed them

Closes #284 and #285 

## Type of change

  - possibly

# How Has This Been Tested?

Tested with three clients (browsers): professor and/or TA, a student, and another student.
Tested the behaviour when the anonymity and visibility changes across clients.

- [x] Added integration tests in addition to previously existing ones
- [x] Updated existing integration tests and snapshots where necessary

# Checklist:

  - ?? maybe


packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
```diff
@@ -47,11 +57,17 @@ const Comment: React.FC<CommentProps> = ({
   const [isEditing, setIsEditing] = useState(false)
   const [newContent, setNewContent] = useState(content)
   const [editLoading, setEditLoading] = useState(false)
+
+  const [newAnonymous, setNewAnonymous] = useState<boolean>(isAnonymous)
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:61]:
```suggestion
  // Form state that allows the user to specify whether or not they would like their comment to be anonymous
  const [newAnonymous, setNewAnonymous] = useState<boolean>(isAnonymous)
```

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
```diff
@@ -305,6 +329,15 @@ const Comment: React.FC<CommentProps> = ({
           >
             Save
           </Button>
+          {!IAmStaff && !(author.isAuthor && isSelf) && (
+            <Switch
+              className={'mx-2'}
+              checked={newAnonymous}
+              checkedChildren={'Anonymous'}
+              unCheckedChildren={'Non-Anonymous'}
+              onChange={() => setNewAnonymous(!newAnonymous)}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:338]: Maybe instead of a switch, a checkbox might look better?
image

Like it would be: Post Anonymously? [ x ]

packages/common/index.ts
```diff
@@ -841,6 +840,9 @@ export type AsyncQuestion = {
   aiAnswerText?: string
   closedAt?: Date
   visible?: boolean
```
AdamFipke, line 842: should the old visible attribute be removed?

packages/frontend/app/(dashboard)/components/CourseFeaturesForm.tsx
```diff
@@ -77,6 +77,22 @@ const CourseFeaturesForm: React.FC<CourseFeaturesFormProps> = ({
...
+          courseId={courseId}
+        />
+
+        <CourseFeatureSwitch
+          featureName="asyncCentreAllowPublic"
+          defaultChecked={courseFeatures.asyncCentreAllowPublic}
+          title="Require Author Permission to make Anytime Questions Public"
+          description="When toggled, requires both the author of an Anytime Question and a staff member to toggle the question's visibility. If not toggled, only a staff member's permission is required to make an Anytime Question visible to all members of the course."
+          courseId={courseId}
+        />
```
AdamFipke, line 95: So with this, I don't think I really see the value in having this as an option? Wouldn't you just always want the ability to make posts public as a staff member, and wouldn't this just be limiting your own permissions? I guess I could see the case where you might want to limit the permissions of your TAs but the feature should be renamed/adjusted to only apply to TAs (though that may be better suited for #298 ).

I was moreso thinking that if the student asked for their question to remain private and a staff member tried to do this toggle, a warning popup would show up saying "Are you sure you want to make this public? The student initially made their question private. They will be anonymous to other students"
image

I sorta mentioned this in the description #285

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx
```diff
@@ -279,6 +301,32 @@ const CreateAsyncQuestionModal: React.FC<CreateAsyncQuestionModalProps> = ({
           <QuestionTagSelector questionTags={questionTypes} />
         </Form.Item>
       )}
+      <Form.Item
+        name="setAnonymous"
+        label="Hide name and avatar from other users?"
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx:306]:
```suggestion
        label="Hide name and avatar from other users?"
        layout="horizontal"
```
I would add this to both of these switches

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx
```diff
@@ -142,11 +142,11 @@ const PostResponseModal: React.FC<PostResponseModalProps> = ({
...
-        name="visible"
+        name="staffSetVisible"
         label={
           <div className="flex flex-row items-center gap-1">
             Set question visible to all students
-            <Tooltip title="Questions can normally only be seen by staff and the student who asked it. This will make it visible to all students (the student themselves will appear anonymous to other students)">
+            <Tooltip title="Questions can normally only be seen by staff and the student who asked it. This will make it visible to all students as long as the author also sets it to be visible.">
               <QuestionCircleOutlined style={{ color: 'gray' }} />
             </Tooltip>
           </div>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx:152]: So I realised something with this. I guess this whole part of the form should be re-made:

If the student initially set their question to private, this checkbox would look similar to how it does now except with a warning popup (or a red Alert that appears beneath it) that warns the staff member that the student initially set it to private.

BUT, if the student initially set their question to *public*, perhaps the UI should compensate and allow the professor to override it to private.

I think the way to do this is to rework this into first having some text that states what the student set the visibility of it to, with an override checkbox for the staff beneath it. sorta like so:

---

Student posted this *privately* - hidden from other students (appears red, "publically" would appear blue maybe. Also have a question mark icon with a tooltip that says "the student posted or modified this question to be private").
Override visibility?: Leave as is | Private | Public  (use an antd Segmented for this)
(The tooltip for this would be "you can override the visibility status of this question to control if other students are allowed to see it. If being made public, the student will appear anonymous to other students) 

---

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -230,15 +255,40 @@ export class asyncQuestionController {
...
     }
 
+    const myUserCourse = await UserCourseModel.findOne({
+      where: {
+        userId: userId,
+        courseId: question.courseId,
+      },
+    });
+    const myRole = myUserCourse.role;
+
```
AdamFipke, line 265: I think this can be replaced with a use of the `CourseRole` decorator (not to be confused with the CourseRole**s** decorator ;P)

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx
```diff
@@ -279,6 +301,32 @@ const CreateAsyncQuestionModal: React.FC<CreateAsyncQuestionModalProps> = ({
...
+          isStaff
+            ? 'Anytime Questions authored by staff members do not have optional visibility. To make it publicly visible, finish creating the question and then post a response to it.'
+            : courseFeatures?.asyncCentreAllowPublic
+              ? 'If toggled, your question will be visible to other users. Your profile will only be visible if you do not opt for anonymity.'
+              : 'This course does not require approval by question authors for questions to be public, only approval by course staff members.'
+        }
+        valuePropName="checked"
+      >
+        <Switch disabled={!courseFeatures?.asyncCentreAllowPublic || isStaff} />
+      </Form.Item>
```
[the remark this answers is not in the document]
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx:329]: Also, I would still allow staff to post questions "anonymously", since, again, that's mostly just there to allow them to test the system. 

Though, in the off chance they're not just using it to test the system and they do actually want to post like a "welcome" post or something on here, then they will still have the option to just turn off Appear Anonymous when they go to post it

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -230,15 +255,40 @@ export class asyncQuestionController {
...
+      )
+        return;
       if (body[key] !== undefined && body[key] !== null) {
         question[key] = body[key];
       }
     });
 
     const updatedQuestion = await question.save();
 
+    // If anonymity setting changed, update anonymity of any comments by the author
```
AdamFipke, line 281: hey nice catch!

packages/server/src/course/course.controller.ts
```diff
@@ -281,7 +280,7 @@ export class CourseController {
         throw new NotFoundException('Your course enrollment is not found');
 
       userCourse.favourited = !userCourse.favourited;
-      userCourse.save();
+      await userCourse.save();
```
AdamFipke, line 283: lol nice

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:338]: i'll switch to a checkbox

packages/frontend/app/(dashboard)/components/CourseFeaturesForm.tsx
bhunt02: my impl was intended more to protect student privacy; they might not want their questions shared (anonymous or not). i'll discuss this with you though to see if we can come to a conclusion rather than just making it a `staffSetVisible || authorSetVisible` situation which just makes the student capable of making it public. 

an additional reason i went this direction is because what if a student asks a question that's completely inappropriate, then gets to set authorSetVisible to true, and staff have no say whether or not other students see it before it is. i think at that point, we don't want to give students permission to set things to be visible; we want staff to have the final say.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx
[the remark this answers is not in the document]
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx:329]: i'll remove the blurb, but i don't see the use in making the anonymous toggle dynamically appear; we should present the user with the options they have from the get-go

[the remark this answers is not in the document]
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx:329]: the separate logic for staff members is because i interpreted it as a way for staff members to post a 'frequently asked question' answer so that students can see it, see that it was made by staff, etc. i guess if it's just for testing it out, i can revert that... but because staff members shouldn't be able to be anonymous (why would they need to be), and because they can never set 'authorSetVisible' later on, it doesn't make sense to give access to these parts of the form

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx:152]: re: initially set question to public, see earlier comments as to why i made it require both staff & author to set it to be public. i understand the idea here, but i think it unnecessarily overcomplicates the UI. information for the staff member as to what the student opted for in terms of question publicity is a good idea though.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx
```diff
@@ -267,39 +277,54 @@ const AsyncQuestionCard: React.FC<AsyncQuestionCardProps> = ({
...
+                        {question?.isAnonymous && (
+                          <Tooltip
+                            title={
+                              isStaff
+                                ? 'The author of this question will appear anonymous to non-staff.'
+                                : 'You will appear anonymous to other students who see this question.'
+                            }
+                          >
+                            <Tag color={'default'}>Anonymous</Tag>
+                          </Tooltip>
```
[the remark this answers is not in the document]
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/AsyncQuestionCard.tsx:325]: i'll remove it

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
bhunt02: whoops, i'll fix where i've used it

packages/server/src/course/course.controller.ts
bhunt02: it was silly

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/Comment.tsx:61]: only if 'newContent' gets a corresponding one because wym this one needs one but that one doesn't

packages/frontend/app/(dashboard)/components/CourseFeaturesForm.tsx
AdamFipke: > an additional reason i went this direction is because what if a student asks a question that's completely inappropriate, then gets to set authorSetVisible to true, and staff have no say whether or not other students see it before it is. i think at that point, we don't want to give students permission to set things to be visible; we want staff to have the final say.

I think that's a fair point, though doesn't other platforms like piazza and canvas also have this problem where students can just post inappropriate things to other students?

Though, this would certainly be a problem if staff didn't have moderation tools but they *do* have the ability to delete questions if need be, though with the "override visibility" thing I suggested below, they could also just override it to be private too.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx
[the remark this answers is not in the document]
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx]: > i don't see the use in making the anonymous toggle dynamically appear; we should present the user with the options they have from the get-go

I guess the idea is to hide irrelevant UI, since Appear Anonymous is only relevant if the student plans to Post Publicly, and less text to read and less decisions to be made is less cognitive load required to make a post.

Though, I can see the value in showing all form options right away. You could go the disabled route and just disable the Appear Anonymous checkbox if Post Publicly is false, but I would do two things:
- indent Appear Anonymous so that it looks like a sub-option of Post Publically (so it's clear the two are linked)
- This is a small thing, but whenever Post Publicly is false, make the checkbox for Appear Anonymous look to be "checked" (since antd checkboxes can be disabled when checked or unchecked, and I don't want someone to see Appear Anonymous as unchecked when Post Publicly is also unchecked since the user might get anxiety that their privacy is in danger since appear anonymous is unchecked and disabled)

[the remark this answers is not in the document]
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/CreateAsyncQuestionModal.tsx]: > but because staff members shouldn't be able to be anonymous (why would they need to be)

I'm mostly thinking that if I'm a professor and demonstrating the system to my students, I want them to see that when they post a question, they can appear anonymous if they want. 

Also, if I'm a new professor to the system, I want to be able to test out things to see what certain things do. And if I'm not allowed to interact with those UI elements, I'm probably not going to get a good understanding of how they work.

Also, tbh, I don't think many professors at all are going to use this Staff Post Question thing outside of testing. Since, if they wanted to, they would probably just make a canvas announcement or a suggested question in the chatbot.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx:178]: Hang on lemem explain my reasoning on why I think an override might be better rather than doing authorSetVisible && staffSetVisible:
- If I'm a professor and I see a really good student question but they posted it as private, I can't make it public. And I'm basically SOL without just recreating the Q&A myself.
- "Should the anytime question/answer be visible to other users once responded to?" is a lot more confusing to understand imo than "Post Publicly ". But the problem with renaming it to "Post Publicly" is it's inaccurate in its current implementation since I as a student could check "Post Publically" and save it just to see that it was not made public.
- If a student marks their question as private, and I as a staff member mark it public and save, nothing appears to change on my end because it's not indicated to me that the student must also be public, and thus the UI needs to be made more complicated anyway to indicate this
- Requires more maintenance and time from professors since they must log in to the system quite often just to make questions public (that students were fine with being public anyway). This is more maintenance than something like Piazza, where even if a professor does not go on it for weeks, students can post questions and help each other via comments.

Though with the override, a simple checkbox wouldn't work since sometimes you would want to override a public question to make it *hidden* (probably rare case, but giving profs the moderation tools they need is good), hence why I went with a Segmented to say: Leave as is | Private | Public

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx
```diff
@@ -83,6 +99,19 @@ const PostResponseModal: React.FC<PostResponseModalProps> = ({
...
+              authorCanSetVisible &&
+              !hasCheckedPopconfirm &&
+              question.authorSetVisible != staffSetVisible
+            ) {
+              setConfirmPopoverOpen(true)
+            } else {
+              onFinish()
+            }
+          })
+        },
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx:114]: does this onClick do anything? Since I noticed I can still click Finish even if I haven't closed the popconfirm. Also, an alternative way is to disable the okButton if that popconfirm is open.

AdamFipke (changes requested):
Overall, really really good! I guess I left you some comments in discord (probably could've put them here but oh well). Code looks pretty good for the most part I think. All the comments at this point are minor

AdamFipke (changes requested):
Overall, really really good! I guess I left you some comments in discord (probably could've put them here but oh well). Code looks pretty good for the most part I think. All the comments at this point are minor

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx:122]: it overrides the form's default onClick interaction I believe, which is a submit action

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx
```diff
@@ -109,6 +138,24 @@ const PostResponseModal: React.FC<PostResponseModalProps> = ({
...
+              open={confirmPopoverOpen}
+              arrow={false}
+              okText="Yes"
+              cancelText="No"
+              onConfirm={() => {
+                onFinish().then()
+                setConfirmPopoverOpen(false)
+              }}
+              onCancel={() => setConfirmPopoverOpen(false)}
+            ></Popconfirm>
```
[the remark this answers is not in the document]
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/PostResponseModal.tsx:166]: it does open! if the student changes their visibility setting _after a staff member posts the question_, and the staff member does the same. it's just a weird edge case

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -410,13 +459,35 @@ export class asyncQuestionController {
...
+          : (body.isAnonymous ??
+            courseSettings?.asyncCentreDefaultAnonymous ??
+            true),
       createdAt: new Date(),
     }).save();
 
+    for (const other of otherComments) {
+      other.isAnonymous = comment.isAnonymous;
+      await other.save();
+    }
```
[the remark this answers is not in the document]
bhunt02: meh, if we're really hurting for performance maybe in the future this is something to keep in mind. but it's a simple UPDATE interaction and not too costly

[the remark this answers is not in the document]
AdamFipke: well it's more so how it's an update for each comment. So if there's like 30 comments that's 30 update queries. But like i said i'm not sure maybe typeorm is smart enough to not run an update if there's nothing to update

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -487,11 +480,31 @@ export class asyncQuestionController {
...
+      await AsyncQuestionCommentModel.createQueryBuilder()
+        .update()
+        .set({
+          isAnonymous: comment.isAnonymous,
+        })
+        .where({
+          creatorId: user.id,
+          questionId: qid,
+        })
+        .execute();
```
AdamFipke, line 503: So, since this otherComments query is after the comment is created, it includes the newly created comment.

So, otherComments.length > 0 is always true.
And it checks if the first comment (which is arbitrary and could just be the newly created comment) is anonymous and see if it's different than the newly created comment, then update all the anonymous values.

I think all that's needed here is to add `id: Not(comment.id)` to the otherComments query so that it doesn't include the newly created one. Could also change the if statement to use .find and loop through every comment

PR #297, opened 2025-06-06
akshat/discussion-board-integration

### LINKED ISSUE #290: Integration of Discussion Board for Canvas LMS
**Description**

To integrate the discussion board from the Canvas LMS

Ideas:

- Have another tab for `Discussion` beside these, on the `LMS Integration` page.
Image

- This is needed because some instructors get creative with their discussion board uses, some use them to have students submit work, and in some cases even provide additional information that is relevant to the course

- Responses should not be included, as these may contain unverified/sensitive data from students. Only basic details such as topic title, topic content, dates, etc. should be included.

- This integration should be similar to the `assignments` where only fields like the `title`, `description` and maybe other details are added

- This would also need a check so that only discussion posts written by instructors are captured and embedded in  the chatbot.

**Success Criterias**

-   step 1
-   step 2
-   ...

**Context**

-   Figma:
-   Branch: akshat/improve-canvas-integration

Akshat-Kalra:
# Description

Modified the current announcement endpoint, it now gets the instructor ids and adds them to a set, then gets all the discussion posts and then filters those using the id set and then combines them with the announcement data.

Note: Also the reason it is double fetching from the same endpoint is because apparently the `only_announcements` field is set to false by default and it does not return announcements at all.

Closes #290

## Type of change

# How Has This Been Tested?

mostly manual testing

# Checklist:

### DIFF
--- a/packages/server/src/lmsIntegration/lmsIntegration.adapter.ts
+++ b/packages/server/src/lmsIntegration/lmsIntegration.adapter.ts
@@ -198,12 +198,38 @@ class CanvasLMSAdapter extends ImplementedLMSAdapter {
     status: LMSApiResponseStatus;
     announcements: LMSAnnouncement[];
   }> {
-    const { status, data } = await this.GetPaginated(
-      `courses/${this.integration.apiCourseId}/discussion_topics?only_announcements=true`,
+    const { status: instructorStatus, instructorIds } =
+      await this.getInstructorIds();
+
+    // Get announcements using only_announcements=true
+    const { status: announcementStatus, data: announcementData } =
+      await this.GetPaginated(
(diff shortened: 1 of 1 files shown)


Akshat-Kalra:
Just added this change to the UI aswell

Screenshot 2025-06-06 125940

Akshat-Kalra:
Updated the label so that its more explicit that it's only the instructor discussion posts

image

FabianLCH (approved):
All good with me too!

bhunt02 (approved):
good implementation for now; i don't see any issues that might arise from these changes. i do think maybe in the future we'll have to reorganize, perhaps add an "isAnnouncement" boolean column to the LMSAnnouncement entity; that way we could separate the tabs easily and the title wouldn't take up so much space. but good job!

PR #304, opened 2025-06-12
LMS Resource Type Selector

### LINKED ISSUE #303: Resource Type Selector for LMS Integration
**Description**

Add a resource type selector for Canvas that lets the instructor select the type of resource they want the chatbot to have access to.

**Success Criterias**

- Able to select amongst all type of resources ideally
- Have to think about granularity of the selection, i.e. just resource types or specific assignment/annoucement.
- UI change would be needed in the LMS Integration section of the Course Settings Page for Instructors.

**Context**

Reference for the UI

Image

Akshat-Kalra:
# Description

Added this LMS resource type selector that lets the instructor select the type of resource they want the chatbot to have access to.

image

TODO: ~~hv to fix the UI for mobile, will remove this note when completed :)~~
TODO: ~~also have to make the checkboxes persist state across renders~~
TODO: write integration tests

Edit: Also made this UI change to the individual document selector

image

Closes #303 

## Type of change

# How Has This Been Tested?

Manually for now but going to write integration tests as well.

# Checklist:

### DIFF
--- a/packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx
@@ -51,7 +51,11 @@ export default function LMSDocumentList<
   const [page, setPage] = useState(1)
   const [input, setInput] = useState('')
   const [search, setSearch] = useState('')
-  const [performingSyncOp, setPerformingSyncOp] = useState<boolean>(false)
+  const [syncingItems, setSyncingItems] = useState<Record<string, boolean>>({})
+
+  const isItemSyncing = (itemId: string) => {
+    return syncingItems[itemId] || false
+  }
 
   const getNameCell = (item: T) => {
     const isOutOfDate =
@@ -95,43 +99,62 @@ export default function LMSDocumentList<
           )}
         </div>
         {isOutOfDate && <Badge count={'Out of Date'} />}
-        {!ineligible && (
-          <Popover
-            content={
-              (lmsSynchronize && !item.syncEnabled && (
-                <Button
-                  onClick={() => toggleSyncDocument(item, type)}
-                  icon={<SyncOutlined />}
-                  color={'blue'}
-                  variant={'outlined'}
-                  loading={performingSyncOp}
-                >
-                  {item.uploaded != undefined
-                    ? 'Enable synchronization'
-                    : 'Force synchronization'}
-                </Button>
-              )) ||
-              (item.syncEnabled && (
-                <Button
-                  onClick={() => toggleSyncDocument(item, type)}
-                  icon={<DeleteOutlined />}
-                  color={'danger'}
-                  variant={'outlined'}
-                  loading={performingSyncOp}
-                >
-                  {!lmsSynchronize
-                    ? 'Delete document'
-                    : 'Disable synchronization'}
-                </Button>
-              ))
-            }
-            trigger={'click'}
-          >
-            <Button type={'text'} icon={<MoreOutlined />} />
-          </Popover>
-        )}
       </div>
     )
+  }
+
+  const getSyncCell = (item: T) => {
+    let ineligible = false
+    if (type == 'Assignment') {
+      const asAssignment = item as LMSAssignment
+      ineligible =
+        (asAssignment.description == undefined ||
+          asAssignment.description == '') &&
+        asAssignment.due == undefined
+    }
+
+    if (ineligible) {
+      return <span className="italic text-gray-500">Not eligible</span>
+    }
+
+    if (lmsSynchronize && !item.syncEnabled) {
+      const tooltipText =
+        item.uploaded != undefined ? 'Enable sync' : 'Force sync'
+      return (
+        <Button
+          onClick={() => toggleSyncDocument(item, type)}
+          icon={<SyncOutlined />}
+          color={'blue'}
+          variant={'outlined'}
+          loading={isItemSyncing(item.id.toString())}
+          size="middle"
+          className="h-8 w-full"
+          title={tooltipText}
+        >
+          <span className="hidden md:inline">{tooltipText}</span>
+        </Button>
+      )
+    }
+
+    if (item.syncEnabled) {
+      const tooltipText = !lmsSynchronize ? 'Delete' : 'Disable sync'
+      return (
+        <Button
+          onClick={() => toggleSyncDocument(item, type)}
+          icon={<DeleteOutlined />}
+          color={'danger'}
+          variant={'outlined'}
+          loading={isItemSyncing(item.id.toString())}
+          size="middle"
+          className="h-8 w-full"
+          title={tooltipText}
+        >
+          <span className="hidden md:inline">{tooltipText}</span>
+        </Button>
+      )
+    }
+
+    return null
   }
 
   const columns = useMemo(() => {
@@ -165,6 +188,12 @@ export default function LMSDocumentList<
                 ></div>
               )) || <i>No message</i>,
             colSpan: 3,
+          },
+          {
+            dataIndex: 'sync',
+            header: 'Sync?',
+            cellFormat: (item: T) => getSyncCell(item),
+            colSpan: 1,
           },
         ] as LMSDocumentListColumn[]
       case 'Assignment':
@@ -201,11 +230,17 @@ export default function LMSDocumentList<
               )) || <i>No description</i>,
             colSpan: 4,
           },
+          {
+            dataIndex: 'sync',
+            header: 'Sync?',
+            cellFormat: (item: T) => getSyncCell(item),
+            colSpan: 1,
+          },
         ] as LMSDocumentListColumn[]
       default:
         return []
     }
-  }, [getNameCell, type])
+  }, [getNameCell, getSyncCell, type])
 
   const ncols = useMemo(
     () => columns.reduce((acc, column) => acc + column.colSpan, 0),
@@ -260,6 +295,8 @@ export default function LMSDocumentList<
     doc: LMSAssignment | LMSAnnouncement,
     type: 'Assignment' | 'Announcement',
   ) => {
+    const docIdStr = doc.id.toString()
+
     const thenFx = (result?: string) => {
       if (result) {
         message.success(result)
@@ -274,10 +311,17 @@ export default function LMSDocumentList<
     }
 
     const finallyFx = () => {
-      setPerformingSyncOp(false)
+      setSyncingItems((prev) => ({
+        ...prev,
+        [docIdStr]: false,
+      }))
     }
 
-    setPerformingSyncOp(true)
+    setSyncingItems((prev) => ({
+      ...prev,
+      [docIdStr]: true,
+    }))
+
     switch (type) {
       case 'Announcement':
         return await API.lmsIntegration
@@ -340,7 +384,7 @@ export default function LMSDocumentList<
               <div
                 key={`column-${index}`}
                 className={cn(
-                  'border border-gray-100 p-2',
+                  'overflow-hidden border border-gray-100 p-2 py-3',
                   col.colSpan == 2 ? 'col-span-2' : '',
                   col.colSpan == 3 ? 'col-span-3' : '',
                   col.colSpan == 4 ? 'col-span-4' : '',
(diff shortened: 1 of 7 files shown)


packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -391,6 +424,112 @@ export default function CourseLMSIntegrationPage(props: {
       })
     }
 
+    const renderSynchronizationOptions = () => (
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:427]: oh btw while you *can* put component definitions inside other components like this, it's not recommended with react since it will essentially completely re-generate this renderSyncronizationOptions component everytime the parent page component re-renders (which happens when its state changes) instead of using the already generated dom elements. 

This happens because react essentially has a cache with the components it generates. So when a component is made this way, react is unable to see that its returning the same component each time and thus doesn't cache it. 

This causes two problems: performance (due to many more unnecessary calculations) and state resets (if renderSyncronizationOptions or any of its subcomponents has any state, they will be reset to their default everytime this page re-renders)

In this example, the sub Tooltip components have their own state for controlling whether or not they are open. But due to what I described above, if any state on the page changes, it will reset these tooltips to their default state (closed). So, you could test this by hovering a tooltip and clicking the Force Synchronization button (triggering a state change on the page), and you will notice that the tooltip will instantly close with no animation (because it's completely re-generating the component).

Anyways, to fix this, you can either:
- Move the component definition outside this page component (and then make it take in a bunch of props)
- Delete this component definition and move it to where you called it below so that it's with the other DOM pieces

Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:427]: And for more context, this was an issue I suffered at one point, where I had a nested component definition and one of its subcomponents was a little form but the form kept resetting itself. It wasn't until I got some ideas from reading the react docs that i decided to get rid of the nested component definition and that fixed it.

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -57,6 +61,13 @@ export default function CourseLMSIntegrationPage(props: {
   const [modalOpen, setModalOpen] = useState<boolean>(false)
   const [delModalOpen, setDelModalOpen] = useState<boolean>(false)
   const [isTesting, setIsTesting] = useState<boolean>(false)
+  const [selectedResources, setSelectedResources] = useState<unknown[]>([])
+
+  const onChange: GetProp<typeof Checkbox.Group, 'onChange'> = (
+    checkedValues,
+  ) => {
+    setSelectedResources(checkedValues)
+  }
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:70]: since this is a pretty busy page, maybe onSelectedResourcesChange could be a better name

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -57,6 +61,13 @@ export default function CourseLMSIntegrationPage(props: {
   const [modalOpen, setModalOpen] = useState<boolean>(false)
   const [delModalOpen, setDelModalOpen] = useState<boolean>(false)
   const [isTesting, setIsTesting] = useState<boolean>(false)
+  const [selectedResources, setSelectedResources] = useState<unknown[]>([])
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:64]:
```suggestion
  const [selectedResources, setSelectedResources] = useState<string[]>([])
```

not sure if this works

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -453,124 +592,76 @@ export default function CourseLMSIntegrationPage(props: {
...
+                            <div className={'flex flex-col items-center gap-2'}>
+                              <Checkbox.Group
+                                style={{ width: '100%' }}
+                                onChange={onChange}
+                              >
+                                <Row gutter={[0, 20]}>
+                                  <Col span={10}>
+                                    <Checkbox value="Assignments">
+                                      Assignments
+                                    </Checkbox>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:618]: I may have missed it, but I don't see the states of these being set from data from the backend. Meaning that if you were to check/uncheck one of these checkboxes and save and then refresh the page, it will reset back even though it was saved in the database.

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -453,124 +592,76 @@ export default function CourseLMSIntegrationPage(props: {
...
+                                    ? 'blue'
+                                    : 'default'
+                                }
+                                icon={<SyncOutlined />}
+                                disabled={!integration.lmsSynchronize}
+                                style={{ marginTop: '30px' }}
+                                onClick={handleSaveAndResync}
+                                loading={syncing && integration.lmsSynchronize}
+                              >
+                                Save and Re-Sync Documents
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:658]: I would disable this button if what they have selected in the database is the same as what's already selected (or in other words, they have not modified it).

packages/server/src/lmsIntegration/lmsIntegration.controller.ts
```diff
@@ -564,4 +565,42 @@ export class LMSIntegrationController {
 
     return `Successfully ${newState ? 'synced' : 'cleared'} document from ${integration.orgIntegration.apiPlatform ?? 'LMS'} in HelpMe.`;
   }
+
+  @Post('course/:courseId/resources')
+  @UseGuards(JwtAuthGuard, CourseRolesGuard)
+  @Roles(Role.PROFESSOR)
+  async updateSelectedResourceTypes(
+    @User() _user: UserModel,
```
AdamFipke, line 573:
```suggestion
```

Akshat-Kalra:
Looks decent on mobile now aswell. The text is short only for mobile view

image

Akshat-Kalra:
@AdamFipke Also do you think any sort of formal testing is needed for this additions? Lmk, thanks :)

Akshat-Kalra:
> Hmm, well I guess I could see some tests, either service or integration, for testing the new logic on the lms services to make sure it works (that way it's easy to find when it breaks from future changes). And seeing as how this also runs through a cron job, and thus sometimes users or us won't know if it fails, it might be a good idea. Though maybe I'll leave this call up to @bhunt02
> 
> As for the new updateSelectedResourceTypes, yeah I would probably add a handful of basic tests for it. Make sure only logged in users and only professors in the course can access it, and that it is actually updating the fields in the database (and that those fields in the database are _not_ getting updated when the users try to do something malicious).

Got it

bhunt02:
> Just had a chance to go over the changes and try them out in my dev environment. Overall, the resource selection logic and the added UI look good. Below are a few things I noticed:
> 
> * Selecting currently unsupported resources (anything other than Assignments and Announcements) and saving the selection will show a success message and trigger a re-sync, but it will not save any unsupported resources to the DB
>   
>   * This is probably a temporary limitation while new resource types are implemented
> * The "Course Assignments" and "Course Announcements..." tabs (`LMSDocumentList` component) have the same placeholder text for the search input
>   
>   * Should be easily fixable by referring to the component's `type` prop
> * Trying to search for an announcement results in HTTP 500
>   
>   * Seems to happen in the `LMSDocumentList` component in the `matchingDocuments` function when the code tries to convert  `LMSAnnouncement.posted` property to a locale date string
>   * Issue seems to be present in prod too
> * Looks like the "Force Synchronization" feature can silently fail and still display a 'sync success' message
>   
>   * An example of this is when the chatbot service isn't available, then calls to the chatbot service to update/clear documents that fail

W.R.T. the last point, that's something that was done partially intentionally. documents can fail to sync (albeit rarely) at any time due to a myriad of factors out of our control(such as Canvas outtages); the original interface was developed with a much more involved sync interaction with full lists of errored documents but after presentation it was reworked to be a lot lighter weight. imo, if a document continues to say "not synced" it's a better indication of failure than inundating the prof's screen with a long list of all the failed operations (and reasons for failure), especially since the chatbot service should never be offline

Akshat-Kalra:
> * Selecting currently unsupported resources (anything other than Assignments and Announcements) and saving the selection will show a success message and trigger a re-sync, but it will not save any unsupported resources to the DB
>   
>   * This is probably a temporary limitation while new resource types are implemented

I added a strikethrough the resource types for now, until they are implemented with the resource selector

image

Akshat-Kalra:
> Trying to search for an announcement results in HTTP 500
> * Seems to happen in the `LMSDocumentList` component in the `matchingDocuments` function when the code tries to convert  `LMSAnnouncement.posted` property to a locale date string
> * Issue seems to be present in prod too

Also fixed

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx
```diff
@@ -95,45 +99,64 @@ export default function LMSDocumentList<
...
-            trigger={'click'}
-          >
-            <Button type={'text'} icon={<MoreOutlined />} />
-          </Popover>
-        )}
       </div>
     )
   }
 
+  const getSyncCell = (item: T) => {
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx:106]: this is a nested component definition and thus is going to have some state and/or performancing issues. I would move this into its own component outside of the LMSDocumentList component and then update the props accordingly (perhaps put the new component just at the end of the file). Though, I guess I just realised that getNameCell also has this issue, and renderDocumentList... and you didn't write any of those things so... idk. Maybe put it into a todo

AdamFipke:
> > * Selecting currently unsupported resources (anything other than Assignments and Announcements) and saving the selection will show a success message and trigger a re-sync, but it will not save any unsupported resources to the DB
> >   
> >   * This is probably a temporary limitation while new resource types are implemented
> 
> @FabianLCH I added a strikethrough the resource types for now, until they are implemented with the resource selector
> 
> image

Hmm with this, I am a little worried some prof is going to find this and go "why can't I toggle this", so maybe putting a little tooltip for the disabled ones may not be a bad idea.

AdamFipke (approved):
Yeah looks pretty good, good work! Though I'll admit I only gave a brief thought on the logic but it seems alright, and if @bhunt02 said logic is good than it should be good.

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx
Comment (Akshat-Kalra) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx:106]: oh also by putting it into a TODO, should i make it like an issue or just mark it a TODO in the code and write a brief description?

Comment (Akshat-Kalra) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx:109]: nvm, I added it as a top-level TODO in that file, maybe i'll come back to it later and can fix that.

FabianLCH:
> the original interface was developed with a much more involved sync interaction with full lists of errored documents but after presentation it was reworked to be a lot lighter weight. imo, if a document continues to say "not synced" it's a better indication of failure than inundating the prof's screen with a long list of all the failed operations (and reasons for failure), especially since the chatbot service should never be offline

Ah, got it. Agree on the current implementation being more user-friendly than a detailed list of errors that a professor likely wouldn't go through. I also now realize even if the sync isn't fully successful, there shouldn't be any big effects (it would be more akin to a temporary interruption) and future sync attempts should take care of any pending changes.

> a change to accomodate this without going overboard with specific messages would be changing the return behaviour to return the number of failed docs (or something similar) and using that with a message.warn/success/error ANTD call to present that to the user, e.g., "Some documents failed to sync"

I think this would work well. @Akshat-Kalra suggested turning this into a separate issue/PR to address after merging this PR which might be better to avoid bundling up too much into a single PR and tackle one thing at a time

FabianLCH (approved):
All good on my end! I'd maybe suggest disabling the checkboxes for the pending resource types in addition to displaying the tooltips, but it's nothing major.

Akshat-Kalra:
> All good on my end! I'd maybe suggest disabling the checkboxes for the pending resource types in addition to displaying the tooltips, but it's nothing major.

Done

PR #306, opened 2025-06-23
Refactor Chatbot Service (Sister PR)

bhunt02:
# Description

Sister PR to the [Chatbot Refactor](https://github.com/ubco-db/chatbot/pull/34).

Makes changes to the following:
* Frontend chatbot components to accommodate changes to what routes return (if applicable) (e.g., it is possible that addDocumentChunk and related functions/routes return multiple; it's an array return type now. Had to be factored in!)
* LMS document changes: since the content may be split into multiple, now generates a prefix from the LMS document and sends it along with the document text. This ensures that each document chunk produced is stamped with crucial information to relate (e.g., assignment name/announcement title, due/posted dates, page link)
  * LMS document ID is now an array; the migration file has SQL which will populate the existing LMS document's array with the original chatbot document ID; this isn't that important since the re-embedding process defined in the chatbot repository deletes the original chunk for any inserted LMS documents (as they will eventually, due to CRON job, be resynchronized)
  * Alterations to logic to support the likely presence of multiple chunks per document (as LMS documents are much larger than typically inserted chunks) has been accounted for
  * Added a CLI command to perform the CRON job manually. We'll probably want to use this after running the regenerate embeddings command from the chatbot repository.
* Docker compose configuration change: added 'chatbot_test' database to list of databases to create on startup as it is needed for the chatbot repository's test suites.
* For devs and for prod, can just run:

```sql
CREATE DATABASE chatbot_test;
GRANT ALL PRIVILEGES ON DATABASE chatbot_test TO <user>;
ALTER DATABASE $db OWNER TO <user>;
```
* User will depend on dev or prod. For dev, this is probably fine just being that found in the POSTGRES_NONROOT_USER environment variable.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] This requires a run of `yarn install`
- [ ] This change requires an addition/change to the production .env variables. These changes are below:
- [ ] This change requires developers to add new .env variables. The file and variables needed are below:
- [x] This change requires a database query to update old data on production. This query is below:
  - **Either run the migration or the code that is within it.**
```sql
CREATE DATABASE chatbot_test;
GRANT ALL PRIVILEGES ON DATABASE chatbot_test TO <user>;
ALTER DATABASE $db OWNER TO <user>;
```

# How Has This Been Tested?

[...]

PR #311, opened 2025-06-25
implemented Pages integration for Canvas LMS; closes issue #299

### LINKED ISSUE #299: Integration of Pages for Canvas LMS
**Description**

Integrating the Pages resource type from the Canvas LMS

**Success Criterias**

- Might involve creating new data entities similar to the lms models
- Also would require looking into how the content inside it is embedded with the chatbot.
- Would also require some UI changes, hopefully just a new tab along the existing LMS resources in the LMS settings page for instructors

Akshat-Kalra:
# Description

Pages as a resource type are now supported just like assignments, announcements and instructor discussion posts.

Here's what the UI element looks like 

image

TODO:
- [x] Integration Test
- [ ] ~~Looking into the chatbot replying with embedded image in inner html~~ I'll add this to the parent issue
- [x] Also making the resource selector work for Pages once that PR is merged
- [x] Oh also, write the migration file before merging

Closes #299 
## Type of change

# How Has This Been Tested?

Manually for now, will be modifying the existing integration tests for canvas documents and making relevant changes to include pages aswell.

# Checklist:

### DIFF
--- a/packages/server/src/lmsIntegration/lmsIntegration.service.spec.ts
+++ b/packages/server/src/lmsIntegration/lmsIntegration.service.spec.ts
@@ -5,6 +5,7 @@ import {
   LMSUpload,
 } from './lmsIntegration.service';
 import { Test, TestingModule } from '@nestjs/testing';
+import { CACHE_MANAGER } from '@nestjs/cache-manager';
 import { TestConfigModule, TestTypeOrmModule } from '../../test/util/testUtils';
 import {
   AbstractLMSAdapter,
@@ -37,6 +38,13 @@ import { FactoryService } from 'factory/factory.service';
 import { ChatbotModule } from '../chatbot/chatbot.module';
 import { ChatbotApiService } from '../chatbot/chatbot-api.service';
 
+const mockCacheManager = {
+  get: jest.fn(),
+  set: jest.fn(),
+  del: jest.fn(),
+  reset: jest.fn(),
+};
+
 /*
 Note:
   The majority of methods in the LMSIntegrationService require external API calls.
@@ -58,6 +66,10 @@ describe('LMSIntegrationService', () => {
         LMSIntegrationService,
         LMSIntegrationAdapter,
         ChatbotApiService,
+        {
+          provide: CACHE_MANAGER,
+          useValue: mockCacheManager,
+        },
       ],
     }).compile();
 


packages/server/src/lmsIntegration/lmsPage.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+  id: number;
+
+  @PrimaryColumn()
+  courseId: number;
+
+  // Would rather this be enum, but evolving enums are not well-supported...
+  @Column({
+    type: 'text',
+  })
+  lmsSource: LMSIntegrationPlatform;
```
AdamFipke, line 24: I don't have any comments here since I haven't had to bang my head against the wall with enums much but iirc @bhunt02 has, so idk bridgette if you have anything to mention here

AdamFipke (commented):
I like what I'm seeing so far! I don't really see any issues with it

packages/server/src/lmsIntegration/lmsPage.entity.ts
Akshat-Kalra: Oh this comment was already there in the entity files for the other resources as well, and I just kept the comment to keep it consistent.

bhunt02: > I don't have any comments here since I haven't had to bang my head against the wall with enums much but iirc @bhunt02 has, so idk bridgette if you have anything to mention here

nothing exactly to mention here, since the platforms may expand in definition, typeorm will wreak havoc when a potential enum type changes (sometimes causing updates/migrations to fail entirely even if they would otherwise make sense), since it won't intuitively alter the column. i think even if it does, it's going to error when it tries to cast it to the enum type afterward probably. so we keep it text to be safe and consistent

bhunt02:
not reviewing quite yet, but you'll see I've made some changes in #306 to accommodate for the chatbot refactor and in merging with main after you merged #304 in, i did notice some discrepancies afterward and made those fixes. so I'm going to wait for my PR to be merged in before reviewing, so you have the chance to pull it into your branch

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx
```diff
@@ -291,6 +333,12 @@ export default function LMSDocumentList<
           .then(thenFx)
           .catch(errFx)
           .finally(finallyFx)
+      case 'Page':
+        return await API.lmsIntegration
+          .toggleSyncPage(courseId, doc.id, doc as LMSPage)
```
Comment (FabianLCH) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx:338]: Looks like this function has not yet been added to the frontend API, which causes the page synchronization status toggle to fail!

Comment (FabianLCH) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/components/LMSDocumentList.tsx:338]: Also, the endpoint that would called as a result (`toggleSyncDocument` in `lmsIntegration.controller.ts`) does not accept a `docType` of `"page"`, which would make it default to an HTTP 400 error

FabianLCH (commented):
Tried out the changes locally. Other than what I mentioned in my previous comments, functionality looks good so far!

packages/server/src/lmsIntegration/lmsIntegration.service.ts
```diff
@@ -718,6 +793,11 @@ export class LMSIntegrationService {
         documentText = `${a.description && `\nDescription: ${convert(a.description)}`}`;
         break;
       }
+      case LMSUpload.Pages: {
+        const p = item as LMSPage;
+        documentText = `(Course Page)\n Title: ${p.title}\nContent: ${p.body}\n${p.frontPage ? 'Front Page: Yes' : 'Front Page: No'}\n${!isNaN(new Date(p.modified).valueOf()) ? `Modified: ${new Date(p.modified).getTime()}` : ''}`;
+        break;
```
bhunt02, line 799: See the above implementations and add a prefix instead of including it all in the documentText variable. We define a prefix with the metadata information because the document text may be long enough that it suffices being split into multiple chunks, and we can provide the information for this page across all the chunks by using the prefix impl I added

Akshat-Kalra: resolved!

packages/server/src/lmsIntegration/lmsIntegration.controller.ts
```diff
@@ -314,6 +316,13 @@ export class LMSIntegrationController {
     );
   }
```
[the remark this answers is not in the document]
Akshat-Kalra: resolved

bhunt02 (approved):
All the functionality works as intended, it blends well with the changes to how LMS documents are saved as well. Good job!

I would have left a comment asking to improve upon the load times with the pages by using Promise.all or a similar method, like the 'If-Modified-Since' header, but I did my due diligence and tried out both solutions. Neither worked, as the Canvas API doesn't seem to implement that header, and Promise.all caused each fetch to fail for some reason.

I'd maybe add another property to the 'useLMSIntegration' hook to keep track of the 'pages' being loaded, since they take much, much longer than the assignments and announcements/discussion posts due to the nature of what we're doing (multiple GET requests to retrieve full documents, that are not cached by Canvas)

A way you might be able to implement it is by using CacheManager, which you can add by doing the following:
* Add `CacheModule.register()` to the imports array in the LMS Integration module
* Add it as a provider to the LMSService: `@Inject(CACHE_MANAGER) private cacheManager: Cache` (note that the import for the type of cacheManager will need to be from `cache-manager`, not the default TypeScript definition for Cache) 
* Add it as a constructor argument to the LMS Adapters, and as an argument for the `getAdapter` function
* Use some key (e.g., the API url being contacted) and cache the values for a short time once retrieved: `this.cacheManager.set(url, { data and such }, ttl)` (ttl is measured in milliseconds)
* Retrieve it with the key you decide

This is all optional, but I thought I'd mention it as a solution. You can see an example of how we use CacheManager in the Insights module.

FabianLCH (approved):
Tried out everything post big refactor merge, all functionality looks and works well on my end!

AdamFipke (approved):
gave it another real quick look, I don't see any issues!

Akshat-Kalra:
> We can think about implementing a separate boolean flag to keep the 'pages' section loading separate from the rest later (which would be good to implement so the rest of the LMS page is accessible whilst the pages, which are slowest, are still being retrieved before the first cache)

Also resolved now!

PR #315, opened 2025-07-02
(Proposal) Allow Semesters To Be Nullable

bhunt02:
# Description

**Note:** This is more of a 'proposal' to structure changes; we don't _have_ to merge it, but it makes things simpler.
The only issues it may cause is if users clone courses without changing information, but this shouldn't happen frequently. If this happens and users are having trouble, we/future system administrators can intercede as well.

Rationale: not every organization is going to use semesters, or section IDs, but should have access to these tools regardless
* Also avoids having to backfill and assign a relevant semester to the courses on production

* frontend changes to accommodate semesters being strictly nullable
* backend changes to accommodate semesters being strictly nullable
* course cloning operations allow for both setting of default section ID and semester, but neither is required
* removed some legacy/deprecated models and their interactions

~~Closes # (issue number)~~

## Type of change

# How Has This Been Tested?

* Modified test cases as needed

# Checklist:


packages/server/src/course/course.service.ts
```diff
@@ -532,20 +455,25 @@ export class CourseService {
...
-      if (cloneData.useSection) {
-        clonedCourse.sectionGroupName = cloneData.newSection;
-        clonedCourse.semester = originalCourse.semester;
-      } else if (cloneData.useSection === false) {
-        const semester = await manager.findOneOrFail(SemesterModel, {
-          where: { id: cloneData.newSemesterId },
-        });
-        clonedCourse.semester = semester;
-        clonedCourse.sectionGroupName = originalCourse.sectionGroupName;
+      //console.log(originalCourse)
```
AdamFipke, line 458:
```suggestion
```

packages/server/test/course.integration.ts
```diff
@@ -1115,46 +1105,6 @@ describe('Course Integration', () => {
...
-      const professor = await UserFactory.create();
-      const semester = await SemesterFactory.create();
-      const potato = await CourseFactory.create({
-        name: 'Potato',
-        semester: semester,
-      });
-      const tomato = await CourseFactory.create({
-        name: 'Tomato',
-        semester: semester,
-      });
```
AdamFipke, line 1128: RIP (I have never seen these tests before lmaoo)

AdamFipke (changes requested):
Overall I think the changes make sense. But if we're allowing courses to just not have semesters, I think I would prefer if we removed the little tag on each course card (under /courses) that says "No Semester" for students (or I guess org members) since that information is irrelevant to them (the only reason that check wasn't there before was the assumption that all courses would have a semester), it's a small change.

Also, just to clarify, when we do the prod push, we are  still going to create new semesters and assign all the old courses to the semester that corresponds with whatever string semester they had previously, as mentioned in the discord, right? For archival/organisational purposes

packages/server/test/course.integration.ts
bhunt02: death to them all

bhunt02:
> Overall I think the changes make sense. But if we're allowing courses to just not have semesters, I think I would prefer if we removed the little tag on each course card (under /courses) that says "No Semester" for students (or I guess org members) since that information is irrelevant to them (the only reason that check wasn't there before was the assumption that all courses would have a semester), it's a small change.
> 
> Also, just to clarify, when we do the prod push, we are still going to create new semesters and assign all the old courses to the semester that corresponds with whatever string semester they had previously, as mentioned in the discord, right? For archival/organisational purposes so we can hide courses where their semester has ended

can do on the first part; was thinking that as I was doing it but thought I'd leave it up to you when you were reviewing

so the thing is they're already technically assigned to semesters, that were backfilled, that just won't display properly since they're not organization semesters (no linked organization, which was weird to see), are you fine with leaving those or do you still want me to figure out something for that?

AdamFipke (approved):

PR #326, opened 2025-07-14
Convert Queue Question to Anytime Question

### LINKED ISSUE #69: Unanswered questions optionally converted to anytime questions after queue ends
**Description**

So right now when a queue session ends (the last TA checks out), the PrompStudentToLeaveQueueModal will show up and ask the student if they would like to stay in the queue or leave. There should also be an option to convert their queue question into an anytime question (so that they can get an AI answer that the prof or TA can see later). Doing so should also open up a modal that lets the student modify their question before posting (since I noticed queue questions can usually be a little brief or lack detail).

I have already implemented a similar feature where students can convert their chatbot questions into anytime questions if they're unsatisfied with their chatbot answer, and thus this can be implemented similarly.

I think this can purely be a frontend feature (you probably won't need to adjust or add anything to the backend).

You will want to edit PromptStudentToLeaveQueueModal.tsx and add a new button (that's actually a Link) that takes the user to the asyc_centre page with a query param, kinda like so:

```tsx
                <Link
                    href={{
                      pathname: `/course/${cid}/async_centre`,
                      query: { convertQueueQ: true },
                    }}
                  >
                    Convert to anytime question
                  </Link>
```

On the async_centre page.tsx, you can add logic that checks for this "convertQueueQ" query param and will automatically open the modal to convert the queue question (I did a very similar thing with converting chatbot questions to anytime questions, so you should be able to find similar code for this there).

You will also want to create a new modal component "ConvertQueueQToAntimeQModal.tsx", and you can use ConvertChatbotQToAnytimeQModal.tsx as a base. You can probably do something similar where the question abstract is the queue question (trimmed to 100 chars) and you can set the question body to the full question text (if the queue question is greater than 100 chars). Note that unlike the ConvertChatbotQToAnytimeQModal, you will likely always want to get a AI answer (so you can remove the code around getting a new AI answer, or alternatively just name the checkbox to "get an AI answer" with it default to true). 

Also don't forget to delete the students' questions out of the queue when you convert them!

---

As a little implementation detail around how the backend for how the PrompStudentToLeaveQueueModal popup works: If they choose to stay, a cron job is set that will check again in 10mins, and if there's still no staff it will prompt the student again asking if they would like to leave or stay. If they don't respond to the popup for 10 minutes, they will be kicked out of the queue (allowing the queues to clean themselves out basically when the last TA checks out). You shouldn't need to adjust any of this logic for this issue.

Eneshjakhar:
# Description

This PR implements the "Convert Queue Question to Anytime Question" feature, allowing students to convert their queue questions into anytime questions for more detailed, asynchronous answers from course staff. The feature includes AI-generated abstracts, tag merging logic, and seamless conversion workflow.

**Key Features:**
- Modal interface for converting queue questions to anytime questions
- AI-powered abstract generation with cleanup and fallback logic
- Merged question types from both anytime and queue sources
- Automatic queue question deletion and alert cleanup
- Consistent UI/UX with existing chatbot conversion modal

**Motivation:** Students often need more detailed answers than what's possible during office hours.  The office hours might end early, so the student needs an option to always convert their queue question to an anytime Question. This feature bridges the gap between real-time queue questions and comprehensive anytime questions.

**Dependencies:** None - uses existing API endpoints and components.

Closes #69 

## Type of change

# How Has This Been Tested?

- [x] **Test A: Modal Opening** - Navigate to `/course/{cid}/async_centre?convertQueueQ=true&qid={questionId}` and verify modal opens with pre-filled data
- [x] **Test B: AI Abstract Generation** - Verify AI generates appropriate abstracts and fallback logic works for long/invalid responses
- [x] **Test C: Tag Merging** - Confirm anytime question tags take priority over queue tags, with case-insensitive duplicate detection
- [x] **Test D: Form Submission** - Test successful conversion with proper API calls and cleanup
- [x] **Test E: Error Handling** - Handling of API failures and missing data

**Reproduction Steps:**
1. Create a queue question in any course
2. Navigate to async centre with URL params: `?convertQueueQ=true&qid={queueQuestionId}`
3. Verify modal opens with AI-generated abstract and queue question text
4. Select question types and submit
5. Confirm anytime question is created and queue question is deleted

# Checklist:


packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -429,6 +440,20 @@ export default function AsyncCentrePage(
           }}
           chatbotQ={{ messages: messages }}
         />
+        <ConvertQueueQToAnytimeQModal
+          courseId={courseId}
+          queueId={questionId}
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:445]: um

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx
```diff
@@ -0,0 +1,407 @@
...
+    const fallbackTitle = words.join(' ')
+    return fallbackTitle.length > 100
+      ? fallbackTitle.substring(0, 97) + '...'
+      : fallbackTitle
+  }
+
+  useEffect(() => {
+    if (open) {
+      const fetchQuestion = async () => {
+        if (queueId) {
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx:128]: Add an `initialLoading` state so that the person can't mess around with the form or submit it while it's loading the queue question

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx
```diff
@@ -0,0 +1,407 @@
...
+              // Generating abstract using chatbot
+              let generatedAbstract = ''
+              try {
+                const response = await API.chatbot.studentsOrStaff.queryChatbot(
+                  courseId,
+                  {
+                    query: myQuestion.text,
+                    type: 'abstract',
+                  },
+                )
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx:144]: Personally, I wouldn't `await` this, or at least load in the other stuff first and then do the AI. Mostly since it can take a second or two to get a response from the AI (or longer), which doesn't lead to a very good UX while you're waiting around for the form to get filled in.

I would instead first load in the question text and abstract into the form, and *then* load in the abstract (and probably have a separate loading state for said abstract field so they can't edit it while it's being generated. Could even say "Generating abstract..." while doing so).

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx
```diff
@@ -0,0 +1,407 @@
...
+          aiAnswerText: aiAnswer,
+          answerText: aiAnswer,
+          questionAbstract: values.QuestionAbstract,
+          status: asyncQuestionStatus.AIAnsweredNeedsAttention,
+        },
+        courseId,
+      )
+
+      const questions = await API.questions.index(queueId)
+      const myQuestion = questions.yourQuestions?.[0]
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx:260]: you do this in two areas in this modal. 

Now, one comment I would have is that it's inefficient to retrieve something from the server when it could probably just be saved in local storage or something. However, in this case it's probably a good idea to grab it from the server since it's less error-prone compared to a purely front-end solution (e.g. since localStorage might not always be cleared from previous conversions).

However, the API.questions.index(questionId) is a pretty heavy endpoint. Though, I realise that we also don't have an endpoint to just get the details of a question with a given questionId. Maybe you can do that? Create a simple GET endpoint with a questionId under question.controller.ts? 
Maybe call it `getMyQuestionDetails` and have it protected so that only the question creator can get details for their question. Just make sure to log the output of the endpoint  while you're testing to make sure it's not like returning anything it shouldn't (e.g. the user's password hash). You will need to create some tests for it, but luckily making tests for CRUD endpoints is easy to do.

Then, once you do so, you can store this question details inside a state variable, that way you don't need to retrieve it multiple times. 

Also, I would *really* make sure to handle the error case where no question was retrieved, probably just by doing `message.error("Your queue question could not be retrieved: " + getErrorMessage(e)).` (getErrorMessage is a utility function in our codebase).

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -77,14 +78,24 @@ export default function AsyncCentrePage(
 
   const [convertChatbotQModalOpen, setConvertChatbotQModalOpen] =
     useState(false)
+  const [convertQueueQModalOpen, setConvertQueueQModalOpen] = useState(false)
   const convertChatbotQSearchParam = searchParams.get('convertChatbotQ')
+  const convertQueueQSearchParam = searchParams.get('convertQueueQ')
+  const questionId = Number(searchParams.get('qid'))
```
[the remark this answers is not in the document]
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:84]: it's also not a questionId it's a queueId

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
Comment (bhunt02) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx:445]: he uses router.push i believe to fill in the parameters. this is probably a better solution, though. but i think it's so in case it fails it doesn't reload and lose context.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx:260]: or maybe a `getMyQuestions` endpoint would actually be more useful. 

Actually on second thought, scratch that idea of an endpoint. It's probably not worth it. The idea to load the question details into a state variable is probably a good idea though

Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertQueueQToAnytimeQModal.tsx:260]: ALSO, when you're setting said state variable, have a check for the length of yourQuestions. If the length is >1, choose the question with the matching queueId and isTaskQuestion == false.

packages/frontend/dev.env
```diff
@@ -1,21 +0,0 @@
-# NOTE: all variables you want to be accessible on the client side must be prefixed with NEXT_PUBLIC_
```
AdamFipke, line 1: you should probably un-delete this (and postgres.env.example). Don't forget to review your own PR!

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/PromptStudentToLeaveQueueModal.tsx
```diff
@@ -34,16 +40,61 @@ const PromptStudentToLeaveQueueModal: React.FC<
...
+      setIsConvertLoading(false)
+      setIsStayLoading(false)
+      setIsLeaveLoading(false)
+    }
+  }, [])
+
+  // Reset convert loading state when component mounts (when user navigates back)
+  useEffect(() => {
+    setIsConvertLoading(false)
+  }, [])
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/PromptStudentToLeaveQueueModal.tsx:64]: erm, this doesn't do anything (or it shouldn't, anyway)

When the component mounts, the states get set to their default automatically (in this case false).

AdamFipke (changes requested):
Good start! Left some comments

AdamFipke:
I guess re-request a review from me when you're ready!

packages/frontend/dev.env.html
```diff
@@ -0,0 +1,2191 @@
...
+  <nav-list>
+    <ul data-target="nav-list.topLevelList" data-view-component="true" class="ActionListWrap">
+        
+          
+<li data-item-id="" data-targets="nav-list.items" data-view-component="true" class="ActionListItem">
+    
+    
+    <a data-hotkey="g d" data-analytics-event="{&quot;category&quot;:&quot;Global navigation&quot;,&quot;action&quot;:&quot;HOME&quot;,&quot;label&quot;:null}" id="item-659a5287-8d1b-46bf-8af4-44647c7c5005" href="/dashboard" data-view-component="true" class="ActionListContent ActionListContent--visual16">
+        <span class="ActionListItem-visual ActionListItem-visual--leading">
+          <svg aria-hidden="true" height="16" viewBox="0 0 16 16" version="1.1" width="16" data-view-component="true" class="octicon octicon-home">
```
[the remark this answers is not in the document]
Eneshjakhar: restoring the .env file accidentally gave it an .html extension. I’ve corrected this in the latest commit, so the file is now properly named as .env.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/PromptStudentToLeaveQueueModal.tsx
```diff
@@ -1,21 +1,64 @@
...
+          setCanConvert(false)
+          setConvertTooltip('Demos cannot be converted into anytime questions')
+        } else {
+          setCanConvert(true)
+          setConvertTooltip('')
+        }
+      } else if (myQuestions.length > 1) {
+        myQuestion = myQuestions.find(
+          (q) => q.queueId === qid && q.isTaskQuestion === false,
+        )
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/modals/PromptStudentToLeaveQueueModal.tsx:48]:
```suggestion
        myQuestion = myQuestions.find(
          (q) => q.isTaskQuestion === false,
        )
```
Questions don't have a .queueId attribute (it gives a linter error). Also that check is not necessary since the endpoint is only getting questions from this queue

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertChatbotQToAnytimeQModal.tsx
```diff
@@ -42,10 +43,56 @@ const ConvertChatbotQToAnytimeQModal: React.FC<
...
+      const generateAbstract = async () => {
+        setIsGeneratingAbstract(true)
+        try {
+          const response = await API.chatbot.studentsOrStaff.queryChatbot(
+            courseId,
+            {
+              query: questionText,
+              type: 'abstract',
+            },
+          )
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertChatbotQToAnytimeQModal.tsx:69]: Oh yeah I just realised that it may be wise to add a timeout with these of probably 5s. Since if for some reason if the chatbot server freezes up, the user will be stuck there unable to continue (until it eventually times out after a handful of minutes).

I went and quickly added this to chatbot-api.service (for only abstract generation) and pushed it

AdamFipke (approved):
I haven't manually tested the new changes but the code changes look pretty good. Just that one suggestion comment that you can click commit on, and then also fixing the merge conflicts

packages/server/src/queue/queue-clean/queue-clean.service.ts
```diff
@@ -167,7 +168,14 @@ export class QueueCleanService {
           ...Object.values(LimboQuestionStatus),
         ],
       })
-      .getRawMany<{ studentId: number; courseId: number }>();
+      .andWhere('QuestionModel.isTaskQuestion = :isTaskQuestion', {
+        isTaskQuestion: false,
+      })
```
AdamFipke, line 173:
```suggestion
```

Issue #336, opened 2025-07-29
Staff don't know if an Anytime Question already got help when checking Email


AdamFipke:
## Scenario

Picture this:
- Big course, like 5 TAs
- Student asks Anytime Question and requests more help, which emails all the TAs and prof
- TA 1 checks the email first, logs in and opens Anytime Question Hub and provides human response
- TAs 2-5 and the prof all check their email throughout the day, see the initial email that the anytime question needs help, goes through the hassle of logging in and everything, just to find it already got a response. 

Arguably, this *loses* more time and causes more overhead than if the student just sent an email to all the TAs + prof.

## Solutions

- I'm fairly certain this is unfeasible since I've never seen it before, but it would be cool if we could get the email service to ***reply*** to all of the initial "this question needs more help" emails. That way it helps keep their inbox organised and possibly with less notifications (depending on how the email's receiver service does this). This reply could even mention what the abstract, the question, the question types, the ai response, and the human response all are, and possibly even who gave the response (((maybe, idk if i would like that as a TA))). This reply could also be a single reply email if #335 is figured out
- Send another batch of emails just to notify the staff that the anytime question has received an AI answer. I don't like this 'cause it just clutters up the inbox even more.

bhunt02:
If we replied to our own email, we'd have to CC all the original recipients. We'd also have to keep track of the original email... I don't know if we're able to do that with our libraries?

PR #344, opened 2025-08-06
Async Question Email Improvements

### LINKED ISSUE #335: Send staff notification emails with multiple recipients rather than individual
This might be a bad idea for reasons I have not yet thought of, but basically I realised that there's nothing really 'personal' about some of the emails we send and we could send them to multiple recipients in the same email.

For example, the "Anytime question needs attention" email that gets to all staff doesn't really have anything specific to the user receiving the email.

And like our "Unsubscribe" button just takes the user to the /profile page, so that would still work.

## Advantages

- Less emails means slightly better performance and less chance of being marked spam.
- Staff can actually *reply* to said emails in case they want to discuss how they want to proceed in answering the question (could even mention it in the email body)

## Disadvantages(?)

- I guess the only thing that the other staff members would have access to is the emails of the other staff members (which could technically be a personal email address). But meh they likely already know each others emails

And I guess if there's another reason this is a really bad idea (or if it's just unfeasible), this issue can just be closed.

bhunt02:
# Description

* Group 'Needs Attention' recipients to send only one email, also to make the next part easier:
* Reply-to previous email when question is answered

Closes #335, #336

## Type of change

# How Has This Been Tested?

Updated existing unit test
Added new unit test

# Checklist:


bhunt02:
@AdamFipke FYI i do not know how to test whether this works locally, with actual email sending, so it's kind of flying blind

packages/server/src/mail/mail.service.ts
```diff
@@ -83,14 +89,45 @@ export class MailService {
...
+        emailPost.subject,
+        fullContent,
+      );
       return;
     }
-    await this.mailerService.sendMail({
-      to: emailPost.receiver,
+
+    const result: SentEmailResponse = await this.mailerService.sendMail({
+      to: emailPost.receiverOrReceivers,
```
AdamFipke, line 101: i really hope it's as easy as that

packages/server/src/mail/sent-email.entity.ts
```diff
@@ -0,0 +1,26 @@
+import { BaseEntity, Column, Entity, PrimaryColumn } from 'typeorm';
+import { EmailMetadata, MailServiceType } from '@koh/common';
+
+@Entity('sent_email_model')
```
AdamFipke, line 4:
```suggestion
/* Need to save email ids in order to be able to reply to emails sent earlier (e.g. to reply that an anytime question already received help) */
@Entity('sent_email_model')
```

packages/server/src/asyncQuestion/asyncQuestion.service.ts
```diff
@@ -202,6 +206,38 @@ export class AsyncQuestionService {
...
+      .select()
+      .where('se.serviceType = :serviceType', {
+        serviceType: MailServiceType.ASYNC_QUESTION_FLAGGED,
+      })
+      .andWhere('se.metadata @> :metadata', {
+        metadata: { asyncQuestionId: question.id },
+      })
+      .getMany();
+
+    if (needsAttentionEmails.length > 0) {
```
AdamFipke, line 220: hmm, just a small edge case. I wonder if we can *not* send these emails if the course only has 1 staff member in it (since, obviously they were the one that gave the answer).

Other than that though, I unfortunately can't think of any other ways that we can avoid sending these emails to staff that were the ones that gave the answer (due to it being a reply), but ah well.

AdamFipke:
okay I tried doing a little manual testing, and it seems like emails are not sending? No messages appearing in console though, so it could just be my environment. Ran out of time today, might do more testing later

I will note that i think you forgot to put the new entity in ormconfig.ts

packages/server/src/mail/mail.service.ts
bhunt02: documentation says it can accept either a single address/string, or array of addresses/strings :- )
the stuff for replying is like, replyTo: {generated email identifier from mail service} and references: {same ID}

bhunt02: also, the subject line has to be either the exact same or have "Re:" attached to the start if it's different for most inbox providers to recognize them as the same thread, even if its replying to the same ID. kind of silly but i guess that's just how email works

AdamFipke: huh interesting!

AdamFipke (approved):
yippie!

PR #345, opened 2025-08-06
Improve syncDocuments functionality to detect changed LMS items, add …

### LINKED ISSUE #312: [BUG] "Force Synchronization" feature can silently fail and still display a 'sync success' message
**Describe the bug**

This is related to the last point of this comment https://github.com/ubco-db/helpme/pull/304#issuecomment-3002161158

And a fix could be one described in this comment https://github.com/ubco-db/helpme/pull/304#issuecomment-3002217932

> a change to accomodate this without going overboard with specific messages would be changing the return behaviour to return the number of failed docs (or something similar) and using that with a message.warn/success/error ANTD call to present that to the user, e.g., "Some documents failed to sync"

FabianLCH:
# Description

- Removed code in the LMS integration service's `syncDocuments` that resulted in inconsistent syncing behaviour when there are persisted items to remove (e.g. because they were removed from the LMS course)
- Modified `syncDocuments` to augment objects returned by the LMS's API (e.g. LMSPage, LMSAssignment, etc.) rather than fully replace them with persisted items to properly update persisted data when LMS data changes
- Added counters for synced/removed documents and modified return types to send synced/removed document counts to frontend
- Modified frontend messaging on the course settings page's "LMS Integrations" tab to provide more detailed but still brief feedback to the user after forcing synchronization

Closes #312

## Type of change

# How Has This Been Tested?

Tested manually.

# Checklist:

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -1354,6 +1354,7 @@ export type LMSAssignment = {
   due?: Date
   modified?: Date
   uploaded?: Date
+  chatbotDocumentId?: string
 }
 
 export type LMSAnnouncement = {
@@ -1364,6 +1365,7 @@ export type LMSAnnouncement = {
   syncEnabled?: boolean
   modified?: Date
   uploaded?: Date
+  chatbotDocumentId?: string
 }
 
 export type LMSPage = {
@@ -1375,6 +1377,7 @@ export type LMSPage = {
   syncEnabled?: boolean
   modified?: Date
   uploaded?: Date
+  chatbotDocumentId?: string
 }
 
 export type LMSFile = {
@@ -1386,6 +1389,7 @@ export type LMSFile = {
   syncEnabled?: boolean
   modified?: Date
   uploaded?: Date
+  chatbotDocumentId?: string
 }
 
 export enum SupportedLMSFileTypes {
@@ -1414,6 +1418,12 @@ export type LMSFileUploadResponse = {
   success: boolean
   documentId?: string
   reason?: LMSErrorType
+}
+
+export type LMSSyncDocumentsResult = {
+  itemsSynced: number
+  itemsRemoved: number
+  errors: number
 }
 
 export enum LMSApiResponseStatus {
--- a/packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
@@ -252,7 +252,14 @@ export default function CourseLMSIntegrationPage(props: {
             `Unknown error occurred, could not force synchronization with the LMS.`,
           )
         } else {
-          message.success(result)
+          if(result.errors > 0) {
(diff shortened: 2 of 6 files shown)


FabianLCH:
@bhunt02 There's a couple more changes than I initially planned, but I figured I should include them in this PR as they all relate to the same feature. Still, it's mostly fairly minor bug fixes/improvements (no major functionality changes). Please feel free to review when you get a chance!

packages/common/index.ts
```diff
@@ -1386,6 +1389,7 @@ export type LMSFile = {
   syncEnabled?: boolean
   modified?: Date
   uploaded?: Date
+  chatbotDocumentId?: string
```
bhunt02, line 1392: why? this was omitted from the frontend types so that we'd expose as little as possible, we don't reference it either; i'm assuming you're doing this to get around some TypeError stuff due to the fact the canvas-retrieved versions have the chatbotDocumentId property but you can get around that with casting inline (it's ugly, but works)

i don't want to be sending chatbotDocumentId down to the frontend, either

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
```diff
@@ -351,7 +358,12 @@ export default function CourseLMSIntegrationPage(props: {
   )
 
   const failedToSync = useMemo(
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:360]: this is largely irrelevant; there's a reason the "can't sync" designator exists, and it's precisely for this. technically, a "can't sync" still failed to sync, but we show the label for "can't sync" rather than "failed" for the reason of giving the full picture.

bhunt02 (changes requested):
few little things i noticed, overall looks good

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx
Comment (FabianLCH) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:360]: My assumption here was that the three labels are mutually exclusive:
- "Synced" > item that can be synced and was successfully synced
- "Can't Sync" > unsupported item that is not included in the sync operation
- "Not Synced" > item that can be synced but for whatever reason has not been synced

With the recent addition of LMS files, I noticed each file with an unsupported MIME types was counted as both "Can't Sync" and "Not Synced" in the labels under "Force Synchronization":

Force Sync Labels

But in the table it's only tagged with "Can't Sync":

Course Files Table Cant Sync Item

Thought this could be confusing to the user as the tagging difference makes it seem like there's an extra item the should sync but hasn't ("Not Synced") for each unsupported item. My idea with the extra condition in the `failedToSync` filter is to differentiate between "Not Sync" and "Can't Sync" to be consistent with the table below. What do you think?

Comment (FabianLCH) [reply packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/lms_integrations/page.tsx:360]: Got it, I'll go ahead and revert that change then!

PR #351, opened 2025-08-13
LTI Integration

### LINKED ISSUE #308: Canvas App (HelpMe inside Canvas)
So rather than just pulling canvas stuff onto HelpMe, the idea being to put HelpMe things on Canvas (e.g. a Chatbot app)

Apparently it might be more technical, and an iframe apparently might not be good enough.

But we are having a meeting soon about this, and we can use this issue to track what is required and progress regarding it

bhunt02:
# Description

~~This will get a better description later~~
Oh god... do I really have to???

Fine...

### Major Changes/Additions
* **Learning-Teaching Interoperability**
  * Added LTI implementation to allow LTI-compliant platforms to implement HelpMe as a tool
  * As with LMS integrations, only Canvas has an implementation. If we get more clients, it should be easy enough to extend over to other LMS's so long as they have:
    * An ID corresponding to the course identifier (like Canvas does) (used to identify course via LMS course integration via user account)
    * Supports sharing user email claims with us (used to identify user account if exists)
  * Developed a customized, limited frontend for use with the LTI tools (embedded in an iframe within the platform)
    * Currently, courses dashboard and course page + course integration management page for course staff
    * LTI context and customized header added
      * LTI tool will resize itself to the window's full size when launched
        * **BECAUSE CANVAS DOESN'T ALLOW SPECIFYING FULL HEIGHT BY DEFAULT IN LAUNCH SPECIFICATION :))))**
  * Developed a customized authentication pipeline - more on that later
    * This is related to why we are presenting a limited frontend to LTI tools: we **have to issue SameSite=None authentication tokens when HelpMe is embedded in an iframe**
      * These tokens can be a bit dangerous for us
      * More on the authentication pipeline later to explain how this danger is circumvented
  * Comes with addition of an admin dashboard - this is where the 'userRole' column of the User table can become relevant.
    * Only accessible to `userRole = admin`.
    * Currently only has the list of registered LTI platforms
      * Platforms can be added manually (ew don't do that, we will tell site admins to use `https://<domain>/api/v1/lti/register` for dynamic registration)
      * Platforms can be deleted
      * Platforms can be set to be active/inactive (accept and process requests for LTI launches or not)
      * Platforms (on our end) can be updated with new credentials if they change
      * Platforms that implement the scopes for updating and reading the OpenID registration will be able to be updated and also automatically dropped if the configuration doesn't exist on their end anymore
        * **UNFORTUNATELY CANVAS DOESN'T ALLOW THESE SCOPES AND IT'S ANNOYING BECAUSE I SPENT A BIT IMPLEMENTING THIS ONLY TO FIND CANVAS DOESN'T ALLOW IT (WHY??? WE MIGHT HAVE HANGING PLATFORMS ON OUR END???)**
  * Uses `lti-typescript` (more on that later) to perform the LTI launch authorization/pipeline

* **Authentication Pipeline Expansion/Refactor**
    * Secondary Authentication Pipeline, specific for LTI
      * Modifications to all frontend pages under (auth), with behaviours specific to if their pathname starts with '/lti'
        * Baby-gate LTI launches to the LTI pages as much as possible (I think I've covered all my bases)
        * To maximize re-use of code, the structure is duplicated, but the same components are imported and exported from the routes under `/lti/(auth)/...`
      * When the site is embedded in an iframe, we need to:
        * Issue access tokens with `SameSite=None`;
        * Secondary redirects to other platforms (e.g., Google SSO, UBC Shibboleth) need to occur in a new window/tab;
          * These automatically close after the authorization is complete (but don't if it fails so the user can see the error message)
          * The tokens are attached to the domain on the user's browser so it's fine this way, the iframe presents an interface to reload the embedded page, the middleware can then access the new authentication state and behave accordingly
    * **Solution:** extract and modularize authentication methods used in the backend application. They have also been updated.
      * `auth.service.ts` now contains methods which process the response, but take in several optional parameters.
      * `login.service.ts` was created to facilitate the login methods for both (at least, the two **current** versions, default and lti) with variable behaviours
        * `entry` is a function which maps to the old login entry behaviour that was repeated in several instances of the code (e.g., SSO auth & legacy auth had shared behaviours)
        * `handleCookies` performs the management of redirects with email-verification specific behaviours added, and LTI-specific ones too
      * `auth.controller.ts` heavily modified with a lot of behaviour extracted to `auth.service.ts`
      * `lti-auth.controller.ts` created to call `auth.service.ts` methods with LTI-specific parameters (e.g., cookie options, redirects)
    * **Scoped Authorization Tokens**
      * LTI implementation introduces this, with `lti_auth_token` specifically
        * Secondary type of `auth_token` so we can differentiate (at least slightly)
        * Implements new 'restrictPaths' property of our tokens that checks whether the API route or the frontend path being accessed is allowed for the token
          * This prevents LTI launches from accessing routes we don't want them to
        * We could apply this in the future, setting non-verified accounts to be restricted to any page outside /verify, and upon verification, allowing them through
        * We could also apply this in the future to route protection
    * SSO Authentication Changes
      * No longer uses a cookie to store the organization ID. 
      * Uses the supported `state` query parameter to pass a unique, 32-bit identifier that is later used to authenticate the SSO login request (at the callback) and retrieve the stored organization ID.
      * `AuthStateModel` is used for this. It stores the `state` sequence as a primary key, expires in 60 seconds, and identifies the organization being logged in with.
      * This is a lot simpler than using cookies, and it's supported by most OAuth providers.

* **LMS Integrations**
  * **LMS Access Token Generation**
    * Using OAuth syntax/pipelines, LMS course integrations can now have access tokens generated on demand. This is the expected implementation from Instructure and our own UBC CTL.
    * **Added an organization setting which disables API key usage** (so that developers can still use **API keys** to authorize with Canvas)
        * Functionally on production we'll disable this option
    * Related types & token management added to user pages and within the LMS course integration upsert menu
    * Access token generation requires the LMS organization integration to have a client ID and client secret set (supplied by the platform upon creation of the developer key)
        * This might be Canvas-specific but I don't know, I've not worked with Moodle, or Blackboard, or any of the rest
        * We can address it later if it is Canvas-specific but I feel as though it's not since this is deeply intertwined with LTI protocols
    * LMS user access tokens are stored as `LMSAccessTokenModel` entities
        * LMSCourseIntegration now have a 0..1 relationship with these entities
        * Users have 0..* of these, but they are softly identified by the user-organizationIntegration relationship
          * A.K.A., user ID, organization ID, organization API platform
        * Internally identified by a unique serial ID so the FK with LMSCourseIntegration isn't insane
    * During the authorization pipeline (which redirects to the LMS to perform authorization to obtain the token):
      * An `LMSAuthStateModel` entity is created (it has some useful properties such as storing a redirect right back to the page with certain query params)
        * It can also expire so users have to authorize within 60 seconds
        * It makes sure we can authenticate that the request to generate an access token actually came from the platform (as long as a valid, matching state entity exists)
    * Users can manage their access tokens in the profile page under a submenu
      * It's implemented in such a way that if we have more 3rd party token types other than LMS in the future, those could be added easily
  * **Updates to Organization Integration Creation/Update**
      * Added three properties:
        * secure (default: true): boolean property indicating whether the platform is HTTP or HTTPS
          * Really only useful if you're hosting Canvas locally and need to contact an HTTP route, but I digress
          * Hence why it defaults to true
        * clientID (optional): corresponds to the tool client ID, from the LMS platform, for the developer key used to create access tokens
        * clientSecret (optional): corresponds to the tool client secret, blah blah blah
      * Extracted upsert modal to its own component
      * Organization integration management also comes with the ability to invalidate user access tokens
        * I don't know I just thought it would be a good idea since otherwise, the user's access tokens are managed by the user solely
      * Backend: slight refactor of organization upsert
  * **Updates to Course Integration Creation/Update**
    * When there's an access token available/selected, it uses this token to retrieve the list of user courses from Canvas
      * No more explaining to users what their Canvas course ID is
    * Can use either API key or access token (if API key is enabled in organization settings)
    * If it's not, can only use access token generation
    * When either is uploaded, the other is deleted from the database (so it's easier for us to decide what to use)
      * API key expiry is deleted alongside API key in the case when an access token replaces it
    * Various updates to the interface and properties that can be input/logic of when they're shown
    * Backend: overhaul of course integration upsert

### What's an LTI-Typescript???

A library called `ltijs` was offered to me as an option for configuring our integration. 
I started trying to use the library and then I realized it was pure Javascript.
This deeply upset me. I then embarked on a 2-week-long endeavor to fork and rebuild this library using TypeScript.
It uses TypeORM, like we do, for compatibility. Also because the original implementation used MongoDB and I wasn't about to add another DBMS and database server/container we have to use and secure.
Learned a lot about the LTI protocol and how it works. We now have a library that is strongly typed and I'll try to maintain.
You can find it [here](https://github.com/bhunt02/lti-typescript).

### What else has been added?
* Configuration
  * New Environment Variables
    * `server/.env`
      * `LTI_SECRET_KEY` - the secret key used by LTI-Typescript for en/decrypting data in encrypted tables
    * `server/postgres.env`
      * `POSTGRES_LTI_DB` - the name of the database used for lti (it's `lti`)
  * 2 new databases: `lti`, `lti_test` (purposes are exactly what you think)
    * Includes updates to the `test_createdb.sql` file that is used for creating the test DBs in CI
* Frontend
  * `api/server.ts` file added, provides server-side version (**NOT FOR MIDDLEWARE**) of the `api/index.ts` file
    * Since it's server-side, it needs to add the authorization cookie header explicitly
      * APIClient constructor customized for this reason
  * `/lti` path
    * All lti-specific interfaces are stored here
  * `/admin` path
    * Stub definition for main admin console
    * `/admin/lti` - LTI platform management
  * `constants.tsx` - File for defining constants. Currently exports and holds the HELPME_VERSION variable
    * Hard coded still but easier to locate than it was before
* Backend
  * Added `admin.controller.ts` (empty, for now) and `admin.service.ts` (same)
  * Added `lti.middleware.ts`, a middleware that is registered and used to configure the LTI-Typescript instance
    * e.g., dynamic registration settings, whitelisted routes
  * `dataEndec.ts`, which will decrypt and encrypt values for storage in an entity that has an encrypted column
    * Make sure the key is the same when you decrypt/encrypt the values
    * Currently only used for `LMSAccessTokenModel` as it encrypts the token details with the client secret.
  * Various decorators for use with the LTI backend routes (currently only one)
  * `AdminRoleGuard` is used to guard routes only for users with `userRole = admin`
  * `LTIGuard` checks that the `response.locals['token']` object is defined.
    * Mostly just a safeguard if the LTI middleware fails or is inactive, prevents accessing the LTI launch route without the IdToken.
  * Added CRON jobs which delete expired tokens for `LMSAuthStateModel`, `AuthStateModel`.
  * `AbstractLMSAdapter` / LMS Adapters in general
    * Static methods
      * `createState` - creates an LMSAuthStateModel instance with a unique state sequence
      * `logoutAuth` - initializes a logout (invalidation) operation of a user's access token
        * Determines which implemented adapter to use, then calls its implementation of (non-static) logoutAuth
      * `redirectAuth` - initializes an authorization redirect operation for access token generation.
        * Determines which implemented adapter to use, then calls its implementation of (non-static) redirectAuth
      * `postAuth` - accepts an authorization response for access token generation. 
        * Determines which implemented adapter to use, then calls its implementation of (non-static) postAuth
    * Non-static methods
      * `logoutAuth` - implemented in subclasses. Used for invalidating access tokens.
      * `checkAccessToken` - implemented method for checking whether an access token is valid. Refreshes the access token if not.
        * Throws an exception if the access token can't be refreshed.
        * Otherwise, saves the refreshed token and returns it.
        * If the token is not expired, returns it outright.
      * `getAuthorization` - implemented method for getting the authorization header.
        * If API key is used for course integration, simply creates a `Bearer <token>` header.
        * Otherwise, uses checkAccessToken to get the (possibly refreshed) access token, and creates a `<token-type> <token>` header.
      * `getUserCourses` - implemented in subclasses, gets the courses to which a user belongs, pending they have an access token.
        * This only works with an access token as when a token is created, **it provides the user's Canvas ID.**
          * We do not get this ID from the API key. We could, but doesn't seem worth it if we're sunsetting that.
  * `lmsIntegration.controller.ts`
    * Added routes for:
      * `GET org/:oid/token` -> (ORG ADMIN) retrieves the tokens for the organization and (optional) platform
      * `DELETE org/:oid/token/:tokenId` -> (ORG ADMIN) invalidates user access token if found
      * `GET oauth2/token` -> retrieves the user's options for tokens (used in course integration upsert)
      * `DELETE oauth2/token/:tokenId` -> invalidates user access token if found and belonging to user
      * `GET oauth2/authorize` -> starts the pipeline for generating an access token
      * `GET oauth2/response` -> performs the response actions for generating an access token after redirected back
      * `GET course/list/:tokenId` -> gets the platform course list for the user given their token
  * `lmsIntegration.service.ts`
    * Added `createAccessToken`, `destroyAccessToken` methods
      * `createAccessToken` - takes the data from the oauth2 response, encrypts it into an `LMSAccessTokenModel`
      * `deleteAccessToken` - performs the logout auth and deletes the token if successful
  * `login.service.ts`
    * `initLoginEnter` - used in login controller and lti auth controller to try to call LoginService.enter, fails if token is invalid
    * `enter` - performs the old entry method with modifications
    * `handleCookies` - performs the redirect/cookie handling that the old entry method (and duplicates of it) would do
    * `handleLTICourseInviteCookie` - used in specific circumstances to check the LTI course invite without calling for a redirect
    * `generateAuthToken` - generates an authentication token for HelpMe with given parameters
  * `lti-auth.controller.ts` - implements several methods from `auth.controller.ts`, customized for LTI (nothing unique besides parameter differences)
  * `lti.controller.ts` - entrypoint for LTI functionality.
    * `ALL /` - entrypoint for LTI. returns a redirect to the `/lti` frontend route, or `/lti/<courseId>` frontend route if course matching API course is found. 
      * Authorizes the user with `LoginService.enter` method if the user is found.
      * Redirects to `/lti/login` frontend if user is not found.
      * Performs other operations, such as generating (if the course is found but user is not) `LTICourseInviteModel`s and cookies to add the user to a course after they've registered, if not found
      * Automatically adds the user to the course if the course exists and the user exists
      * Sets query parameters that are eventually put into session storage and used for the LTI launch session
        * Particularly, the platform and the api course ID which are used in integration management
    * Includes `AdminGuard`'d endpoints for platform management.
    * Has exported function in file for mapping LTI-Typescript platform to representation HelpMe uses of it
  * `lti.middleware.ts` - defines the configuration for the LTI-Typescript provider-as-middleware
    * Launches the configuration as a middleware, used to perform the LTI pipeline except for whitelisted routes
    * Also automatically drops platforms which no longer store a configuration for us, so long as they have the scopes necessary enabled to get that information
  * `lti.service.ts`
    * Provides access to the active LTI provider as an instance property
    * `createCourseInvite` - creates an LTI course invite
    * `checkCourseInvite` - checks an LTI course invite for validity, if valid, adds user to course
    * `findMatchingUserAndCourse` - uses the LTI launch parameters to find a) a matching user, and b) a matching course, if possible.
      * If neither are found, behaviour in the controller defines how the LTI launch acts
    * `extractCourseId` - uses the LTI launch parameters and finds the course ID from the platform
      * This course ID is then checked if it matches with any existing course LMS integrations. If it does, that course is used for the launch.

### What else has changed?
* Common
  * 'userRole' attribute in `User` class in `common/index.ts` now maps to the enum.
  * Renamed `UBCOloginParam` to generic `LoginParam` (just for convenience, it's not even UBCO-specific - in fact, it's less UBCO-specific than the shibboleth login!!!)
* Frontend
  * Dashboard layout now shows verification page if redirect does not properly occur (bandaid)
  * Adjusted `common/index.ts` in general for placements, added class-transformer decorators to the LMS types/classes
  * Migrated/removed many types that are declared within `typings/*.ts` to `common/index.ts` so the backend and frontend declarations are fully aligned
  * `registration/page.tsx` extrapolated to optionally take in an `oid` URL parameter, rather than relying on the value in localStorage
    * Also, no longer deletes this value from localStorage upon successful registration
  * `coursesSection.tsx` modified to support the needed behaviour within the LTI pages
  * Updated some control logic related to LMS integrations
    * Also updated some descriptions
    * Also made it so if a resource is disabled the related fetch for the resource elements is not called (it's wasteful otherwise/quizzes throw an ugly error if disabled and a fetch is attempted)
  * Updated various erroneous control logic related to roles (e.g., EmailNotifications.tsx assumes 'userRole' mapped to OrganizationRole type, but it never would)
  * `api/cookieApi.ts` renamed to `api/cookie-utils.ts` and modified for LTI-specific behaviours
  * `api/index.ts`
    * Added the various `auth` routes here as a result of removing `api/authApi.ts`
    * `profile.index` -> `profile.getUser`, with original implementation that was in `userApi.ts`
      * This is for compatibility reasons
    * `profile.index` -> `profile.fullResponse`, `profile.getUser` but returns full response object
    * `mail` added, `mail.resendVerificationCode` added (result of removing `api/mailAPI.ts`)
    * Useful function for `fetchUserDetails` (reused to supply profile to UserInfoContext) extracted/added
  * `middleware.ts`
    * Updated/added public pages
    * Restricted routes from authentication token are processed
    * LTI-specific behaviours added
      * HelpMe tab that's launched from the 'Open HelpMe' button in the LTI interface will delete the lti_auth_token
      * Default redirect page is different for LTI, e.g., '/lti' vs. '/courses'
      * All pages redirected to in LTI are different, usually prepending '/lti' to the redirect URL
* Backend
  * `auth.controller.ts` methods extracted to `auth.service.ts` for use with other authentication controllers, like `lti-auth.controller.ts`
  * `auth.service.ts` now contains additional general-purpose authentication methods
  * `bootstrap.ts` modified for LTI usage
    * Helmet and cookieParser default settings (used elsewhere in the app) were incompatible with the LTI implementation
    * Standard Helmet and cookieParser are used app-wide, except for on the LTI routes
    * LTI routes have:
      * Helmet: `{ frameguard: false, contentSecurityPolicy: false }` - prevents issues with displaying app in iframe
      * CookieParser: called with `LTI_SECRET_KEY` as the key used to decrypt the cookies (cookies written along the LTI routes will also be signed with this key)
    * Also had to make `addGlobalsToApp` function take test as an argument as the Regex used for the custom middlewares would be far too complicated (and unrealistic to use) for tests
      * Tests do not have the `/api/v1/` prefix, complicates things, can't make that optional
  * `JWTAuthGuard` updated with custom `handleRequest`
    * Checks for restrictPaths in the 'user' (decoded and verified JWT)
      * If there's none, returns the user (they're not restricted)
      * If there is, it:
        * If restrictPaths is an array, it checks whether the path being requested is contained within it, either through Regex matching or string equality
        * If it's a string, it checks whether the path being requested is allowed, either through Regex matching or string equality
  * `lmsIntegration.controller.ts`
    * Moved data validation from the controller method into the service method for `POST org/:oid/upsert` -> `upsertOrganizationLMSIntegration`
    * Overhauled `POST course/:courseId/upsert`
    * Updated `POST :courseId/test`
  * `lmsIntegration.service.ts`
    * `upsertOrganizationLMSIntegration`
      * Moved data validation originally in controller to here
      * Operations now drop undefined columns
      * Invalidates user access tokens if the client ID or client secret changes
        * Removes them all after even if some operations were unsuccessful (we won't need them, but we're doing a courtesy to the platform by invalidating them outright)
    * `updateCourseLMSIntegration`
      * Overhauled, overwrites properties only if they're defined (safeguard)
    * `testConnection`
      * Now works with access tokens
    * `getAPICourses`
      * Satellite method to call an adapter for `Adapter.getUserCourses`
  * `JwtStrategy`
    * Now either gets the `auth_token` (default) or if it is null, the `lti_auth_token`, rather than just `auth_token` or null
  * `login.controller.ts`
    * Extracted methods to `login.service.ts`
    * Changed 'legacy login attempt with SSO account' HTTP status to `I_AM_A_TEAPOT`
      * This brings me great joy
    * `logout` - now tries to clear both `auth_token` and `lti_auth_token`
  * `mail.controller.ts`
    * Token behaviours updated. `resendRegistrationToken` adjusted.
  * `profile.controller.ts`
    * `get` profile returns restrictPaths as well.
  * `user-token.entity.ts`
    * `created_at` -> createdAt, CreateDateColumn with `timestamp with timezone type`, Date type in TS
    * `expires_at` -> expiresIn, `bigint` column that aligned with epoch millis to `bigint` column that represents the number of seconds until the token expires.
    
### What's been removed?
* Environment Variables
  * Removed `GOOGLE_REDIRECT_URI` - we just compute this now since we have multiple redirect URIs (default, LTI) and other SSO providers will follow similar URIs
* Frontend
  * `api/authApi.ts` file was removed, methods extracted to central `index.ts` file
  * `api/mailAPI.ts` file removed, ...
  * `api/userAPI.ts` file removed, ...
  * Any usages of functions from these files have been corrected, the server-side version of the frontend API is used in **server components** (NOT THE MIDDLEWARE)
  * `api/organizationAPI.ts` almost entirely scrapped with methods extracted
    * Only `getOrganization` remains... might remove this in this PR as well
* Backend
  * Deprecated proxy mappings and remnants of the routes for the static admin pages
    * Deprecated `admin.command.ts` command file
    * Deprecated `admin-user.entity.ts` file

Closes #308, #389 

## Type of change

  - Kind of already did it with environment variables
**server/.env**
```dotenv
LTI_SECRET_KEY=<it's a secret to everyone>
```
**server/postgres.env**
```dotenv
POSTGRES_LTI_DB=lti
```
**server/.env**
```dotenv
LTI_SECRET_KEY=<it's a secret to everyone>
```
**server/postgres.env**
```dotenv
POSTGRES_LTI_DB=lti
```

# Production Environment Changes

* **Shibboleth authentication URL needs to be changed**
* **Admin proxy routes need to be removed**

# How Has This Been Tested?

I wrote a lot of tests. A LOT.

# Checklist:

### DIFF
--- a/packages/frontend/app/lti/layout.tsx
+++ b/packages/frontend/app/lti/layout.tsx
@@ -10,7 +10,8 @@ export function CookieWrapper({ children }: { children: React.ReactNode }) {
 
   useEffect(() => {
     const checkThirdPartyCookieStatus = async () => {
-      if (!(await document.hasStorageAccess())) {
+      const hasStorageAccess = document.hasStorageAccess ?? (async () => false) // assume false if not implemented
+      if (!(await hasStorageAccess())) {
         setHasCookieAccess(false)
       }
     }


packages/frontend/app/lti/(embed)/(pages)/layout.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  if (frameRatio == undefined) {
+    return children
+  }
+
+  const factor = Math.floor(frameRatio * 100)
+  const offset = Math.floor((100 - factor) / 2)
+  const transform = `scale(${factor}%) translate(0,-${offset}%)`
+
+  console.log(factor)
```
bhunt02, line 108:
```suggestion
```

packages/frontend/app/(auth)/login/page.tsx
```diff
@@ -1,36 +1,53 @@
...
+  const [ltiSSOInitiated, setLtiSSOInitiated] = useState(false)
+
   const recaptchaRef = React.createRef<ReCAPTCHA>()
   const router = useRouter()
   const searchParams = useSearchParams()
+  const pathName = usePathname()
+
+  const isLti = useMemo(() => {
+    return pathName.startsWith('/lti')
+  }, [pathName])
```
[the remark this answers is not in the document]
AdamFipke: ~~I guess the same question goes for all of the other modified `(auth)` pages~~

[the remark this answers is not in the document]
AdamFipke: Nvm I'm silly, I see now that all the /lti pages just import the pages from here. That makes sense

packages/frontend/app/(auth)/register/page.tsx
```diff
@@ -122,214 +161,233 @@ export default function RegisterPage(): ReactElement {
...
-      fetchOrganizations()
-    }
   }, [])
 
-  return (
-    <div>
-      {domLoaded && (
+  if (!organizationId) {
+    return (
+      <Suspense fallback={<CenteredSpinner tip={'Loading...'} />}>
```
AdamFipke: thank you for fixing this lol. Using a proper suspense instead of having a `domLoaded` state is much better (and I guess you could probably remove the `domLoaded` state altogether)

packages/frontend/app/(dashboard)/components/FooterBar.tsx
```diff
@@ -48,7 +49,7 @@ const FooterBar: React.FC = () => {
             IReadTheChangelog()
           }}
         >
-          Version 1.6.2{' '}
+          Version {HELPME_VERSION}{' '}
```
AdamFipke: personally I don't see a need for this change since I don't see us ever placing the helpme version in any other component (and if we do ever do that, we can refactor it then). Plus now we have one new file with a single variable in it, which could probably be placed somewhere else (like generalUtils).

packages/frontend/app/(dashboard)/organization/lms_integrations/components/UpsertOrgIntegrationModal.tsx
```diff
@@ -0,0 +1,167 @@
...
+                rootUrl: focusIntegration.rootUrl,
+                clientId: focusIntegration.clientId,
+                secure: focusIntegration.secure,
+              }
+            : {
+                secure: true,
+              }),
+        }}
+        onValuesChange={(changedValues) => {
+          console.log(changedValues)
```
AdamFipke:
```suggestion
```

packages/frontend/app/(dashboard)/organization/lms_integrations/components/UpsertOrgIntegrationModal.tsx
```diff
@@ -0,0 +1,167 @@
...
+          <Form.Item
+            name={'rootUrl'}
+            label={'LMS Base URL'}
+            tooltip={'The base URL of the LMS API to connect to'}
+            rules={[
+              {
+                required: true,
+                message: 'Enter a base URL for the integration',
+              },
+            ]}
```
AdamFipke: could probably stick in a rule in here to warn the user if they put in a link that starts with `https://` or `http://`

packages/frontend/middleware.ts
```diff
@@ -23,186 +33,316 @@ const isPublicPage = (url: string) => {
...
+}
+
 const isEmailVerified = (userData: User): boolean => {
   return userData.emailVerified
 }
 
+const axiosInstance = Axios.create({
+  baseURL: process.env.NEXT_PUBLIC_API_BASE_URL,
+})
+async function fetchUser(
```
AdamFipke: this seems to be some duplicate code from /app/api/index.ts. I would either:

- refactor the /api/index.ts to only have the general `profile.getUser` method (that does exactly the same as the `profile.index()` method worked before with none of the weird logic) and then give this file a special fetch function
- or refactor the api/index.ts to have a method that works here

The former is probably easiest.

I should also note that the `provideBaseResponse` parameter (and all associated logic) from the old method was only used for this middleware.

Also, I'm not too keen on using axios here (especially if the middleware here is creating a new axios instance for every single request). It would probably be best to keep using `fetch` in an effort to make the middleware as lightweight as possible

packages/frontend/app/api/server.ts
```diff
@@ -0,0 +1,9 @@
+'use server'
+
+import { APIClient } from '@/app/api/index'
+import { fetchAuthToken } from '@/app/api/cookie-utils'
+
+export default async function getAPI() {
+  const token = await fetchAuthToken()
+  return new APIClient(process.env.NEXT_PUBLIC_API_BASE_URL, token)
```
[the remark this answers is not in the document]
AdamFipke: Also I just read that apparently you're not actually supposed to fetch your own APIs in server components https://nextjs-faq.com/fetch-api-in-rsc (creates an extra round trip for all calls) and instead you're supposed to like call the endpoint or database functions directly

packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -340,7 +343,12 @@ const NavBar = ({
               {checkCourseCreatePermissions(userInfo, organizationSettings) && (
                 <NavigationMenuItem>
                   <Link
-                    className="md:pl-8"
+                    className={cn(
+                      '!md:pl-8',
```
AdamFipke:
```suggestion
                      'md:pl-8',
```
The other links don't have the `!important`, so I'm assuming adding it was a mistake

packages/frontend/app/admin/layout.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { LogoutOutlined, ReloadOutlined } from '@ant-design/icons'
+import { AsyncToasterProvider } from '@/app/contexts/AsyncToasterContext'
+import { UserInfoProvider } from '@/app/contexts/userContext'
+import StandardPageContainer from '@/app/components/standardPageContainer'
+import Link from 'next/link'
+import HeaderBar from '@/app/components/HeaderBar'
+import FooterBar from '@/app/(dashboard)/components/FooterBar'
+import { User } from '@koh/common'
+
+export default function AdminLayout({
```
AdamFipke: I would maybe just add a comment in this file and in `(dashboard)/layout` mentioning that these two files are really similar, and that if you're changing something in one file you should change it in the other.

Or as another perhaps better option, perhaps this layout.tsx and `(dasboard)/layout.tsx` can be merged (and then add a little bit of additional logic to it checking if the page they're visiting is a `/admin` route). Then you can move the whole `/admin` under `/(dashboard)`

packages/frontend/app/admin/lti/components/UpsertLtiPlatformModal.tsx
```diff
@@ -0,0 +1,364 @@
...
+    onWatchTokenChange(watchedUpdateToken as any)
+  }, [watchedUpdateToken])
+
+  useEffect(() => {
+    onWatchMethodChange(watchedCreateMethod)
+  }, [watchedCreateMethod])
+
+  useEffect(() => {
+    onWatchMethodChange(watchedUpdateMethod as any)
+  }, [watchedUpdateMethod])
```
AdamFipke: According to react docs, it's probably not a good idea to do useEffects like this (since it can create code that's hard to follow or cause cascading useEffects creating many unnecessary re-renders).

I would've suggested the actual react solution here, but i'll admit i'm having a little trouble following what is happening. Why is there two `form` states? If it's to re-use a form for both updating and creation, then I believe all you really need is to adjust the form's `initialValues` parameter (CreateAsyncQuestionModal has an example of this), but there could also very well be something you're trying to pull off here that I'm unaware of

packages/frontend/middleware.ts
AdamFipke: Oh yeah another reason that fetch might be better for performance is that axios will always serialize the response. But meanwhile with fetch you can do the request, handle the status code responses, and then choose to serialize the response based on that. Granted, in the end that probably doesn't account for much since we already serialize all the responses when the status code is successful

packages/frontend/app/contexts/LtiContext.tsx
```diff
@@ -0,0 +1,346 @@
...
+const LtiMessages = [
+  'lti.capabilities',
+  'lti.getPageContent',
+  'lti.getPageSettings',
+  'lti.fetchWindowSize',
+  'lti.frameResize',
+  'lti.put_data',
+  'lti.get_data',
+  'lti.showAlert',
+] as const
```
AdamFipke: neato, TIL about const assertions!

packages/frontend/app/contexts/LtiContext.tsx
```diff
@@ -0,0 +1,346 @@
...
+  const [capabilities, setCapabilities] = useState<SupportedMessages[]>()
+  const [pageContent, setPageContent] = useState<string>()
+  const [pageSettings, setPageSettings] = useState<PageSettings>()
+  const [windowSize, setWindowSize] = useState<WindowSize>()
+
+  const generateMessageId = () => {
+    const generate = (len: number) => {
+      let gen = ''
+      for (let i = 0; i < len; i++) {
+        gen += Math.floor(Math.random() * 16).toString(16)
```
AdamFipke: I'm a little confused here, what's the purpose of this random number message id? And what is an Lti Messenger?

packages/frontend/app/contexts/LtiContext.tsx
```diff
@@ -0,0 +1,346 @@
...
+    postMessage(window, 'lti.fetchWindowSize', undefined, lti_storage_target)
+  }, [window, lti_storage_target])
+
+  useEffect(() => {
+    if (windowSize) {
+      postMessage(window, 'lti.frameResize', {
+        height: windowSize.height - windowSize.offset.top,
+      })
+    }
+  }, [window, windowSize])
```
AdamFipke: wait, I just want to confirm. `window` and `windowSize` is static, right? As in it won't change when the browser's viewport changes? (and won't change if anything to do with `window` changes?). Or, alternatively, the cost of doing `postMessage` is really really small and can be spammed many times without consequence?

Because if it *does* change as the browser viewport changes, this can cause at least two big problems:
- Many many useEffect calls, but that can be fixed with a debounce 
- Certain things such as a keyboard opening on mobile will cause the viewport height to change and thus cause this useEffect to run (this was an issue with a Cascade component on the profile page a while back, where opening it on mobile would cause a browser resize causing it to close itself, it was kinda funny)

packages/frontend/app/lti/(embed)/(pages)/[cid]/layout.tsx
```diff
@@ -0,0 +1,74 @@
...
+          setGetCourseError(getErrorMessage(err))
+        })
+      if (succeeded) {
+        await API.course
+          .getCourseFeatures(courseId)
+          .then((courseFeatures) => {
+            setCourseFeatures(courseFeatures)
+          })
+          .catch((_) => {})
+      }
```
Comment (AdamFipke) [inline packages/frontend/app/lti/(embed)/(pages)/[cid]/layout.tsx]:
```suggestion
  useEffect(() => {
    const getData = async () => {
      await API.course
        .get(courseId)
        .then((course) => {
          setCourse(course)
          await API.course
          .getCourseFeatures(courseId)
          .then((courseFeatures) => {
            setCourseFeatures(courseFeatures)
          })
          .catch((err) => {setGetCourseError(getErrorMessage(err))})
        })
        .catch((err) => {
          setGetCourseError(getErrorMessage(err))
        })
```

packages/frontend/app/lti/(embed)/(pages)/page.tsx
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  useEffect(() => {
+    checkForceClose()
+  }, [window, window.self, window.top, searchParams])
+
+  useEffect(() => {
+    checkForceClose()
+  }, [])
+
```
AdamFipke: having two useEffects like this doesn't feel right but I can't think of an easy nicer way of doing it

packages/frontend/app/lti/(embed)/(pages)/page.tsx
```diff
@@ -0,0 +1,167 @@
...
+                        the page.
+                      </p>
+                      <p>
+                        If refreshing the page does not work, contact your
+                        instructor to ask them to connect their HelpMe course
+                        with this {lmsInfo.platform} course so you can be
+                        automatically added to their course.
+                      </p>
+                      <p>
+                        If you&#39;re logged into an account which does not
```
AdamFipke:
```suggestion
                        If you&#39;ve logged into HelpMe with a different email
```
New github PR editor has a nasty bug where I making suggestions across multiple lines switches files, so this is getting split across two suggestions

packages/frontend/app/lti/(embed)/layout.tsx
```diff
@@ -0,0 +1,55 @@
...
+      setContent(
+        <LtiContextProvider
+          window={global.window}
+          lti_storage_target={ltiStorageTarget ?? undefined}
+        >
+          {children}
+        </LtiContextProvider>,
+      )
+    }
+  }, [global.window, children, ltiStorageTarget])
```
AdamFipke: putting the whole component in a state set by useEffect seems maybe not good. I think instead react would have you do something like:
```tsx
if (global.window) {
   return <LtiContextProvider ...> {children} </Lticotnextprovder>
} else {
   return {children}
}
```

packages/frontend/app/contexts/LtiContext.tsx
bhunt02: when browser viewport or window size changes, this changes. this would only be trigger for the iframe its contained in. resizing the content doesn't override the window size. it also doesn't work effectively in general but it's used to scale the UI properly when in an iframe

packages/frontend/app/lti/(embed)/(pages)/page.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const [shouldClose, setShouldClose] = useState(false)
+  const [lmsInfo, setLmsInfo] = useSessionStorage<{
+    apiCourseId: string
+    platform: LMSIntegrationPlatform
+  }>('lms_info', null)
+
+  function checkForceClose() {
+    if (
+      window &&
+      window.self == window.top &&
```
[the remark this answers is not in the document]
bhunt02: all that does is check they have the same type... since these are objects i dont think it's necessary. thats more relevant to strictly check nullish/falsy values

packages/frontend/app/lti/(embed)/(pages)/page.tsx
```diff
@@ -0,0 +1,167 @@
...
+                        If refreshing the page does not work, contact your
+                        instructor to ask them to connect their HelpMe course
+                        with this {lmsInfo.platform} course so you can be
+                        automatically added to their course.
+                      </p>
+                      <p>
+                        If you&#39;re logged into an account which does not
+                        share the same email as your {lmsInfo.platform} account,
+                        the above options will not work. Ask your instructor for
+                        an invite link to their HelpMe course.
```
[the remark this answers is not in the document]
bhunt02: ?? for what? if they logged in with a bad email it means we don't have the right credentials on hand.

packages/frontend/app/lti/(embed)/layout.tsx
bhunt02: doesn't necessarily work with global.window not triggering rerenders but triggering in dependency arrays i believe, but maybe.

packages/frontend/app/api/server.ts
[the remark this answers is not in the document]
AdamFipke: ah right

packages/server/src/auth/auth-state.entity.ts
```diff
@@ -0,0 +1,31 @@
...
+} from 'typeorm';
+import { OrganizationModel } from '../organization/organization.entity';
+
+@Entity('auth_state_model')
+export class AuthStateModel extends BaseEntity {
+  @PrimaryColumn({ type: 'text' })
+  state: string;
+
+  @CreateDateColumn({ type: 'timestamp with time zone' })
+  createdAt: Date;
```
AdamFipke: oh neat! I didn't know you could put `timestamp with time zone` like that (I couldn't actually find it in the docs anywhere either). Do you think that other CreateDateColumns in the system should also follow this?

packages/server/src/auth/auth.service.spec.ts
```diff
@@ -145,50 +183,876 @@ describe('AuthService', () => {
...
+  });
+
+  describe('loginWithShibboleth', () => {
+    afterEach(() => {
+      roleChangeSpy?.mockClear();
+    });
+
+    // TODO: Remove test
+    // it('should throw an error when user already exists with password', async () => {
+    //   await UserModel.create({
```
AdamFipke: did you mean to remove these?

packages/server/src/auth/auth.service.ts
```diff
@@ -293,24 +778,27 @@ export class AuthService {
         organizationUser: true,
       },
     });
-    return user && user.organizationUser.organizationId === oid ? true : false;
+    return user != undefined && user.organizationUser?.organizationId == oid;
```
AdamFipke: I kinda detest the use of `!=` and `==` in js due to it having a lot of wacky bs gotchas (that people like to meme about, usually to try to make JS look bad. And the solution is always to use `!==` and `===` rather than have developers remember the gotchas with `!=` and `==`). Or I guess in short I think using `!=` and `==` hurts the readability of JS code 'cause they're a more loaded operation

packages/server/src/auth/auth-state.entity.ts
```diff
@@ -0,0 +1,31 @@
...
+  CreateDateColumn,
+  Entity,
+  JoinColumn,
+  ManyToOne,
+  PrimaryColumn,
+} from 'typeorm';
+import { OrganizationModel } from '../organization/organization.entity';
+
+@Entity('auth_state_model')
+export class AuthStateModel extends BaseEntity {
```
AdamFipke: I'm a little lost. What's the purpose of having an auth state? Maybe a comment could be added to describe?

packages/server/src/course/course.entity.ts
```diff
@@ -25,13 +25,14 @@ import { UnreadAsyncQuestionModel } from '../asyncQuestion/unread-async-question
...
 import { SuperCourseModel } from './super-course.entity';
 import { CourseChatbotSettingsModel } from '../chatbot/chatbot-infrastructure-models/course-chatbot-settings.entity';
+import { LtiCourseInviteModel } from '../lti/lti-course-invite.entity';
 
 @Entity('course_model')
 export class CourseModel extends BaseEntity {
   @PrimaryGeneratedColumn()
   id: number;
 
-  @OneToMany((type) => QueueModel, (q) => q.course)
```
AdamFipke: despite `type` being unused I think it was better for readability. Probably would've been better to add an underscore and make them all `_type` if your goal was to get rid of the linter warning

packages/server/src/auth/auth-state.entity.ts
```diff
@@ -0,0 +1,31 @@
...
+  @PrimaryColumn({ type: 'text' })
+  state: string;
+
+  @CreateDateColumn({ type: 'timestamp with time zone' })
+  createdAt: Date;
+
+  @Column({ type: 'integer', default: 60 })
+  expiresIn: number;
```
AdamFipke: expires in 60ms or 60s? Might be best to rename the variable name to expiresInMS for example. Same idea goes for the changes made to user-entity.ts. Could also change it to a date for `expiresAt`

packages/server/src/guards/lti.guard.ts
```diff
@@ -0,0 +1,24 @@
...
+  async setupData(response: any): Promise<{ token: IdToken }> {
+    const token = response.locals['token'];
+
+    return {
+      token,
+    };
+  }
+
+  async hasAuthorization(token: IdToken): Promise<boolean> {
+    return !!token;
```
AdamFipke: Idk the whole picture but doesn't this only just check if the user *has* a token, not necessarily that it's valid?

packages/server/src/lti/lti-auth.controller.ts
```diff
@@ -0,0 +1,213 @@
...
+  'r^\\/lti.*$',
+  'r^\\/api\\/v1\\/courses\\/[0-9]+(\\/features)?$',
+  'r^\\/api\\/v1\\/profile$',
+  'r^\\/api\\/v1\\/chatbot\\/question\\/suggested\\/[0-9]+$',
+  'r^\\/api\\/v1\\/semesters\\/[0-9]+$',
+  'r^\\/api\\/v1\\/chatbot\\/ask\\/[0-9]+$',
+  'r^\\/api\\/v1\\/chatbot\\/askSuggested\\/[0-9]+$',
+  'r^\\/api\\/v1\\/lms.*$',
+  'r^\\/api\\/v1\\/lti\\/auth.*$',
+];
```
AdamFipke: This is probably a very important piece of code for any future development for adding new features/endpoints that we want LTI to have access to. Unfortunately, I can't think of a better spot for it really but I thought I'd point it out anyway in case a future dev is looking at this PR

packages/server/src/lti/lti-course-invite.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+    onUpdate: 'CASCADE',
+  })
+  @JoinColumn({ name: 'courseId' })
+  course: CourseModel;
+
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @Column({ type: 'integer', nullable: true, default: 600 })
+  expires?: number;
```
AdamFipke: probably should be renamed to `expiresInSeconds` or something, since it might be unclear when looking in the database or certain pieces of code

AdamFipke (commented):
Backend code looks pretty good! Only minor comments/questions apart from the one potential security vulnerability thing I thought've with allowing people to login with SSO with legacy accounts. 

Also as a general thing with all PRs I guess, when I ask a question about some code it might be that it wasn't obvious to me why it's there or why it was done like that (which means it might be a pretty good candidate for a comment).

Note: I was more thorough with the parts of the system I was familiar with, I didn't review the new routes new code as closely

packages/server/src/auth/auth-state.entity.ts
bhunt02: probably best, yeah

packages/frontend/app/lti/(embed)/(pages)/layout.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export default Layout
+
+function IFrameWrapper({
+  children,
+}: {
+  children: React.ReactNode
+}): React.ReactNode {
+  const { windowSize } = useLtiContext()
```
AdamFipke: This is a pretty minor thing that I doubt can be fixed easily (if at all) but I mostly just want to document: on mobile there's a big empty gap below the iframe. 

Image

Image

BUT, something that can be improved is to reduce the bottom margin of the chatbot (probably just make it the same as the top margin):

Image

This is also probably the wrong spot for this comment (code-wise) but I think putting it in comments like this makes the replies more organized.

packages/frontend/app/lti/(embed)/components/LtiHeaderBar.tsx
```diff
@@ -0,0 +1,405 @@
...
+          <div className={'flex flex-col md:hidden'}>
+            <NavigationMenuItem>
+              <Link href={launchUrl} target={'_blank'}>
+                <ExpandOutlined className={'mr-2'} /> Open HelpMe
+              </Link>
+            </NavigationMenuItem>
+            <NavigationMenuItem className="md:hidden">
+              <SelfAvatar size={40} className="mr-2" />
+              {userInfo?.firstName}
+            </NavigationMenuItem>
```
AdamFipke:
```suggestion
```
On mobile the avatar and name should probably just be removed. I mistakenly tried to tap on it a bunch of times just to realise that it doesn't do anything.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -546,7 +554,8 @@ const Chatbot: React.FC<ChatbotProps> = ({
                 />
               )}
               <div ref={messagesEndRef} />
-              {courseFeatures.asyncQueueEnabled &&
+              {!isLti &&
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]: oh yeah I noticed the System chatbot isn't working but I imagine that's probably just because there isn't a defined helpme course in your .env for this staging setup. Ideally it might be good to have it set up so we can confirm it's still working (I think it should work but who knows)

Image

packages/frontend/app/lti/(embed)/(pages)/layout.tsx
```diff
@@ -0,0 +1,121 @@
...
+
```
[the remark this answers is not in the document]
AdamFipke: I should also note that there are no error messages. I just immediately get redirected to the login page (which come to think of it ONLY happen if the user has never logged in or they just logged out. There should probably always be an error message otherwise). I think it would be nice if there was some error message stating that the session expired or something. Probably just with adding some `err` in the query params.

Though, this might be impossible to implement depending on how things are set up exactly.

packages/server/src/auth/auth-state.entity.ts
bhunt02: basically, rather than embedding the information inside a cookie when we do login through external providers (like we did with google previously), we can use a 'state' query parameter that contains a key that can be perform two things: verify the request processed **is** valid for our system, and store information on our end to the nature of the request. we previously put the organization ID in a cookie for google login which was clunky, but with this we can store it in a database row and query for it using the state that the SSO provider sends back (generate state --> send to SSO provider with request details --> provider sends state back --> verify it)

packages/server/src/guards/lti.guard.ts
bhunt02: yeah, this one essentially just locks out LTI routes if there's not an ID token associated. it's just a secondary precaution that prevents the LTI routes from running at all if, say, the LTI-typescript middleware is not up and running, since that middleware performs the validation and generation of the id token from the user request itself. let me know if you need more info

packages/server/src/login/login.controller.ts
```diff
@@ -69,11 +71,23 @@ export class LoginController {
...
+      return res.status(HttpStatus.METHOD_NOT_ALLOWED).send({
         message: 'Organization does not allow login with username/password',
       });
     }
 
+    // Allow developers to use LEGACY auth on any account type,
+    // but don't allow this on production
+    if (
+      (isProd() && user.accountType != AccountType.LEGACY) ||
+      !user.password
```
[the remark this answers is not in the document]
bhunt02: (yes it was me, I do this in **development** so I can access any account)

packages/server/src/login/login.controller.ts
```diff
@@ -69,11 +71,23 @@ export class LoginController {
...
+    ) {
+      return res.status(HttpStatus.I_AM_A_TEAPOT).send({
+        message:
+          'Account was registered with SSO and cannot be accessed with email/password login',
+      });
```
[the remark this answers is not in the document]
bhunt02: :( - can't commit this directly as it'll break tests, but will keep it in mind

packages/frontend/app/lti/(embed)/(pages)/layout.tsx
bhunt02: you notice the different background color i'm sure, that's because the iframe isn't taking up the full height. this is more a canvas iframe embedding issue + sending us the wrong info w.r.t. the size of the browser window or something

packages/frontend/app/lti/(embed)/(pages)/layout.tsx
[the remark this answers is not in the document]
bhunt02: the issue after clicking 'open helpme' is because it has to overwrite the LTI-cookies to prevent a conflicting state.  i can just remove that button to prevent that issue in the first place. the mobile login stuff is a bit more complicated - maybe browser tabs/windows/sessions/cookies behave differently in the mobile environment? because what should happen is:

google auth -> open new tab with google window -> helpme validates the request and if successful, closes the tab because it's launched by the same JS source file, otherwise shows login page with error message (it also writes some cookies here) -> reloading the original tab will show dashboard

if this isn't occurring, i'm assuming a breakdown is occuring w.r.t. the session state variables and stuff. i'll have to investigate but i'm not sure how much i can fix that without adding yet another alternate auth pipeline behaviour

packages/frontend/app/lti/(embed)/components/LtiHeaderBar.tsx
AdamFipke: > i'm going to keep it so users can see who they're logged in as, at least

idk, if we want to keep that then it should probably be moved outside of this navigation drawer. Or do something to it so it doesn't look 'tappable', maybe by separating it from the other interactable elements, put it in a header somewhere, or do something with the styling to make it look static/in the background compared to the other interactable elements.

If you can think of a way to do this feel free, but I'd argue it's probably not too bad if it doesn't show who they're logged in as since the user is "already logged in" to canvas. 

Idk, could go either way with it.

packages/frontend/app/lti/(embed)/(pages)/layout.tsx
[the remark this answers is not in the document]
AdamFipke: oooooh yeah prefetching /logout would explain it. I remember I had that same issue when I remade the navbar during the great frontend refactor

packages/frontend/app/lti/(embed)/(pages)/page.tsx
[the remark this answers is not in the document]
AdamFipke: Though I guess sentry has functions for sending and capture events? So couldn't we just do some `captureEvent` in the backend (or frontend) whenever this case happens

packages/frontend/app/contexts/LtiContext.tsx
AdamFipke: Oh yeah idk if you have an Apple phone, but I realised that scrolling in Safari triggers resize events for some awful reason (I was running into this with my dad's site). Though I guess for us as long as the windowSize stays the same this should be fine

packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -462,11 +531,29 @@ const HeaderBar: React.FC = () => {
...
-  const courseId =
-    URLSegments[1] === 'course' && URLSegments[2] && Number(URLSegments[2])
-      ? Number(URLSegments[2])
-      : null
+
+  const courseId = useMemo(() => {
+    const temp = parseInt(URLSegments[2])
+    if (!isNaN(temp)) return temp
+    return undefined
+  }, [URLSegments])
```
AdamFipke: I think this might be a little *too* general. I think right now this would be fine but if we ever add some new pages that use this navbar where the 2nd URL segment is some number, then courseId will get set to whatever that number is. Previously there was a check for ` URLSegments[1] === 'course'`

bhunt02: ah. probably from merging w/ the lti version (which didn't care since LTI only uses path params w/ course id)

packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -75,7 +75,7 @@ const Link = ({
 }) => {
   const pathname = usePathname()
   const isActive = href === pathname
-  const isLogout = href === '/api/v1/logout'
+  const isLogout = href.startsWith('/api/v1/logout')
```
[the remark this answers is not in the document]
bhunt02: 🫡

packages/frontend/app/lti/(embed)/components/LtiHeaderBar.tsx
bhunt02: look, i removed it, but i only kept something like this so a user could see that they were logged into the wrong account without having to figure out why they don't have the right courses based on what courses they have on their dashboard. it's not interactable but i think part of the reason you struggled with trying to press it a bunch of times is that *your* brain is trained to do so with your experience developing on helpme, but there's no 'dropdown' indicator for this to indicate an (incorrect) interactability to it

let me know your thoughts, i would like to restore it if possible

AdamFipke: yeah, my prior experience could certainly be part of it lol. I guess I can also name a couple sites where you click on yourself to go to your profile page n whatnot. Though tbh I kinda forgot what it looked like. We can take a look at it tomorrow, I can probably come up with some small style adjustments. Maybe it's as simple as adding the text "Logged in as {your name}" and making it look a little different from the other nav links

packages/frontend/app/(auth)/login/page.tsx
```diff
@@ -1,36 +1,53 @@
...
+  useEffect(() => {
+    if (window.sessionStorage.getItem('lms_info') && !isLti) {
+      const params = new URLSearchParams(searchParams.toString())
+      const redirect = params.get('redirect')
+      if (redirect && !redirect.startsWith('/lti')) {
+        params.delete('redirect')
+      }
+      router.push(`/lti/login${params.size > 0 ? '?' + params.toString() : ''}`)
+    }
+  }, [])
```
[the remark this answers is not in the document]
AdamFipke: hoho it does work! weird

image

I changed the url in the same tab from the staging canvas url to the helpme deployment and it instantly redirected me to the lti version.

Probably not an issue, just thought it was interesting

AdamFipke (approved):
i think it should be ready!

Did some manual testing:

- Chrome on Desktop - works fine
- Brave on Mobile - works fine except I get stuck in an infinite loop of login page -> "login with google" -> click reload button -> login page. But auto login seems to work when refreshing the whole page
- Default android browser (very very old version) - Doesn't work at all, just gets a black screen with "MISSING_VALIDATION_COOKIE. Error validating ltik or IdToken" from the LTI middleware. This is probably some issue with this browser that stops cookies from being set correctly/securely. Not much to be done here I think but I thought I'd document it anyway. **EDIT**: Bridgette investigated with an old samsung browser: "they prevent canvas from receiving cookies written by us, and that prevents canvas from sending those cookies right back. That's the handshake; canvas initiates a login with info, we authorize it and write cookies we need, then canvas starts the session with those cookies sent back to us"
image

- On the Canvas app - Works great! After clicking "HelpMe" in canvas, it will open the iframe in a new Brave tab, but it all works well

More testing:

- Firefox on Desktop - Got an error message

image

- Edge on Desktop - Works fine

PR #393, opened 2025-09-11
Make semester start/end dates nullable

### LINKED ISSUE #391: Make Start/End dates optional fields for a Semester
To support test semesters, we would like to make the start/end date fields nullable in the Semester Model.

ribhavsharma:
# Description
Making start and end dates for semesters nullable to support test semesters.

Closes #391

## Type of change

# Checklist:


packages/frontend/app/(dashboard)/components/coursesSection.tsx
```diff
@@ -55,7 +55,7 @@ const CoursesSection: React.FC<CoursesSectionProps> = ({
         if (semesterA && !semesterB) return -1
         if (semesterB && !semesterA) return 1
 
-        if (semesterA && semesterB) {
+        if (semesterA && semesterB && semesterA.endDate && semesterB.endDate) {
           const diff =
             new Date(semesterB.endDate).getTime() -
             new Date(semesterA.endDate).getTime()
```
AdamFipke, line 61: i'll admit i'm having a little trouble with the sorting logic here, but I think this will make it so semesters without endDates will pretty much not be sorted? 

Shouldn't a diff be calculated with endDate being treated as -infinity if there's no endDate?

packages/frontend/app/(dashboard)/organization/components/SelectCourses.tsx
```diff
@@ -107,9 +107,15 @@ const SelectCourses: React.FC<SelectCoursesProps> = ({
...
-                ).toLocaleDateString()})`}
+                {`${semester.name} (${
+                  semester.startDate
+                    ? new Date(semester.startDate).toLocaleDateString()
+                    : '—'
+                } - ${
+                  semester.endDate
+                    ? new Date(semester.endDate).toLocaleDateString()
+                    : '—'
+                })`}
```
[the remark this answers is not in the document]
AdamFipke: Also, I guess you could add that extra logic for when there's only no startdate or only no enddate to that `formatSemseterDate()` function

packages/frontend/app/utils/timeFormatUtils.ts
```diff
@@ -155,35 +155,46 @@ export function formatDateAndTimeForExcel(date: Date | undefined): string {
...
+    if (startDate.getDate() > 25) {
+      adjustedStartDate.setMonth(adjustedStartDate.getMonth() + 1)
+    }
+    startMonth = adjustedStartDate.toLocaleString('default', { month: 'short' })
+    startYear = adjustedStartDate.getFullYear()
   }
-  const startDate = new Date(semester.startDate)
-  const endDate = new Date(semester.endDate)
-
-  // Adjust start month if day is > 25 (show as next month)
```
AdamFipke, line 161: mah comments are gone D:

AdamFipke (approved):
haven't manually tested it myself but code looks good!

packages/frontend/app/(dashboard)/components/coursesSection.tsx
```diff
@@ -55,17 +55,19 @@ const CoursesSection: React.FC<CoursesSectionProps> = ({
...
+          : -Infinity
+        const bEnd = semesterB?.endDate
+          ? new Date(semesterB.endDate).getTime()
+          : -Infinity
+        const diff = bEnd - aEnd
+        if (diff === 0) {
+          return a.course.name.localeCompare(b.course.name)
+        } else {
+          return diff
         }
```
bhunt02, line 70:
```suggestion
 /*
         Note that this is implicitly true at this point, but to avoid TypeScript errors
         we define the conditional statement:
         */
        if (semesterA && semesterB) {
          /*
          Try to place the semester at **some point** in time, whether it's the start or end date.
          Default to endDate, but use startDate if endDate does not exist, if neither exist, use -Infinity
          */
          const aTime = semesterA.endDate
            ? new Date(semesterA.endDate).getTime()
            : semesterA.startDate
              ? new Date(semesterA.startDate).getTime()
              : -Infinity;
          const bTime = semesterB.endDate
            ? new Date(semesterB.endDate).getTime()
            : semesterB.startDate
              ? new Date(semesterB.startDate).getTime()
              : -Infinity;
          
          const diff = bTime - aTime
          if (diff === 0) {
            return a.course.name.localeCompare(b.course.name)
          } else {
            return diff
          }
        }
        return a.course.name.localeCompare(b.course.name);
```

bhunt02: might wanna double check this in ur IDE, but i think this might be a better (if not a bit overly complex) way of going about it. this way we can still anchor the semester to some date even if the end date is nullish. also, yeah, that return at the end is technically unnecessary. we leave it as a fallback in case literally everything goes wrong i guess?

bhunt02 (changes requested):
i thiiiink everything here looks good... it's really important that we're careful with this as this is going to effect 1 page which is the most important page, the entrypoint to the rest of our app for users who've just logged in, so let's be extra sure.

i went through with my IDE to check usages but since i've got my repo inside the WSL filesystem on my windows device, it's really bad with checking imports like this. i've noticed it's missed some, so be extra diligent you've checked that on yours.

requesting a small change but we can discuss it and if you don't want to, you can skip it. just let me know and i'll give it a re-look

packages/frontend/app/(dashboard)/components/coursesSection.tsx
ribhavsharma: ide has no complaints!

PR #397, opened 2025-09-14
Redirect UBC accounts to CWL in registration flow

### LINKED ISSUE #377: Warn/Don't allow users to create an account with a UBC email when creating HelpMe account
Basically just add a regex in the "email" field when going to create an account that either warns or prevents the user from creating an account and instead suggests them to "Continue with UBC" to login with their CWL. For the regex I would just look for any "ubc" after the `@` symbol

ribhavsharma:
# Description

Currently, UBC users can create new accounts using their UBC email addresses. This PR introduces logic to prevent form submission when this is the case, redirecting the user to the CWL login instead. Also fixed the max width on the container div.

Closes #377

## Type of change

# Screenshot
image

# Checklist:

### DIFF
--- a/packages/frontend/app/(auth)/register/page.tsx
+++ b/packages/frontend/app/(auth)/register/page.tsx
@@ -6,6 +6,7 @@ import ReCAPTCHA from 'react-google-recaptcha'
 import { LeftOutlined } from '@ant-design/icons'
 import { useRouter } from 'next/navigation'
 import { userApi } from '@/app/api/userApi'
+import Link from 'next/link'
 
 interface Organization {
   id: number
@@ -78,7 +79,7 @@ export default function RegisterPage(): ReactElement {
   }
 
   // Check if email matches SSO patterns for the current organization
-  function checkSsoEmailPatterns(email: string): string | null {
+  function checkSsoEmailPatterns(email: string): ReactElement | null {
     if (
       !organization ||
       !organization.ssoEnabled ||
@@ -86,6 +87,18 @@ export default function RegisterPage(): ReactElement {
     ) {
       return null
     }
+    const errorMessage = (
+      <span>
+        SSO email detected! Would you like to{' '}
+        <Link
+          href={`/api/v1/auth/shibboleth/${organization.id}`}
+          prefetch={false}
+        >
+          Continue with {organization.name}
+        </Link>{' '}
+        instead?
+      </span>
+    )
     const patterns = organization.ssoEmailPatterns ?? []
     for (const pattern of patterns) {
       if (pattern.startsWith('r')) {
@@ -93,26 +106,19 @@ export default function RegisterPage(): ReactElement {
           const regexPattern = pattern.substring(1)
           const regex = new RegExp(regexPattern)
           if (regex.test(email)) {
-            return `SSO email detected! Would you like to "Continue with ${organization.name}" on the login page instead?`
+            return errorMessage
           }
-        } catch (error) {
+        } catch (_error) {
           console.warn(`Invalid regex pattern: ${pattern}`)
         }
       } else {
         if (email.includes(pattern)) {
-          return `SSO email detected! Please use "Continue with ${organization.name}" on the login page instead.`
+          return errorMessage
         }
       }
     }
     return null
   }
-
-  useEffect(() => {
-    setDomLoaded(true)
-    fetchOrganizations()
-    const orgId = parseInt(localStorage.getItem('organizationId') ?? '')
-    setOrganizationId(orgId)
-  }, [])
 
   useEffect(() => {
     setDomLoaded(true)
@@ -239,7 +245,7 @@ export default function RegisterPage(): ReactElement {
                       try {
                         const ssoError = checkSsoEmailPatterns(value)
                         if (ssoError) {
-                          return Promise.reject(new Error(ssoError))
+                          return Promise.reject(ssoError)
                         }
                         return Promise.resolve()
                       } catch (err) {
@@ -301,13 +307,9 @@ export default function RegisterPage(): ReactElement {
               </Form.Item>
 
               <Form.Item
-                label="Student Number"
+                label="Student Number (Optional)"
                 name="sid"
                 rules={[
-                  {
-                    min: 1,
-                    message: 'Student number must be at least 1 character',
-                  },
                   {
                     pattern: /^\d*$/,
                     message: 'Student number must be numeric',


bhunt02 (changes requested):
This might seem like an intuitively small issue, right? Your solution is simple, clean, great, nothing against it... but there's a little issue

So we've built a majority of the components of this app so that if a new organization is addef, it won't have conflicts with the UBC settings! It's meant to be modular so that it can be forked/deployed by other unis. It's a research project first and foremost, so we want to make it easy for other institutions to pick up

I think what may improve this is instead adding a column to organizations that can define a few (or a regex - I suggest saving the string with a leading "r" and then substringing it out and parsing the regex when it needs to be used) of email addresses they use for their SSO options, and prevent standard registration when those are detected. 

Would be easier to implement those on the backend as well, since this guards only the interface but not the business logic

packages/frontend/app/(auth)/register/page.tsx
```diff
@@ -178,6 +178,11 @@ export default function RegisterPage(): ReactElement {
                     max: 64,
                     message: 'Email must be at most 64 characters',
                   },
+                  {
+                    pattern: /^(?!.*@.*ubc).*$/i, // Only allow emails that do not contain "ubc" after @
+                    message:
+                      'UBC email detected! Please use "Continue with UBC" on the login page instead.',
+                  },
```
AdamFipke, line 185: so, after thinking about this more, i think it's probably best to not to fully prevent users from signing up with a ubc email. I think it would be better just to show a warning (maybe an antd Alert with type=warning just below the input box). Mostly because i imagine there's probably gonna be some edge case where someone needs to sign in with a secondary ubc email or something.

Also, as an idea:
- You could add a link inside the alert to take people to the login page `[on the login page](/login)`, it'll be a little faster than finding and clicking the back button
- Instead of above, you could make the alert "UBC email detected! Would you like to sign in with UBC instead?" with a link that that does the same as the "Continue with UBC" button, skipping the need to navigate back to the login page. The only thing with this is I would first verify how the frontend logic works with the UBC login; I'm fairly certain all it does is a navigation to a specific SSO link but if there's special headers being set or error handling then maybe this option isn't worth it

Also I assume you checked to make sure that the regex works with something@student.ubc.ca

AdamFipke (changes requested):
see comments

bhunt02:
W.R.T. Adam's comment about a similar solution:
I understand the desire to go for a simple solution at first, but it's better if we implement a robust, formal method for this than to just do it halfway I think. Let's try to avoid hardcoding for UBC as much as we can since we're trying to market this as a product for more than just UBC :)

AdamFipke:
> I would go with an easier more temporary approach of just adding a query param to the link to the "Create Account" link when UBC is selected in the dropdown so it'd be something like `/register?org=UBC` or `/register?orgid=1`. And then only show the warning if that query param is present (idea being to only show the warning if they click "Create Account" when UBC org was selected).

In regards to this, I looked at the code now and realised that the /register page already has an org id (which is retrieved from localstorage, i guess). So in which case all you would do is add a simple check if orgId = 1


> W.R.T. Adam's comment about a similar solution: I understand the desire to go for a simple solution at first, but it's better if we implement a robust, formal method for this than to just do it halfway I think. Let's try to avoid hardcoding for UBC as much as we can since we're trying to market this as a product for more than just UBC :)

Yes I agree, it's just that it's a lot more effort (new endpoint to get the regex, tests for it, fallbacks for if the request fails or if the user writes in the email field before the request completes (which with autofill is a possibility)) with no extra value added to our users and no extra value added to us unless we're very quickly expanding to more orgs. However, I do see your point that a different organization should be able to pick up the project and use it themselves, and any kind of orgId === 1 checks is gonna mess with that... actually yeah i think i remember now, I think I remember ramon mentioning to avoid hard-coding stuff for organizations. So yeah I think I'm wrong here. I might check with him tomorrow though

packages/frontend/app/(dashboard)/organization/settings/page.tsx
```diff
@@ -30,7 +30,26 @@ import { useOrganizationSettings } from '@/app/hooks/useOrganizationSettings'
...
+    try {
+      const formValues = await formSSO.validateFields()
+      const ssoEmailPatterns = formValues.ssoEmailPatterns || []
+      await API.organizations.patch(organization?.id ?? -1, {
+        ssoEmailPatterns,
+      })
+      message.success('SSO patterns updated')
+      setTimeout(() => {
+        window.location.reload()
+      }, 1750)
```
AdamFipke, line 44:
```suggestion
  setUserInfo({
  ...userInfo,
  userInfo.org: {
  ...org,
  ssoEmailPatterns: ssoEmailPatterns
  }
  })
```
I know there's another instance of this "set a timeout to refresh the page" code on this page, but it should properly update the local state variable. For context, the page refresh was done by another developer while I was first starting. They didn't really know how to update the local userInfo context so they just triggered a page refresh lol. They did the same on the profile page, though I hadn't gotten around to giving this page love yet.

Also the formatting and syntax might be a bit scuffed sorry.

You will also need to adjust the code below so that you can access the setUserInfo method.
```
const { userInfo, setUserInfo } = useUserInfo()
```

And maybe, if you have time, you could probably update the other window.location.reload() stuff to instead update the userInfo context with the correct org info.

packages/frontend/app/(dashboard)/organization/settings/page.tsx
```diff
@@ -30,7 +30,26 @@ import { useOrganizationSettings } from '@/app/hooks/useOrganizationSettings'
...
+    } catch (error: any) {
+      const errorMessage =
+        error?.response?.data?.message || 'Failed to update SSO patterns'
+      message.error(errorMessage)
```
AdamFipke, line 48:
```suggestion
      const errorMessage = getErrorMessage(error)
      message.error(errorMessage)
```
Just so you know for the future, we have a function called `getErrorMesssage`  so that way we have like a central way of retrieving the error messages we get from out endpoints (because the error messages aren't always in the same object format each time for some reason)

packages/frontend/app/(dashboard)/organization/settings/page.tsx
```diff
@@ -367,9 +400,9 @@ export default function SettingsPage(): ReactElement {
                     tooltip="SSO URL used by organization to authenticate user"
                   >
                     <Input
-                      allowClear={true}
+                      allowClear
                       defaultValue={organization?.ssoUrl}
-                      disabled={true}
```
AdamFipke, line 372: ..... hang on a sec, I actually don't see anything that handles the ssoUrl or organizationSSOEnabled variables. Have those just been not working on this form this entire time???

packages/frontend/app/(dashboard)/organization/settings/page.tsx
AdamFipke: oh god i just looked at this old code below more, *each form value has its own state variable* that's *so* gross (the `form` state already has all of these and is handled by antd). It also has a bunch of checks inside the submit function instead of using a validator function for the component (so the user would instantly get feedback about something being wrong instead of waiting until they click submit). I guess the person never read the antd docs for creating forms

packages/frontend/app/(auth)/register/page.tsx
```diff
@@ -7,9 +7,17 @@ import { LeftOutlined } from '@ant-design/icons'
...
+  id: number
+  name: string
+  ssoEnabled: boolean
+  ssoEmailPatterns?: string[]
+}
+
 export default function RegisterPage(): ReactElement {
   const [organizationId, setOrganizationId] = useState(0)
   const [domLoaded, setDomLoaded] = useState(false)
+  const [organization, setOrganization] = useState<Organization | null>(null)
```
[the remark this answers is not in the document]
ribhavsharma: woah, good catch! thanks :)

bhunt02 (approved):
Changes look good! Thanks for going above and beyond for this one after my pedantic complaints.

PR #403, opened 2025-09-21
Added bugfix

MithishR:
# Description

Fixed the `OrgOrCourseRolesGuard` to allow course professors to archive their own courses by modifying the `matchOrgRoles` method to return false instead of throwing an exception when users don't have the required organization role.

The `OrgOrCourseRolesGuard.matchOrgRoles()` method was throwing a `NotFoundException` when users lacked the required organization role, preventing the course role check from executing. This caused the guard to fail even when users had the required course professor role.

Closes # (362)

## Type of change

# How Has This Been Tested?
- Created a dummy course `COSC 404` and assigned `orgProfessor@ubc.ca` to it, however the user was unable to archive the course they are a professor for.
- Verified user has organization professor role (not admin) in UBCO.
- Tested database queries to ensure proper role relationships exist.
- Since the professor was unable to delete their course, checked if the admin (Ramon@ubc.ca) was able to delete it, and it was successful.
- Added console statements to isolate the root cause,
- Found the root cause and deleted the exception, replaced it with a  `return false` line. 

# Checklist:

### DIFF
--- a/packages/server/src/guards/org-or-course-roles.guard.ts
+++ b/packages/server/src/guards/org-or-course-roles.guard.ts
@@ -129,9 +129,7 @@ export class OrgOrCourseRolesGuard implements CanActivate {
       }>();
 
     if (!orgCourseUser) {
-      throw new NotFoundException(
-        'This course does not exist in your organization',
-      );
+      return false;
     }
 
     // then check if the user has the right org roles
--- a/packages/server/src/organization/organization.controller.ts
+++ b/packages/server/src/organization/organization.controller.ts
@@ -607,11 +607,11 @@ export class OrganizationController {
   @Patch(':oid/update_course_access/:cid')
   @UseGuards(
     JwtAuthGuard,
-    OrganizationRolesGuard,
-    OrganizationGuard,
+    OrgOrCourseRolesGuard,
     EmailVerifiedGuard,
   )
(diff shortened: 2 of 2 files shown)


packages/server/src/guards/org-or-course-roles.guard.ts
```diff
@@ -129,9 +129,7 @@ export class OrgOrCourseRolesGuard implements CanActivate {
       }>();
 
     if (!orgCourseUser) {
-      throw new NotFoundException(
-        'This course does not exist in your organization',
-      );
+      return false;
```
AdamFipke, line 132:
```suggestion
      throw new NotFoundException(
        'This course does not exist in your organization',
      );
```

Probably should reverse this change. See below:

> The OrgOrCourseRolesGuard.matchOrgRoles() method was throwing a NotFoundException **when users lacked the required organization role**, preventing the course role check from executing. This caused the guard to fail even when users had the required course professor role.

Are you sure? I went through the logic of the query. This error is only thrown if either A: the given course id is not in this organization, B: the user is not in this organization, or C: either the course or organization does not exist. 

If the user is in the organization without the right role, the check slightly below (github wouldn't let me select it for a comment) will run and return false:
```
 if (!roles.includes(orgCourseUser.ou_role)) {
      return false;
    }
```

PR #408, opened 2025-09-25
Module-linked pages toggle

Akshat-Kalra:
# Description

Adds a toggle for only ingesting module-linked pages to the vector store; namely the chatbot context.

I suppose this was a request by some of the instructors using our Canvas Pages Integration.

## Type of change

# How Has This Been Tested?

Hv to test more; so not ready to merge yet!!!

# Checklist:

### DIFF
--- a/packages/server/src/lmsIntegration/lmsIntegration.adapter.ts
+++ b/packages/server/src/lmsIntegration/lmsIntegration.adapter.ts
@@ -7,6 +7,8 @@ import {
   LMSCourseAPIResponse,
   LMSFile,
   LMSIntegrationPlatform,
+  LMSModule,
+  LMSModuleItem,
   LMSPage,
   LMSQuiz,
 } from '@koh/common';
@@ -79,6 +81,13 @@ export abstract class AbstractLMSAdapter {
   async getPages(): Promise<{
     status: LMSApiResponseStatus;
     pages: LMSPage[];
+  }> {
+    return null;
+  }
+
+  async getModules(): Promise<{
+    status: LMSApiResponseStatus;
+    modules: LMSModule[];
   }> {
     return null;
   }
@@ -389,6 +398,10 @@ class CanvasLMSAdapter extends ImplementedLMSAdapter {
 
     if (status != LMSApiResponseStatus.Success) return { status, pages: [] };
 
+    const moduleLinkedPageUrls = await this.getModuleLinkedPageUrls();
+
+    console.log(moduleLinkedPageUrls);
+
     const pages: LMSPage[] = [];
 
     // Individual page calls will now be cached by the Get() method
@@ -405,6 +418,7 @@ class CanvasLMSAdapter extends ImplementedLMSAdapter {
           url: page.url,
           frontPage: page.front_page,
           modified: new Date(pageResult.data.updated_at),
+          isModuleLinked: moduleLinkedPageUrls.includes(page.url),
         });
       }
     }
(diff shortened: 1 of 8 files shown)


Akshat-Kalra:
@FabianLCH Just added you as reviewer rn; but this is definitely not ready to merge yet. There is some race condition or caching issue in play where sometime toggling the button off deletes all the pages 💀.

AdamFipke:
Will this also be checking if the linked module is published/visible?

FabianLCH (changes requested):
Changes are looking good so far, left some comments. I noticed the following:

- Since the "Module-linked pages only" setting is updated separately from the selected resource types and does not trigger a re-sync, any 'unlinked' pages already synced will not get removed until (a) the next CRON job run or (b) the instructor manually forces synchronization. I'm wondering if it would be better to group the toggle with the resource selector so the instructor has to click 'Save & Resync' to save their preferences and automatically trigger a sync.
- The single document toggle sync endpoint doesn't appear to check for the `moduleLinkedPagesOnly` and a page's `isModuleLinked` setting to prevent force syncing an 'unlinked'. Not sure if this is intentional, but even if it is, given `syncDocuments` filters out unlinked pages from the LMS API response, the document would get automatically removed on the next bulk sync.

Akshat-Kalra:
> Will this also be checking if the linked module is published/visible?

Every document type is checked for if its published when its fetched; this would just filter on top of that.

Akshat-Kalra:
> @FabianLCH Just added you as reviewer rn; but this is definitely not ready to merge yet. There is some race condition or caching issue in play where sometime toggling the button off deletes all the pages 💀.

This should be fixed now, there was some caching issue in play.

packages/server/src/lmsIntegration/lmsIntegration.adapter.ts
```diff
@@ -422,6 +436,67 @@ class CanvasLMSAdapter extends ImplementedLMSAdapter {
...
+
+    if (status != LMSApiResponseStatus.Success) return { status, modules: [] };
+
+    const modules: LMSModule[] = data.map((module: any) => {
+      return {
+        id: module.id,
+        name: module.name,
+        items_url: module.items_url,
+      } as LMSModule;
+    });
```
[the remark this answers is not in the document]
AdamFipke: Yeah, I believe we had one professor who had their canvas set up with all the resources already published but only a few (of their several) modules published. Meaning that without checking if the modules themselves are published, their issue would still not be resolved.

Technically, I could also see a case where a professor may want the option to choose between "all module linked resources (including non-published modules)" and "all published-module linked resources", but I think that use case is probably more rare

packages/server/src/lmsIntegration/lmsIntegration.service.ts
```diff
@@ -579,6 +585,22 @@ export class LMSIntegrationService {
...
+            const allPages = items as LMSPage[];
+            const moduleLinkedPages = allPages.filter(
+              (page) => page.isModuleLinked === true,
+            );
+            if (moduleLinkedPages.length === 0 && allPages.length > 0) {
+              console.warn(
+                `Module-linked pages only setting is enabled for course ${courseId}, ` +
+                  `but no pages were detected as module-linked out of ${allPages.length} total pages. ` +
+                  `This might indicate that no pages are linked in course modules, or there's a ` +
+                  `configuration issue. Proceeding with filter - this will remove all page content from chatbot.`,
```
AdamFipke: hmm, it would be cool if this message could somehow be communicated to the user, but that might require too much of a refactor to be worth it

Akshat-Kalra:
> This might be a silly question, but is it only pages that can be linked into modules or can other resources be linked as well? And would it make sense to have the option to only synchronize module-linked resources (and not just pages)?

Any Canvas resource can be linked into modules, from what I've seen it is mostly used to organize pages but it is definitely used to organize all resources of the same module in a course.

I think for the pages toggle, it was an instructor requested feature that bridgette told me about.

I was initially thinking about adding a separate such toggle for every resource, but we could think about adding 1 for all resources.

maybe what we can do is, hv a small checkbox on top of every resource tab titled "sync from modules" or smth like that and then one button on the main resources selector that does it for every resource.

packages/server/src/lmsIntegration/lmsIntegration.service.ts
bhunt02: Agreed on this - but a simple way to implement this is putting a message on the frontend when the toggle is enabled saying 'no pages were detected' if we attempt to retrieve pages, and find none due to their not being any module-linked ones.

PR #414, opened 2025-10-11
Add Notification Feed for per-course-Alerts

### LINKED ISSUE #356: HelpMe Notifications Section
Rather than just have all notifications be through email, have a section where all notifications can be stored and viewed.

This was a suggestion from one of our users:
"Add a notification section on the website, similar to Workday’s "Notifications" or "Your Tasks" sections. This way, faculty and staff aren’t bombarded with emails the day before an exam, but can still get notifications through the website if they turn off email notifications."

ribhavsharma:
# Description

Added a notification feed on the header bar, using REPHRASE_QUESTION and Anytime Question updates as examples.
Extended the existing Alert infrastructure, adding a `deliveryMode` field to switch between modals, and the notification field. 

Closes #356 

## Type of change

# How Has This Been Tested?

Tested manually using interactions between the admin user and studentOne.

# Checklist:

# Demo

https://github.com/user-attachments/assets/15748030-5e5a-4366-aa82-d358001effd3

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -2234,6 +2234,18 @@ export enum AlertType {
   ASYNC_QUESTION_UPDATE = 'asyncQuestionUpdate',
 }
 
+// Allowed combinations to enforce front/back consistency
+export const FEED_ALERT_TYPES: AlertType[] = [
+  AlertType.DOCUMENT_PROCESSED,
+  AlertType.ASYNC_QUESTION_UPDATE,
+]
+
+export const MODAL_ALERT_TYPES: AlertType[] = [
+  AlertType.REPHRASE_QUESTION,
+  AlertType.EVENT_ENDED_CHECKOUT_STAFF,
+  AlertType.PROMPT_STUDENT_TO_LEAVE_QUEUE,
+]
+
 export enum AlertDeliveryMode {
   MODAL = 'modal',
   FEED = 'feed',
@@ -2358,6 +2370,9 @@ export class CreateAlertResponse extends Alert {}
 export class GetAlertsResponse {
   @Type(() => Alert)
   alerts!: Alert[]
+  @IsOptional()
+  @IsInt()
+  total?: number
 }
 
 // not used anywhere
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -2230,6 +2230,12 @@ export enum AlertType {
   REPHRASE_QUESTION = 'rephraseQuestion',
   EVENT_ENDED_CHECKOUT_STAFF = 'eventEndedCheckoutStaff',
   PROMPT_STUDENT_TO_LEAVE_QUEUE = 'promptStudentToLeaveQueue',
+  DOCUMENT_PROCESSED = 'documentProcessed',
+}
+
+export enum AlertDeliveryMode {
+  MODAL = 'modal',
+  FEED = 'feed',
 }
 
 export class AlertPayload {}
@@ -2238,11 +2244,18 @@ export class Alert {
   @IsEnum(AlertType)
   alertType!: AlertType
 
+  @IsEnum(AlertDeliveryMode)
+  deliveryMode!: AlertDeliveryMode
+
   @IsDate()
   sent!: Date
 
   @Type(() => AlertPayload)
   payload!: AlertPayload
+
+  @IsOptional()
+  @IsDate()
+  readAt?: Date
 
   @IsInt()
   id!: number
@@ -2264,6 +2277,14 @@ export class PromptStudentToLeaveQueuePayload extends AlertPayload {
   @IsInt()
   @IsOptional()
   queueQuestionId?: number
+}
+
+export class DocumentProcessedPayload extends AlertPayload {
+  @IsInt()
+  documentId!: number
+
+  @IsString()
+  documentName!: string
 }
 
 export class OrganizationCourseResponse {
@@ -2295,6 +2316,10 @@ export class OrganizationStatsResponse {
 export class CreateAlertParams {
   @IsEnum(AlertType)
   alertType!: AlertType
+
+  @IsOptional()
+  @IsEnum(AlertDeliveryMode)
+  deliveryMode?: AlertDeliveryMode
 
   @IsInt()
   courseId!: number
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -2231,6 +2231,7 @@ export enum AlertType {
   EVENT_ENDED_CHECKOUT_STAFF = 'eventEndedCheckoutStaff',
   PROMPT_STUDENT_TO_LEAVE_QUEUE = 'promptStudentToLeaveQueue',
   DOCUMENT_PROCESSED = 'documentProcessed',
+  ASYNC_QUESTION_UPDATE = 'asyncQuestionUpdate',
 }
 
 export enum AlertDeliveryMode {
@@ -2238,7 +2239,11 @@ export enum AlertDeliveryMode {
   FEED = 'feed',
 }
 
-export class AlertPayload {}
+export class AlertPayload {
+  @IsOptional()
+  @IsInt()
+  courseId?: number
+}
 
 export class Alert {
   @IsEnum(AlertType)
@@ -2267,9 +2272,6 @@ export class RephraseQuestionPayload extends AlertPayload {
 
   @IsInt()
   queueId!: number
-
-  @IsInt()
-  courseId!: number
 }
 
 export class PromptStudentToLeaveQueuePayload extends AlertPayload {
@@ -2285,6 +2287,27 @@ export class DocumentProcessedPayload extends AlertPayload {
 
   @IsString()
   documentName!: string
+}
+
+export class AsyncQuestionUpdatePayload extends AlertPayload {
+  @IsInt()
+  questionId!: number
+
+  @IsInt()
+  courseId!: number
+
+  @IsString()
+  @IsOptional()
+  subtype?:
+    | 'commentOnMyPost'
+    | 'commentOnOthersPost'
+    | 'humanAnswered'
+    | 'statusChanged'
+    | 'upvoted'
+
+  @IsString()
+  @IsOptional()
+  summary?: string
 }
 
 export class OrganizationCourseResponse {
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -2239,11 +2239,7 @@ export enum AlertDeliveryMode {
   FEED = 'feed',
 }
 
-export class AlertPayload {
-  @IsOptional()
-  @IsInt()
-  courseId?: number
-}
+export class AlertPayload {}
 
 export class Alert {
   @IsEnum(AlertType)
@@ -2272,6 +2268,9 @@ export class RephraseQuestionPayload extends AlertPayload {
 
   @IsInt()
   queueId!: number
+
+  @IsInt()
+  courseId!: number
 }
 
 export class PromptStudentToLeaveQueuePayload extends AlertPayload {
--- a/packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/TAQuestionCardButtons.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/TAQuestionCardButtons.tsx
@@ -7,6 +7,7 @@ import {
   UndoOutlined,
 } from '@ant-design/icons'
 import {
+  AlertDeliveryMode,
   AlertType,
   ClosedQuestionStatus,
   ERROR_MESSAGES,
@@ -159,6 +160,7 @@ const TAQuestionCardButtons: React.FC<TAQuestionCardButtonsProps> = ({
         courseId,
         payload,
         targetUserId: question.creatorId,
+        deliveryMode: AlertDeliveryMode.FEED,
       })
       // await mutateQuestions()
       message.success('Successfully asked student to rephrase their question.')
--- a/packages/frontend/app/api/index.ts
+++ b/packages/frontend/app/api/index.ts
@@ -37,6 +37,7 @@ import {
   DesktopNotifPartial,
   EditCourseInfoParams,
   GetAlertsResponse,
+  AlertDeliveryMode,
   GetAvailableModelsBody,
   GetChatbotHistoryResponse,
   GetCourseResponse,
@@ -1080,6 +1081,14 @@ class APIClient {
   alerts = {
     get: async (courseId: number): Promise<GetAlertsResponse> =>
       this.req('GET', `/api/v1/alerts/${courseId}`),
+    getAll: async (
+      mode: AlertDeliveryMode = AlertDeliveryMode.FEED,
+      includeRead: boolean = false,
+    ): Promise<GetAlertsResponse> =>
+      this.req('GET', `/api/v1/alerts`, undefined, undefined, {
+        mode,
+        includeRead: includeRead ? 'true' : 'false',
+      }),
     create: async (params: CreateAlertParams): Promise<CreateAlertResponse> =>
       this.req('POST', `/api/v1/alerts`, CreateAlertResponse, params),
     close: async (alertId: number): Promise<void> =>
--- a/packages/frontend/app/api/index.ts
+++ b/packages/frontend/app/api/index.ts
@@ -1079,20 +1079,30 @@ class APIClient {
   }
 
   alerts = {
+    markReadAll: async (): Promise<void> =>
+      this.req('PATCH', `/api/v1/alerts/mark-read-all`),
     get: async (courseId: number): Promise<GetAlertsResponse> =>
       this.req('GET', `/api/v1/alerts/${courseId}`),
     getAll: async (
       mode: AlertDeliveryMode = AlertDeliveryMode.FEED,
-      includeRead: boolean = false,
+      includeRead: boolean = true,
+      limit?: number,
+      offset?: number,
     ): Promise<GetAlertsResponse> =>
       this.req('GET', `/api/v1/alerts`, undefined, undefined, {
         mode,
         includeRead: includeRead ? 'true' : 'false',
+        ...(limit !== undefined ? { limit: String(limit) } : {}),
+        ...(offset !== undefined ? { offset: String(offset) } : {}),
       }),
     create: async (params: CreateAlertParams): Promise<CreateAlertResponse> =>
       this.req('POST', `/api/v1/alerts`, CreateAlertResponse, params),
     close: async (alertId: number): Promise<void> =>
       this.req('PATCH', `/api/v1/alerts/${alertId}`),
+    markReadBulk: async (alertIds: number[]): Promise<void> =>
+      this.req('PATCH', `/api/v1/alerts/mark-read-bulk`, undefined, {
+        alertIds,
+      }),
   }
 
   organizations = {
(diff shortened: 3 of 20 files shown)


packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -509,6 +515,9 @@ const HeaderBar: React.FC = () => {
           </h2>
         )}
       </div>
+      <div className="flex items-center">
+        <NotificationBell />
+      </div>
```
AdamFipke: I think the bell on mobile should be moved since it makes the page title/subtitle off-centre (also, for long course/queue names or on phones with increased text size, it may cause the title text to overflow/wrap easier which looks really bad)
Image

Proposed solution:

Place an antd Badge on the hamburger menu itself
Image

Place the notification bell somewhere in this drawer (though ngl I'm having some trouble thinking of an area to place it. Can't place it at the top since the university name can be longer. Can't place it beside the user's name since their name can be longer)
Image

---

### one option

Perhaps we can place it in-between the profile and logout buttons, similar to the desktop solution I proposed)
Image

But once the user clicks on it... I'm not sure what should happen, all the options I can think of are kinda gross:
- Have a submenu (similar to the Queues dropdown). However, it can't actually drop downwards due to the lack of space beneath so the menu, so it would need to drop "upwards" which feels kinda jank. Also it'd be really easy to accidentally close this notifications submenu by accidentally tapping, thus requiring you to re-open the hamburger menu and then re-open the notifications menu. Overall though, this isn't a *terrible* option i think
- Open a modal (kinda yucky imo, i think it would have higher cognitive load, but it could at least be re-used if the desktop also uses a modal for notifications)

---

### another option

Just say frick it and fit the notification bell up top where the university name is.

Image

This might get kinda iffy for future orgs, but at least now you can make a proper drop*down* submenu.

It is also a little more hidden up there imo so people might have a little more trouble finding it (at least in OC's case where the OC logo has a similar colour to the red "1" for notifications), but overall I think this is the better option.

packages/frontend/app/components/NotificationBell.tsx
```diff
@@ -0,0 +1,412 @@
...
+  [AlertType.PROMPT_STUDENT_TO_LEAVE_QUEUE]: {
+    title: 'Please leave the queue',
+    description:
+      'You have been inactive for a while. Please leave the queue if you no longer need assistance.',
+  },
+  [AlertType.EVENT_ENDED_CHECKOUT_STAFF]: {
+    title: 'Event has ended',
+    description:
+      'The event you were assisting with has ended. Please check out.',
+  },
```
AdamFipke: These alert types probably shouldn't be handled by the notification bell (and ideally there should be a typescript warning for developers/server error if they accidentally create a 'feed' alert for an alert that should only ever be a `modal` alert).

I think it would be better to create two subtypes/enums/arrays/whatever of AlertType called something like `PossibleFeedAlerts` and `PossibleModalAlerts`. Then, there should be a back-end check when creating alerts to ensure that, for example, someone can't create a PROMPT_STUDENT_TO_LEAVE_QUEUE 'feed' alert.

Then, in the frontend for the notification feed, you can assume that all alerts will only be part of `PossibleFeedAlerts` and then you won't need code like this (though I should note that this approach is actually somewhat fragile, so there should still be a frontend check that basically ensures the alert type is only one of the PossibleFeedAlerts and then discard it (with a console.error) if it's not one of these).

packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -355,6 +356,11 @@ const NavBar = ({
           ) : null}
           {/* DESKTOP ONLY PART OF NAVBAR */}
           <NavigationMenuItem className="!ml-auto hidden md:block">
+            <div className="px-2">
+              <NotificationBell />
```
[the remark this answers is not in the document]
bhunt02: i like the appearance as it is, actually. hiding it in a submenu just increases the number of clicks for a user to get to their notifications, and it goes against standard design seen on other websites which makes the mental load on the user worse

packages/frontend/app/components/HeaderBar.tsx
bhunt02: i agree with hiding it in the mobile sidebar menu thing though bc of the lack of usable space on mobile

packages/server/src/alerts/alerts.controller.ts
```diff
@@ -29,17 +32,68 @@ import { IsNull } from 'typeorm';
 export class AlertsController {
   constructor(private alertsService: AlertsService) {}
 
+  @Get()
+  async getAllAlerts(
+    @UserId() userId: number,
+    @Query('mode') mode?: string,
+    @Query('includeRead') includeRead?: string,
```
AdamFipke: I would use nestjs's integrated validation stuff for the query params. https://react.dev/reference/react/useContext

i.e. ParseBoolPipe, parseEnumPipe

packages/server/src/asyncQuestion/asyncQuestion.service.ts
```diff
@@ -55,6 +62,19 @@ export class AsyncQuestionService {
...
+        sent: new Date(),
+        userId: question.creator.id,
+        courseId: question.courseId,
+        payload: {
+          courseId: question.courseId,
+          questionId: question.id,
+          subtype: 'commentOnMyPost',
+          summary: `${commenterIsStaff ? commenter.name : 'Someone'} commented on your question`,
+        },
+      }).save();
```
AdamFipke: Maybe rename these functions from `sendNewCommentOnMyQuestionEmail` to `sendNewCommentOnMyQuestionEmailAndAlert` (not using the word `notification` here as to not confound with desktopNotifs and that Alerts are not considered Notifications on our backend). You could also instead opt to create separate functions, one for the emails, one for alerts, but then you'll be calling the exact same functions in the exact same areas.

packages/frontend/app/components/NotificationBell.tsx
```diff
@@ -0,0 +1,412 @@
...
+      content={content}
+      trigger="click"
+      placement="bottomRight"
+      open={open}
+      onOpenChange={(visible) => setOpen(visible)}
+      overlayStyle={{ padding: 0, borderRadius: 8 }}
+    >
+      {trigger}
+    </Popover>
+  )
```
AdamFipke: don't define components inside components in react where possible, it can cause some really weird state behaviour and performance issues (since the react engine might not see that the component was the same between re-renders and thus it will perform a full DOM element replacement which causes reflows and repaints even though it didn't need to).

Either make trigger its own component in the same file (with its own props n whatnot) or move all of its code content inside the Popover.

packages/frontend/app/components/NotificationBell.tsx
```diff
@@ -0,0 +1,412 @@
...
+
+  const unreadCount = useMemo(
+    () => alerts.filter((alert) => !alert.readAt).length,
+    [alerts],
+  )
+
+  const uniqueCourseIds = useMemo(() => {
+    const ids = new Set<number>()
+    for (const alert of alerts) {
+      if ((alert.payload as any)?.courseId) {
```
AdamFipke: I'm not going to go over the logic here in this file too much since I imagine a bunch of it might need to be changed as a result of https://github.com/ubco-db/helpme/pull/414/files#r2441324128

packages/frontend/app/components/HeaderBar.tsx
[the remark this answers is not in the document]
AdamFipke: > i like the appearance as it is, actually. hiding it in a submenu just increases the number of clicks for a user to get to their notifications, and it goes against standard design seen on other websites which makes the mental load on the user worse

This is true and I also thought of this. I guess it mostly just comes from paranoia from adding more stuff into the already crowded navbar, but yeah i thought about it more and there's probably easier pickings if we're really crammed on space. So I guess disregard the 'hiding it in a submenu' thing

packages/common/index.ts
```diff
@@ -2238,12 +2245,19 @@ export class Alert {
...
   @IsDate()
   sent!: Date
 
   @Type(() => AlertPayload)
   payload!: AlertPayload
 
+  @IsOptional()
+  @IsDate()
+  readAt?: Date
+
```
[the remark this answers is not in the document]
AdamFipke: > why not just paginate the retrieval route and show the notifications
> 
> `ORDER BY "readAt" DESC NULLS FIRST LIMIT [page size] OFFSET [page * page size]`
> 
> theres equivalents in find/findOne and querybuilder btw
> 
> paginating this wld be good

How do you mean by this?

Are you thinking the little notification popover should have buttons to see more notifications? Like it only shows say 10 notifications with a button to see the next 10?

I guess that could work (actually I just checked facebook and they do an infinite scroll it seems), but I think I would rather have it time-bound so that notifications from last year aren't still in the popover, which is likely since we don't give out a lot of notifications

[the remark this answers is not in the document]
ribhavsharma: sry about the super late reply but currently, there's a CTA next to each notif that says "mark as read" 
at the bottom of the popover there's a "mark all as read" button

i think it makes more sense to let the user mark a notif as read, rather than assuming they read whenever they click on the notifications icon. thoughts?

packages/frontend/app/components/HeaderBar.tsx
AdamFipke: this is still not resolved

Image

packages/server/src/alerts/alerts.controller.ts
```diff
@@ -29,51 +36,151 @@ import { IsNull } from 'typeorm';
 export class AlertsController {
   constructor(private alertsService: AlertsService) {}
 
+  @Patch('mark-read-bulk')
+  async markReadBulk(
+    @UserId() userId: number,
+    @Body('alertIds') alertIds: number[],
```
AdamFipke: also needs pipes

packages/server/src/alerts/alerts.controller.ts
```diff
@@ -29,51 +36,151 @@ import { IsNull } from 'typeorm';
...
+      .where('id IN (:...ids)', { ids: alertIds })
+      .andWhere('userId = :userId', { userId })
+      .andWhere('deliveryMode = :mode', { mode: AlertDeliveryMode.FEED })
+      .execute();
+  }
+
+  // Mark all unread FEED alerts for current user as read
+  @Patch('mark-read-all')
+  async markReadAll(@UserId() userId: number): Promise<void> {
+    console.log('Marking all feed alerts as read for user', userId);
```
AdamFipke:
```suggestion
```

packages/server/src/alerts/alerts.controller.ts
```diff
@@ -29,51 +36,151 @@ import { IsNull } from 'typeorm';
...
+      .set({ readAt: () => 'NOW()' })
+      .where('id IN (:...ids)', { ids: alertIds })
+      .andWhere('userId = :userId', { userId })
+      .andWhere('deliveryMode = :mode', { mode: AlertDeliveryMode.FEED })
+      .execute();
+  }
+
+  // Mark all unread FEED alerts for current user as read
+  @Patch('mark-read-all')
+  async markReadAll(@UserId() userId: number): Promise<void> {
```
AdamFipke: If you have time, adding some tests could be good:

- integrations tests for the new endpoints 
- a test or two for pruneOldReadFeed() inside alerts.service.spec.ts
- Add new tests for changes to existing endpoints/service functions

packages/server/src/asyncQuestion/asyncQuestion.service.ts
```diff
@@ -115,6 +138,31 @@ export class AsyncQuestionService {
...
+          alertType: AlertType.ASYNC_QUESTION_UPDATE,
+          deliveryMode: AlertDeliveryMode.FEED,
+          sent: new Date(),
+          userId: uid,
+          courseId: updatedQuestion.courseId,
+          payload: {
+            courseId: updatedQuestion.courseId,
+            questionId: updatedQuestion.id,
+            subtype: 'commentOnOthersPost',
+            summary: `${commenterIsStaff ? commenter.name : 'Someone'} commented on an Anytime Question you follow`,
```
AdamFipke:
```suggestion
            summary: `${commenterIsStaff ? commenter.name : 'Someone'} commented on an Anytime Question you commented on`,
```

Don't like using the term "follow" here since people might think they can "unfollow" it

packages/server/src/asyncQuestion/asyncQuestion.service.ts
```diff
@@ -115,6 +138,31 @@ export class AsyncQuestionService {
...
+          },
+        }).save(),
+      ),
+    );
```
AdamFipke: Can we also make  the `sendNeedsAttentionEmail` send a notification? Since some TAs can be unsbuscribed from emails but they should still be notified.

packages/server/src/asyncQuestion/asyncQuestion.service.ts
```diff
@@ -203,6 +251,19 @@ export class AsyncQuestionService {
           console.error('Failed to send email Human Answered email: ' + err);
           Sentry.captureException(err);
         });
+      await AlertModel.create({
```
AdamFipke: This could be made into its own smallish issue, but one thing that would be nice is to make it so the links in all of these asyncQuestion emails would auto-resolve the notification when you clicked on the link in the email (and you would do so via adding a query param, which you would check for on the anytime question page)

packages/frontend/app/components/NotificationBell.tsx
```diff
@@ -0,0 +1,404 @@
...
+                <List.Item
+                  className={`${item.isUnread ? 'bg-slate-100' : ''} transition-colors hover:bg-slate-50`}
+                  style={{
+                    padding: '8px 12px',
+                    borderBottom: '1px solid #f0f0f0',
+                  }}
+                  actions={
+                    item.ctaLabel
+                      ? [
+                          <Button
```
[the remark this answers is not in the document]
AdamFipke: You could do something like

top row: "New comment on your anyitme question yadayada"
2nd row: "14 minutes ago on <Tag> course name </Tag>"
3rd row: `<div className="grow">Ramon lawrence commented on yada yada </div><Link className="w-20"> Open</Link>`

packages/frontend/app/components/NotificationBell.tsx
```diff
@@ -0,0 +1,404 @@
...
+  const courseNameMap = courseResponses ?? {}
+
+  const markAsRead = async (alert: Alert) => {
+    if (alert.readAt) return
+    await markRead(alert.id)
+  }
+
+  const handleNavigate =
+    (alert: Alert, url: string) => async (): Promise<void> => {
+      await markAsRead(alert)
```
AdamFipke: I'm still a little iffy with this (specifically, that previously read alerts don't appear), since someone could click it by accident or forget and lose the notification forever.

The easiest solution to this is to **include read alerts inside the notification bell**, and then add a sorta "blue-ish" highlight to non-read notifications and a sorta "neutral/uninteresting" look to read notifications.

You already have the cron job for clearing out read notifications after a month, so we don't need to worry about overloading the UI.

Another thing is I would **include a notice that mentions "read alerts are cleared after 1 month"**, which I would squeeze as the last list item. Note that this little notice should be included as part of the pagination so that in the case where a user is trying to find a really old notification, they will eventually come to the last page just to see the notice (instead of seeing the history abruptly ending for no apparent reason).

Image

packages/frontend/app/components/NotificationBell.tsx
```diff
@@ -0,0 +1,404 @@
...
+        if (alert.alertType === AlertType.ASYNC_QUESTION_UPDATE) {
+          const payload = alert.payload as AsyncQuestionUpdatePayload
+          const destination = payload.courseId
+            ? `/course/${payload.courseId}/async_centre`
+            : undefined
+          const title =
+            payload.subtype === 'commentOnMyPost'
+              ? 'New comment on your Anytime Question'
+              : payload.subtype === 'commentOnOthersPost'
+                ? 'New comment on a followed Anytime Question'
```
AdamFipke:
```suggestion
                ? 'New comment an Anytime Question you commented on'
```
Don't like using the term "follow" here since people might think they can "unfollow" it

packages/frontend/app/components/NotificationBell.tsx
```diff
@@ -0,0 +1,404 @@
...
+                : payload.subtype === 'humanAnswered'
+                  ? 'Your Anytime Question was answered'
```
AdamFipke:
```suggestion
                  ? 'Your Anytime Question was answered by staff'
```

packages/server/src/asyncQuestion/asyncQuestion.service.ts
```diff
@@ -306,6 +380,19 @@ export class AsyncQuestionService {
           console.error('Failed to send email Vote Question email: ' + err);
           Sentry.captureException(err);
         });
+      await AlertModel.create({
```
AdamFipke: this is also still inside the subscription.

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/TAQuestionCardButtons.tsx
```diff
@@ -159,6 +160,7 @@ const TAQuestionCardButtons: React.FC<TAQuestionCardButtonsProps> = ({
         courseId,
         payload,
         targetUserId: question.creatorId,
+        deliveryMode: AlertDeliveryMode.FEED,
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/TAQuestionCardButtons.tsx]:
```suggestion
        deliveryMode: AlertDeliveryMode.MODAL,
```

AdamFipke (changes requested):
see comments

AdamFipke:
idk why, but github seems to have removed this PR from the list of PRs and we couldn't find it. I'm leaving this comment to see if it shows up

packages/server/src/alerts/alerts.controller.ts
AdamFipke: needed body class instead

packages/server/migration/1755000000000-add-async-question-alert-type.ts
```diff
@@ -0,0 +1,35 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddAsyncQuestionAlertType1755000000000
+  implements MigrationInterface
+{
+  name = 'AddAsyncQuestionAlertType1755000000000';
```
AdamFipke, line 6: ai generated, need to make a real one (same goes with other migration file)

packages/frontend/app/api/index.ts
```diff
@@ -1078,12 +1079,30 @@ class APIClient {
...
+        mode,
+        includeRead: includeRead ? 'true' : 'false',
+        ...(limit !== undefined ? { limit: String(limit) } : {}),
+        ...(offset !== undefined ? { offset: String(offset) } : {}),
+      }),
     create: async (params: CreateAlertParams): Promise<CreateAlertResponse> =>
       this.req('POST', `/api/v1/alerts`, CreateAlertResponse, params),
     close: async (alertId: number): Promise<void> =>
       this.req('PATCH', `/api/v1/alerts/${alertId}`),
+    markReadBulk: async (alertIds: number[]): Promise<void> =>
```
AdamFipke, line 1102: not actually used anywhere

AdamFipke:
I'm gonna close and re-open this PR since I've been making considerable changes to it

PR #420, opened 2025-10-28
Forgot Password Back-End TLC

AdamFipke:
# Description

Instigated by one of our users having a "User Not Found" error when they probably shouldn't've.

This PR:
- adds more detailed error messages to forgot password endpoint
- improved its robustness a little
- added several new tests
- Now made emails case-insensitive for it (@bhunt02 you might want to review this part in particular). Reason being is I could very well see someone try to reset their password but the case is wrong. I used typeorm's `ILike()` for this

**NEW**: No longer prevent unverified emails from receiving a password reset email. This fixes a scenario where someone malicious could create an account with someone else's email. If the real user then tries to login, they would just get "wrong password" with no option of resetting their password

## Type of change

# How Has This Been Tested?

Didn't manually test it, but I added a bunch of new tests. Local tests decided not to work today weeee

# Checklist:

### DIFF
--- a/packages/server/src/auth/auth.controller.ts
+++ b/packages/server/src/auth/auth.controller.ts
@@ -30,6 +30,7 @@ import * as bcrypt from 'bcrypt';
 import { CourseService } from 'course/course.service';
 import { LoginService } from '../login/login.service';
 import { UserId } from '../decorators/user.decorator';
+import { UserModel } from 'profile/user.entity';
 
 @Controller('auth')
 export class AuthController {
@@ -188,19 +189,19 @@ export class AuthController {
     });
   }
 
-  /* Should probably be renamed to forgotPassword() */
+  /* Could also be named forgotPassword (putting this here to be more ctrl+f-able) */
   @Post('/password/reset')
   async requestPasswordReset(
     @Body() body: PasswordRequestResetBody,
     @Res() res: Response,
   ): Promise<Response<void>> {
-    const { res: newres, user } =
-      await this.authService.validateResetPasswordParams(res, body);
-    if (newres.headersSent) {
+    let user: UserModel = undefined;
+    [res, user] = await this.authService.validateResetPasswordParams(res, body);
+    if (res.headersSent) {
       return;
     }
 
-    return await this.authService.issuePasswordReset(newres, user);
+    return await this.authService.issuePasswordReset(res, user);
   }
 
   @Post('register')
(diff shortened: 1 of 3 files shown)


AdamFipke:
@bhunt02 and I guess speaking of, other parts of the system (especially login and register) should probably have case-insensitive email query as well (if it doesn't already i sorta forgor) but I guess that's probably literally what you did with that one PR that we never merged

bhunt02:
we didn't merge it for other issues. should probably scrap it and just re-implement the same idea. we might have to delete some user accounts though - but to be fair most of the issue was google sending us weird emails and people with google accounts can't password reset anyway. the one problem would be is if it matched based on priority (they're probably stored internally ordered by ID in whatever file they're in, i assume) and the first match would be the relevant one. 

easy way around this: `accountType: AccountType.Legacy` or whatever

AdamFipke:
> we didn't merge it for other issues. should probably scrap it and just re-implement the same idea. we might have to delete some user accounts though 

Right okay yeah, could be saved for another PR, perhaps after the LTI stuff gets merged since that'll probably be a bunch of merge conflicts

> but to be fair most of the issue was google sending us weird emails and people with google accounts can't password reset anyway. the one problem would be is if it matched based on priority (they're probably stored internally ordered by ID in whatever file they're in, i assume) and the first match would be the relevant one.
> 
> easy way around this: `accountType: AccountType.Legacy` or whatever

Yeah, I guess that's sorta what I did with this PR. In writing this though I did realise I had a pretty big logic error in my code but I fixed that now (issue was that if there's multiple accounts with the same email, it could either be that the accounts have different account type OR be a case-sensitivity issue, neither of which *should* be possible)

packages/server/src/auth/auth.controller.ts
```diff
@@ -188,22 +188,19 @@ export class AuthController {
     });
   }
 
+  /* Should probably be renamed to forgotPassword() */
```
bhunt02: then do it 😭

packages/server/src/auth/auth.service.ts
```diff
@@ -465,47 +454,112 @@ export class AuthService {
       recaptchaToken,
       organizationId,
     }: { email: string; recaptchaToken: string; organizationId: number },
-  ): Promise<Response> {
+  ): Promise<{ res: Response; user?: UserModel }> {
```
bhunt02: if the array destructuring works and i've not lost my mind
```suggestion
  ): Promise<[Response,UserModel | undefined]> {
```

bhunt02 (commented):
looks good, i have slight comments but i'll approve after we discuss?

PR #421, opened 2025-11-04
Send Email Notification on Chatbot Answer Change

### LINKED ISSUE #305: Option to Email Notify students who asked this chatbot question when modifying answer
Under the Edit Chatbot Question modal, it would be cool if there were an option that would pop up after modifying an answer that allows the professor to automatically email everyone who asked that question of the updated answer!

Also maybe don't actually do a popup modal, since I could imagine those might accidentally be closed off since the user probably wouldn't be expecting it. It would probably be better to have a disabled checkbox nested underneath the 'answer' field saying "Email notify student(s) of updated answer?" that becomes enabled when the prof makes edits to the 'answer' field.

Image

ribhavsharma:
# Description

Added a checkbox in the edit chatbot answer modal, that lets the professor send out notifications to relevant students when a chatbot answer is changed. 

Closes #305 

## Type of change

# How Has This Been Tested?

Tested locally, the correct email shows up in the "sent emails" log file.

# Checklist:

### DIFF
--- a/packages/server/test/question.integration.ts
+++ b/packages/server/test/question.integration.ts
@@ -29,7 +29,6 @@ import {
   modifyMockNotifs,
   setupIntegrationTest,
 } from './util/testUtils';
-import { forEach } from 'lodash';
 import { QuestionTypeModel } from 'questionType/question-type.entity';
 import { StudentTaskProgressModel } from 'studentTaskProgress/studentTaskProgress.entity';
 
@@ -202,7 +201,7 @@ describe('Question Integration', () => {
 
       const questionTypes = [];
 
-      forEach(QuestionTypes, async (questionType) => {
+      for (const questionType of QuestionTypes) {
         await QuestionTypeFactory.create({
           name: questionType.name,
           color: questionType.color,
@@ -217,7 +216,7 @@ describe('Question Integration', () => {
           queueId: queue.id,
         };
         questionTypes.push(sendQuestionTypes);
-      });
+      }
 
       await TACourseFactory.create({ user, courseId: queue.courseId });
       expect(await QuestionModel.count({ where: { queueId: 1 } })).toEqual(0);
@@ -225,14 +224,7 @@ describe('Question Integration', () => {
       expect(response.status).toBe(403);
     });
     it('post question fails with non-existent queue', async () => {
-      const course = await CourseFactory.create();
       const user = await UserFactory.create();
-      const ta = await TACourseFactory.create({
-        course: course,
-        user: await UserFactory.create(),
-      });
-      // wait 0.25s to help deadlock issue?
-      await new Promise((resolve) => setTimeout(resolve, 250));
       await supertest({ userId: user.id })
         .post('/questions/999')
         .send({
@@ -266,14 +258,14 @@ describe('Question Integration', () => {
         course: course2,
       });
       const questionTypes = [];
-      forEach(QuestionTypes, async (questionType) => {
+      for (const questionType of QuestionTypes) {
         const currentQuestionType = await QuestionTypeFactory.create({
           name: questionType.name,
           color: questionType.color,
           cid: course2.id,
         });
         questionTypes.push(currentQuestionType);
-      });
+      }
 
       const response = await postQuestion(user, queueImNotIn, questionTypes);
       expect(response.status).toBe(404);
@@ -288,14 +280,14 @@ describe('Question Integration', () => {
         isDisabled: true,
       });
       const questionTypes = [];
-      forEach(QuestionTypes, async (questionType) => {
+      for (const questionType of QuestionTypes) {
         const currentQuestionType = await QuestionTypeFactory.create({
           name: questionType.name,
           color: questionType.color,
           cid: course.id,
         });
         questionTypes.push(currentQuestionType);
-      });
+      }
 
       const user = await UserFactory.create();
       await StudentCourseFactory.create({ user, courseId: queue.courseId });
@@ -370,14 +362,14 @@ describe('Question Integration', () => {
       });
 
       const questionTypes = [];
-      forEach(QuestionTypes, async (questionType) => {
+      for (const questionType of QuestionTypes) {
         const currentQuestionType = await QuestionTypeFactory.create({
           name: questionType.name,
           color: questionType.color,
           cid: course.id,
         });
         questionTypes.push(currentQuestionType);
-      });
+      }
 
       const response = await postQuestion(user, queue2, questionTypes);
 
@@ -710,14 +702,14 @@ describe('Question Integration', () => {
         status: OpenQuestionStatus.Drafting,
       });
       const questionTypes = [];
-      forEach(QuestionTypes, async (questionType) => {
+      for (const questionType of QuestionTypes) {
         const currentQuestionType = await QuestionTypeFactory.create({
           name: questionType.name,
           color: questionType.color,
           cid: course2.id,
         });
         questionTypes.push(currentQuestionType);
-      });
+      }
 
       const response = await postQuestion(user, queue2, questionTypes);
       expect(response.status).toBe(201);
@@ -751,14 +743,14 @@ describe('Question Integration', () => {
         status: OpenQuestionStatus.Drafting,
       });
       const questionTypes = [];
-      forEach(QuestionTypes, async (questionType) => {
+      for (const questionType of QuestionTypes) {
         const currentQuestionType = await QuestionTypeFactory.create({
           name: questionType.name,
           color: questionType.color,
           cid: queue.courseId,
         });
         questionTypes.push(currentQuestionType);
-      });
+      }
 
       const response = await postQuestion(user, queue, questionTypes);
       expect(response.status).toBe(201);
@@ -792,14 +784,14 @@ describe('Question Integration', () => {
         courseId: queue.courseId,
       });
       const questionTypes = [];
-      forEach(QuestionTypes, async (questionType) => {
+      for (const questionType of QuestionTypes) {
         const currentQuestionType = await QuestionTypeFactory.create({
           name: questionType.name,
           color: questionType.color,
           cid: queue.courseId,
         });
         questionTypes.push(currentQuestionType);
-      });
+      }
 
       const response = await postQuestion(user, queue, questionTypes);
       expect(response.status).toBe(201);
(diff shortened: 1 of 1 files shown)


packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/chatbot_questions/components/EditChatbotQuestionModal.tsx
```diff
@@ -158,8 +174,33 @@ const EditChatbotQuestionModal: React.FC<EditChatbotQuestionModalProps> = ({
 
     await API.chatbot.staffOnly
       .updateQuestion(cid, valuesWithId)
-      .then(() => {
+      .then(async () => {
         message.success('Question updated successfully')
+        const notify = (values as any).emailNotifyOnAnswerUpdate
+        if (notify) {
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/chatbot_questions/components/EditChatbotQuestionModal.tsx]:
```suggestion
        if (values.emailNotifyOnAnswerUpdate) {
```

Also, why not add `emailNotifyOnAnswerUpdate` to the FormValues type?

packages/common/index.ts
```diff
@@ -484,6 +485,10 @@ export interface UpdateChatbotQuestionParams {
     docId: string
     pageNumbersString: string
   }[]
+  // Frontend-only flag to request notification emails after an update.
+  // Server must validate the answer actually changed and will strip this before
+  // forwarding to the external chatbot service.
+  emailNotifyOnAnswerUpdate?: boolean
```
AdamFipke: Okay I read through the code now. I don't think this attribute is necessary and anything around it should be reverted.

The strategy is to add `emailNotifyOnAnswerUpdate` to `FormValues`. 

Then, for these lines of code:
```ts
const valuesWithId = {
      ...values,
      id: editingRecord.vectorStoreId,
      sourceDocuments: values.sourceDocuments || [],
    }
```

just add the following line:

```
const valuesWithId = {
      ...values,
      emailNotifyOnAnswerUpdate: undefined,
      id: editingRecord.vectorStoreId,
      sourceDocuments: values.sourceDocuments || [],
    }
```

This will make sure `emailNotifyOnAnswerUpdate` will not be sent to the backend and will not give you any typescript errors

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -378,6 +384,103 @@ export class ChatbotController {
...
+          </tr>
+          <tr>
+            <td style="font-weight:bold;">Answer (after)</td>
+            <td>${newAnswer}</td>
+          </tr>
+        </table>
+        <br/>
+        <p>You can revisit the course chatbot from your course page.</p>
+      </div>
+    `;
```
AdamFipke: Most of the body of this endpoint should be moved into its own method inside `chatbot.service.ts` since controllers are technically supposed to just handle like error checking and then calling the right services, and thus are supposed to be pretty slim.

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -378,6 +384,103 @@ export class ChatbotController {
...
   // resets all chatbot data for the course. Unused
+
+  // Professor-only: send notification email to all students who asked this question
+  // Body must include oldAnswer and newAnswer, and can optionally include question changes for context
+  @Post('question/:courseId/:questionId/notify')
+  @UseGuards(CourseRolesGuard)
+  @Roles(Role.PROFESSOR)
+  async notifyUpdatedAnswer(
+    @Param('courseId', ParseIntPipe) courseId: number,
+    @Param('questionId') questionId: string,
```
AdamFipke:
```suggestion
  @Post('question/:courseId/:vectorStoreId/notify')
  @UseGuards(CourseRolesGuard)
  @Roles(Role.PROFESSOR)
  async notifyUpdatedAnswer(
    @Param('courseId', ParseIntPipe) courseId: number,
    @Param('vectorStoreId') vectorStoreId: string,
```

Since chatbot questions are stored in two different databases (one in helpme DB, one in the chatbot vectorstore DB. And no it wasn't my decision), it's important to use consistent naming.

In this instance, `questionId` is referring to the vectorStoreId so it should probably be updated

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -378,6 +384,103 @@ export class ChatbotController {
...
+    // Enforce: only allow notify when answer changed
+    if ((oldAnswer ?? '').trim() === (newAnswer ?? '').trim()) {
+      throw new BadRequestException(
+        'Cannot send notification: answer was not changed.',
+      );
+    }
+
+    // Find all interactions in this course where this question was asked
+    const asked = await ChatbotQuestionModel.find({
+      where: { vectorStoreId: questionId },
```
AdamFipke:
```suggestion
      where: { 
        vectorStoreId: questionId
        course: {
          id: courseId 
      },
```

Saves needing to filter it later.

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/chatbot_questions/components/EditChatbotQuestionModal.tsx
```diff
@@ -158,8 +174,33 @@ const EditChatbotQuestionModal: React.FC<EditChatbotQuestionModalProps> = ({
...
+              {
+                oldAnswer: editingRecord.answer,
+                newAnswer: values.answer,
+                oldQuestion: editingRecord.question,
+                newQuestion: values.question,
+              },
+            )
+            if (resp?.recipients != undefined) {
+              message.success(
+                `Notification email sent to ${resp.recipients} student${
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/chatbot_questions/components/EditChatbotQuestionModal.tsx]:
```suggestion
                `Notification email sent to ${resp.recipients} user${
```

Now, interesting thing is this will also notify any TAs or Professors who have asked the question as well. Though, I could see that as still being useful (like if the prof was testing if it's working), so just the message needs to be updated

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -378,6 +384,103 @@ export class ChatbotController {
...
+         <tr><td style="font-weight:bold;">Question (after)</td><td colspan="2">${
+           newQuestion ?? ''
+         }</td></tr>`
+      : `<tr><td style="font-weight:bold;">Question</td><td colspan="2">${
+          oldQuestion ?? newQuestion ?? ''
+        }</td></tr>`;
+
+    const html = `
+      <div>
+        <p>The answer to a chatbot question you asked has been updated.</p>
```
AdamFipke: Oh right, it might be good to include the course name here too!

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -378,6 +384,103 @@ export class ChatbotController {
...
+            <td style="font-weight:bold;">Answer (after)</td>
+            <td>${newAnswer}</td>
+          </tr>
+        </table>
+        <br/>
+        <p>You can revisit the course chatbot from your course page.</p>
+      </div>
+    `;
+
+    await this.mailService
```
AdamFipke: One thought is to also make this into a mail subscription (that way, users have the option to opt-out of these emails). Though, unlike other emails, this case is more explicit where the staff member is explicitly saying "hey, please send these emails" compared to "do this other action where I'm unaware that this would send emails."

Idk if you have any opinions with this @bhunt02 

If we do decide to make it respect mail subscriptions, I think it could also be cool to mention in the response how many recipients there are total and how many were unsubscribed to this mail service (e.g. "5 users notified total (3 were unsubscribed from these notifications)").

Though, personally I think I'm leaning towards leaving it as it is (no subscriptions).

packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/chatbot_questions/components/EditChatbotQuestionModal.tsx
```diff
@@ -256,6 +298,25 @@ const EditChatbotQuestionModal: React.FC<EditChatbotQuestionModalProps> = ({
       >
         <Input.TextArea autoSize={{ minRows: 1, maxRows: 8 }} />
       </Form.Item>
+      <Form.Item
+        label="Email notify student(s) of updated answer?"
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/(settings)/settings/chatbot_questions/components/EditChatbotQuestionModal.tsx]: Image

Small thing, but I wanna try to "tuck" this form element under the "Answer" field a bit more. I tried playing around but couldn't come up with anything better really so honestly this is fine.

I did reduce the bottom margin of the answer field, and I think that might've helped a bit.
Image

AdamFipke (changes requested):
see comments

AdamFipke:
Oh right, and maybe adding a handful of integration or service tests for this would be a pretty good idea!

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -378,6 +384,103 @@ export class ChatbotController {
...
+      .sendEmail({
+        receiverOrReceivers: recipients,
```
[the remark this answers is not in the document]
AdamFipke: Resolved, just leaving unresolved so that this is found easier later.

packages/server/src/chatbot/chatbot.controller.spec.ts
```diff
@@ -0,0 +1,122 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { ChatbotController } from './chatbot.controller';
+import { ChatbotService } from './chatbot.service';
+import { ChatbotApiService } from './chatbot-api.service';
+import { MailService } from 'mail/mail.service';
+import { ChatbotQuestionModel } from './question.entity';
+import { BadRequestException } from '@nestjs/common';
+import { NotifyUpdatedChatbotAnswerParams, Role } from '@koh/common';
+
+describe('ChatbotController - notifyUpdatedAnswer', () => {
```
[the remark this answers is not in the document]
AdamFipke: > I think chatbot.service.ts does a lot of jobs already... should the FX for notifying not be delegated to a corresponding proviser?

Hmm maybe. Are you thinking a small-scoped provider (e.g. `chatbot-email-notification.service.ts`) or larger provider (e.g. `notification-email.service.ts`)?


Though I don't see us adding many more email notifications to do with the chatbot. And a larger one wouldn't make much sense imo since then it's a big random service file that gets imported to a bunch of other files for only 1 function.

If we're concerned about the size of chatbot.service.ts, it woudl probably make more sense to refactor out the ChatbotOrganizationSettings (or just ChatbotSettings) stuff  into its own service.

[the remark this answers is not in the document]
bhunt02: No what I mean is single responsibility principle. Notifications should be handled by one provider, not many

All the chatbot stuff falls under the purview of the chatbot provider

packages/frontend/app/api/index.ts
```diff
@@ -395,6 +395,26 @@ export class APIClient {
...
         this.req('DELETE', `/api/v1/chatbot/document/${courseId}/${docId}`),
+      notifyAnswerUpdate: async (
+        courseId: number,
+        vectorStoreId: string,
+        body: {
+          oldAnswer: string
+          newAnswer: string
+          oldQuestion?: string
+          newQuestion?: string
+        },
```
AdamFipke: Replace with NotifyUpdatedChatbotAnswerParams that you've created

PR #425, opened 2025-11-07
Prof Invites

### LINKED ISSUE #267: Prof invite link
**Description**

Right now it's kinda annoying to add new profs to the system, since we basically have to create the course, send them an invite link (which makes them a student), and then remember later to promote them to prof.

I think an easy feature fix would be to create a invite link that will automatically make the person a professor.

Probably best to hide the feature somewhere since it'd only be used by us admins

AdamFipke:
# Description

## UPDATE May 5th, 2026

- I've allowed Org Profs to also create Prof Invites for their courses since we have several professors that create and manage courses for their fellow professors. 

Other small things:
- I've made it so the search for getUsers endpoint for org admins now searches both email *and* name (previously, it would only search name)
- Moved the Course Invite Link section above the Course Features section since I think it's slightly more important
- I've added a little "Advanced" divider to the General Course Settings screen

image

I also have some other changes but I think i'll make another PR for that one so this one is already getting a little bloated.

## Original Description (pretty much all still stands)

- Adds Professor Invites (admin-only)
  - Expires (default after 1 week)
  - Has X number of uses (defaults to 1)
  - Emails the admin who created the invite when the invite gets consumed (as well as many edge and error cases, including names of those that attempt to use it)
  - Works the same as course invites, meaning it works with users that don't have accounts yet, aren't logged in, etc.
  - Note that admins don't need to actually be in the course to create a prof invite.

image
image
image

- Remembered to showcase it in the Organization Member Role History
image
image

- Comes with an admin view to show all prof invites (partially for fun, partially so that us admins can easily track what all of us are doing)
image

- Transformed the ProfessorSelector into its own component (it's used in 3 areas).
  - Added emails (since some admins have multiple accounts)
  - Fixed the search with it too
image

- After creating a course, admins now immediately get directed to creating a professor invite
image

- Added a highlightedCourse query param to /courses for a slight highlight. Right now it's only used when an org prof creates a course to highlight it. Could probably be used in some other situations.
image
image

- Refactored Queue+Course invites to use new QUERY_PARAMS constant (to easily connect query params between frontend and backend).
- Made it so `organization-roles.guard.ts` can take an `orgId` param (not just `oid)`, and to also throw a 500error if no `orgId` was given. Throws a 400 error if `!Number(orgId)`

Closes #267 

## Type of change

# How Has This Been Tested?

- Integration tests
- Service tests

- Modified `expectEmailsSent` to allow one to test the email body & subject (and will test to make sure there's no "null" or "undefined" or "object Object" or "{}" in the body & subject.

# Checklist:


docs/DEVELOPING.md
```diff
@@ -125,10 +125,10 @@ Unit test files should be colocated with the file they test.
 
 Integration tests are located in the `test` folder.
 
-`yarn test` at root level runs all tests, but you can also selectively run tests by running `yarn test` while inside a package.
```
AdamFipke: It took me *so* long to figure out what this meant and now I feel silly. It's literally just saying "if you run `yarn test` in `packages/frontend` it will run frontend tests. if you run `yarn test` in `packages/server` it will run backend tests."

We don't have frontend tests, so I just removed it to reduce confusion

packages/server/test/prof-invite.integration.ts
```diff
@@ -0,0 +1,379 @@
...
+
+      expect(res.body).toHaveLength(1);
+      const invite = res.body[0];
+      expect(invite).toHaveProperty('code');
+      expect(invite.course.name).toBe('Course A');
+      expect(invite.adminUser.email).toBe('admin@a.com');
+
+      expect(invite).not.toHaveProperty('organization');
+
+      expect(res.body).toMatchSnapshot([inviteMatcher]);
```
AdamFipke: This is hella nice, certainly gonna be using this more. Makes testing for certain properties way nicer.

Also for what `inviteMatcher` is: from the code above:
```ts
// used when testing with snapshots to say "hey, don't test for exact values for these" since they're gonna be different each time
  const inviteMatcher = {
    code: expect.any(String),
    createdAt: expect.any(String),
    expiresAt: expect.any(String),
  };
```
So in this instance it's saying "hey, the response body should match this snapshot that is an array with 1 object in it with a code, createdAt, and expiresAt properties"

PR #427, opened 2025-11-25
Add away toggle for TAs

### LINKED ISSUE #171: Re-add "away" toggle for TAs
**Description**

Once upon a time, I believe kevin added a way for TAs to toggle if they are currently away. This got commented out at some point and i have no idea why, maybe ask him.

ribhavsharma:
# Description

Adds back an away toggle for TAs per queue.

Closes #171 

## Type of change

- [ ] This change requires a database query to update old data on production. This query is below

# Checklist:

# Screenshot

image


packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx
```diff
@@ -58,15 +60,21 @@ const QueueInfoColumn: React.FC<QueueInfoColumnProps> = ({
...
-  // };
+  const me = queue?.staffList?.find((s) => s.id === userInfo?.id)
+  const isAway = me?.extraStatus === ExtraTAStatus.AWAY
+  const [savingAway, setSavingAway] = useState(false)
+  const toggleAway = async (checked: boolean) => {
+    // checked = Answering; unchecked = Away
+    const newStatus = checked ? null : ExtraTAStatus.AWAY
+    try {
+      setSavingAway(true)
+      await API.taStatus.setExtraStatus(cid, queueId, newStatus)
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx]: probably should add a catch for this

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx
```diff
@@ -113,6 +121,18 @@ const QueueInfoColumn: React.FC<QueueInfoColumnProps> = ({
       {/* buttons and queueUpToDateInfo for desktop (has different order than mobile)*/}
       <div className="hidden sm:block">
         <QueueUpToDateInfo queueId={queueId} />
+        {isStaff && (
+          <div className="mb-2 flex items-center gap-2">
+            <span className="text-sm text-[#5f6b79]">Away</span>
+            <Switch
+              checkedChildren="Answering"
+              unCheckedChildren="Away"
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx]:
```suggestion
```

I think I would just remove this if you already have a label for it saying "away"

packages/server/src/course/course.controller.ts
```diff
@@ -584,6 +586,94 @@ export class CourseController {
     return { queueId: queue.id };
   }
 
+  /**
+   * Allows a TA to set or clear their extra status (e.g., Away) for a specific queue.
+   */
+  @Patch(':id/ta_status/:qid')
```
AdamFipke: This endpoint is probably better suited to be inside queue.controller.ts since ultimately this sets the TA's status for just one queue (it's not really a "course-level" action)

packages/server/src/course/course.controller.ts
```diff
@@ -584,6 +586,94 @@ export class CourseController {
...
+    @User() user: UserModel,
+    @Body()
+    body: {
+      status: ExtraTAStatus | null;
+    },
+  ): Promise<QueueModel> {
+    const queue = await QueueModel.findOne({
+      where: {
+        id: queueId,
+        isDisabled: false,
```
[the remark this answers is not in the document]
AdamFipke: nvm i see that some other endpoints here also have `isDisabled` in their query. They should all probably be changed tbh

packages/server/src/queue/queue.service.ts
```diff
@@ -63,24 +64,38 @@ export class QueueService {
     const StaffHelpingInOtherQueues =
       await this.getStaffHelpingInOtherQueues(queueId);
 
+    // load any self-set extra TA statuses for staff in this queue
+    const staffStatusRows = await QueueStaffModel.find({
+      where: { queueModelId: queueId },
+    });
```
AdamFipke: this query should be unnecessary since doing queue.Stafflist should have the status

packages/server/src/queue/queue-staff.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+  PrimaryColumn,
+} from 'typeorm';
+import { QueueModel } from './queue.entity';
+import { UserModel } from '../profile/user.entity';
+import { ExtraTAStatus } from '@koh/common';
+
+// Map the existing join table used by the ManyToMany relation between QueueModel.staffList and UserModel.queues
+// This allows us to store additional metadata for a TA's status within a queue without refactoring all relations.
+@Entity('queue_model_staff_list_user_model')
+export class QueueStaffModel extends BaseEntity {
```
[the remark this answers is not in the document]
AdamFipke: Huh wow I tested it and apparently it does work?

I'm still not really a fan of this approach, since:
- It kinda adds a weird exception to how all our other entities work since it behaves differently.
- It's kinda hacky and against what the typeorm framework expects us to do
- Whenever you query the queue to map over the staff list (which happens in a couple different areas), you need to add another query *just* for getting the extraStatus attribute. This is because JoinTable makes it so you can't really access the attributes of the many-to-many table (in otherwords, *staffList serializes into an array of user entities instead of an array of "checkedInStaff" entities*. Like you should just be able to do `queue.stafflist.map((checkedInStaff) => { status: checkedInStaff.extraStatus, ...checkedInStaff.user })` kinda thing but you can't. I'm not explaining this super well and it wouldn't matter very much if it wasn't for the fact that our getQueue service method gets called *extremely* frequently, so an extra query can hurt. 

But on the plus side, this is a lot less refactoring required and therefore a *lot* less work.
Since right now, all places of our codebase still assumes that StaffList is an array of users...

Okay, maybe on second thought this is the way. The chances of something breaking from needing that big refactor is far higher than the likelihood of this typeorm patching this out.

The only thing I would say is to modify queue.entity.ts and user.entity.ts and add a comment for both of their staffList `@JoinTable` attributes saying "This should be refactored into its own proper entity instead of using JoinTable, but too many pieces of code still expect the old format so it's not worth it. If you want to modify the attributes of staffList, you can do so at queue-staff.entity.ts but be warned that you will not be included in the `.stafList` attribute and you will need to query QueueStaffModel directly"

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/StaffList.tsx
```diff
@@ -196,6 +200,8 @@ const StatusCard: React.FC<StatusCardProps> = ({
                   helpedAt={helpedAt}
                   extraTAStatus={ta.extraStatus}
                 />
+              ) : ta.extraStatus === ExtraTAStatus.AWAY ? (
+                <span className="text-red-500">Away</span>
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/StaffList.tsx]:
```suggestion
                "Will be back soon"
```

AdamFipke (changes requested):
See comments. Big things are:
- Tests
- Mobile
- Showing "Away" for Queue Invites page

Overall though, I think this is a pretty minimal implementation which I like

packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx
```diff
@@ -113,6 +121,18 @@ const QueueInfoColumn: React.FC<QueueInfoColumnProps> = ({
       {/* buttons and queueUpToDateInfo for desktop (has different order than mobile)*/}
       <div className="hidden sm:block">
         <QueueUpToDateInfo queueId={queueId} />
+        {isStaff && (
+          <div className="mb-2 flex items-center gap-2">
```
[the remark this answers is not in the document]
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/queue/[qid]/components/QueueInfoColumn.tsx]: Also on desktop, the switch could be centered

image

AdamFipke (approved):
Changes look good!


I also thought having the "Away" switch by the "Staff" header looked so good on mobile that I moved it there on desktop too.

before
Image

after

Image

Also the "Away" status on queue invite page looks to be working!

Image

PR #458, opened 2026-01-13
Paginate Anytime Questions

### LINKED ISSUE #434: Paginate Anytime Questions
Anytime questions I think just go on for infinity on the frontend (though it might be hard-capped at a few hundred somewhere). Probably should be paginated.

frasermuller:
# Description

Closes #434

This Issue adds pagination for the anytime questions (async questions) and adds a dev-only seed button to generate 100 dummy anytime questions for pagination testing.

Here is what was changed/added, hopefully it's clear enough to understand...

### 1. Seed 100 anytime questions (dev only)
packages/frontend/app/dev/page.tsx
- added button "Generate 100 Anytime Questions" on localhost:3000/dev
- button calls API.seeds.fillAnytimeQuestions()
- displays success or error message based on result.
I used the same styling/behavior from the other buttons above it in packages/frontend/app/dev/page.tsx

### 2. fillAnytimeQuestions() function
/helpme/packages/frontend/app/api/index.ts
- added new function that the frontend can call. when called it makes backend request to /api/v1/seeds/fill_anytime_questions.

### 3. backend endpoint that actually creates the 100 questions in the database.
/helpme/packages/server/src/seed/seed.controller.ts
- This was made so that I could test out the Pagination (when I implement it in the future) and make sure it actually works.
- it creates 100 questions using AsyncQuestionFactory (bridgettes tip: use createList instead of looping)
- Creates them in the **CS 304** course, and all questions are created by **studentOne.**
Screenshot 2026-01-07 at 10 41 16 PM

### 1. Imported pagination in Async Center (line 17)
file: packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
- Ant Design is used throughout the Project, so decided to use the pagination from that.

### 2. adding state variables
file: packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
- added 2 state variables to track **current page number** (starting at 1) and **page size** (defaulting to 20, ie. 20 questions per page) 
`const [page, setPage] = useState(1)`
`const [pageSize, setPageSize] = useState(20)`

### 3. updated hook to call page and pageSize
file: helpme/packages/frontend/app/hooks/useAsyncQuestions.ts
- added page and pagesize
-  **IMPORTANT: changed return type. instead of returning an array of async questions, it now returns an object with 2 things:
	- questions: the actual questions for this page
	- total: the total count of all questions (so we can calculate ex. "youre on page 2 of 10 pages")**
- added page and pageSize in the key string for the SWR (caching library) so that each page is its own separate cached request.
  - so like page 1 data is cached under key ending in ?page=1&pageSize=20
  - and page 2 data is cached under key ending in ?page=2&pageSize=20
         this means when you go back to page 1, SWR shows it instantly because its already in the cache. 
**basically changed it from "load everything at once" hook, to "load only what you need right now" hook, with caching so you dont refetch pages youve already seen.**

### 4. updated api call to include page and pageSize in api call, so the backend knows what page you want (lines 788 to 798)
file: helpme/packages/frontend/app/api/index.ts
- used same style as other apis in file for format

### 5. now that the hook returns a object with 2 properties (questions and total) I needed to update the hook to include 3 parameters instead of one.
/Users/frasermuller/helpme/packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
##### 4.  added `const totalQuestions = asyncQuestionsData?.total ?? 0`
this extracts the total count (which is needed for the pagination).
- "?."  safely accesses .total without crashing. needed cause when the page first loads, asyncQuestionsData is undefined (cause the data hasnt arrived yet.) without ?. , trying to access undefined.total would crash the app with an error. ?. only accesses .total if the object exists.
-  "?? 0 " if left side is undefined or null, it will use 0 instead, since the pagination component needs an actual number to work. 

### 6. adding pagination component (lines 437 - 454)
helpme/packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx

**used https://ant.design/components/pagination#pagination-demo-more  the "Changer" component style**

- OnChange fires when user clicks a page number or changes page size.
    - when page size changes, page is reset to page 1 (this fixes the possible issue where you would be on a ghost page, ie. User on page 10, they set the page size to a higher number, and they stay on page 10 whenthere is only 5 pages because they increased the number of questions displayed.
    

#### Problem with committing
Got this warning: `4:3  warning  'AsyncQuestion' is defined but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars`
Basically the AsyncQuestion was being imported but not used anymore.
Why? before, the hook returned an array of AsyncQuestion. i changed it so it returns an object, so the typescript can infer the type automatically now. we dont actually reference AsyncQuestion type anymore which makes the component unused. I added a comment on the line where I removed it.

##### When attempting to load questions after seeding. it would infinitely say loading questions and not display anything.
**Problem:** frontend was calling endpoint with ?page=1&pageSize=20 in the url, but the backend wasnt set up to read those params. it was also still returning just an array instead of the {questions,total} object the frontend was expecting. so the frontend was getting back the wrong data format and couldnt render anything which just caused infinite loading.

**FIX:** @Query decorator tells nestjs to grab the params from the URL query string. ParseIntPipe converts them from strings to numbers, then the questions are sliced to only send back the ones for the requested page, and return in the {questions,total} (object format). 

## PAGINATION IMPLEMENTED:
Screenshot 2026-01-12 at 7 06 47 PM

image

Screenshot 2026-01-12 at 7 09 45 PM

Screenshot 2026-01-12 at 7 10 18 PM

## Type of change

# How Has This Been Tested?

- [X] Test A: Added pagination test: tests that pagination works correctly by requesting page 1 with pageSize 2 and confirming we get back 2 questions and total count is correct.
Screenshot 2026-01-12 at 7 14 00 PM

- [X] Test B: I updated current tests to include the new Object response format rather than just the array. (you will see the small tweaks in the asyncQuestion.integration.ts its like 3 different changes repeated for all the aysncQuestion integration tests.)

All of the**UPDATED** integration tests for async questions pass **INCLUDING.** the pagination test I added.

Run `yarn test:integration asyncQuestion.integration`

Screenshot 2026-01-12 at 7 11 56 PM

- [X] Test C: I have also tested the feature thoroughly in the local dev environment. I have tested each button in the pagination feature, aswell as testing possible breaks (the ghost page i mentionned earlier). all buttons work. See photo above

**TO TEST YOURSELF** 
- open local repo
- run:  `yarn dev`
- go to localhost:3000/dev
- press `Seed Data`,   then `Generate 100 Anytime Questions`
- Go to localhost:3000
- Click on  **CS 304** Course
- Click on **Anytime Q's** at the top
- Test out feature (it is at the bottom of the page

# Checklist:

- [X] I have performed a code review of my own code (under the "Files Changed" tab on github) to ensure nothing is committed that shouldn't be (e.g. leftover `console.log`s, leftover unused logic, or anything else that was accidentally committed)
- [X] I have commented my code where needed 
- [unsure if needed] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [updated current tests] I have checked that new and existing tests pass *locally* with my changes
- [X] Any work that this PR is dependent on has been merged into the main branch
- [Not yet checked] Any UI changes have been checked to work on desktop, tablet, and mobile

### DIFF
--- a/packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
+++ b/packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
@@ -208,6 +208,11 @@ export default function AsyncCentrePage(
 
   const totalQuestions = displayedQuestions.length // total length after all filters applied
 
+  // reset to page 1 whenever the filtered question count changes.
+  useEffect(() => {
+    setPage(1)
+  }, [displayedQuestions.length])
+
   const paginatedQuestions = useMemo(() => {
     const startIndex = (page - 1) * pageSize //calculates where to start slicing
     const endIndex = startIndex + pageSize // and where to stop slicing


packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -1,7 +1,7 @@
 'use client'
 
 import {
-  AsyncQuestion,
+  //Removed AsyncQuestion import because its not used directly in this file anymore
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx]: ya if its unused just safely remove it, no need to add a comment itll be visible in the lines removed

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -59,7 +67,17 @@ export default function AsyncCentrePage(
   const { userInfo } = useUserInfo()
   const role = getRoleInCourse(userInfo, courseId)
   const isStaff = role === Role.TA || role === Role.PROFESSOR
-  const [asyncQuestions, mutateAsyncQuestions] = useAsnycQuestions(courseId)
+
+  const [page, setPage] = useState(1)
+  const [pageSize, setPageSize] = useState(20)
+  const [asyncQuestionsData, mutateAsyncQuestions] = useAsnycQuestions(
```
Comment (bhunt02) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx]: this spelling is... interesting

packages/frontend/app/api/index.ts
```diff
@@ -786,8 +786,16 @@ export class APIClient {
       ),
   }
   asyncQuestions = {
-    get: async (cid: number): Promise<AsyncQuestion[]> =>
-      this.req('GET', `/api/v1/asyncQuestions/${cid}`, undefined),
+    get: async (
+      cid: number,
+      page: number,
+      pageSize: number,
+    ): Promise<{ questions: AsyncQuestion[]; total: number }> =>
```
bhunt02: would be good to declare a DTO type shared between FE and BE in common/index.ts

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -742,9 +743,11 @@ export class asyncQuestionController {
   @Roles(Role.STUDENT, Role.TA, Role.PROFESSOR)
   async getAsyncQuestions(
     @Param('courseId', ParseIntPipe) courseId: number,
+    @Query('page', ParseIntPipe) page: number,
+    @Query('pageSize', ParseIntPipe) pageSize: number,
     @UserId() userId: number,
     @Res() res: Response,
-  ): Promise<AsyncQuestion[]> {
+  ): Promise<{ questions: AsyncQuestion[]; total: number }> {
```
bhunt02: same DTO gripe as above

bhunt02 (changes requested):
some comments, nothing serious. i can give it another look once addressed

packages/frontend/app/hooks/useAsyncQuestions.ts
```diff
@@ -4,17 +4,21 @@ import { API } from '../api'
 
 export function useAsnycQuestions(
```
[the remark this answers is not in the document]
AdamFipke: oh my god...

okay that's actually really funny. Yeah I support this. Luckily it seems it would be a very small refactor anyway

image

packages/frontend/app/api/index.ts
frasermuller: yes for sure, I will make those changes

AdamFipke (changes requested):
Okay things were looking very solid! Just some minor comments...

But then I found one big issue that unfortunately throws a wrench into this PR a little bit.

All of these filters are *frontend-only* filters. They only filter out data that exists on the frontend. Because of this, things kinda break when you use them.
Image
(staff filters)
image
(student filters)

The Sort By would also break
image


Here, you can see how on page 1 there are no questions that are verified but on page 5 there are a few.
Image
Image

There are 2 approaches to fixing this:

- A: Move all the async questions to the frontend rather than paginating the endpoint, where the pagination and filtering features are all handled on the frontend. This approach will probably be faster UX up until a point (maybe in the thousands of questions or tens of thousands of questions). Unfortunately, this means you may need to scrap a bunch of the changes you made here.
- B: Move these filter parameters (question types, isVisible, isVerified, allQuestions vs just mine) and the Sort By parameters all on the backend. This would be a fair bit of work I think, and adds a fair bit of loading anytime the user interacts with any of the filters.

Personally, I'm a little more fond of option A since right now the whole page goes loading anytime you select a different page (see image below). This is also the way we handle pagination for our chatbot and queue questions. It won't scale to the many thousands of questions, but I don't think we need to.

Image

packages/frontend/app/hooks/useAsyncQuestions.ts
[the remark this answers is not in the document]
frasermuller: just fixed it and its references

packages/server/src/asyncQuestion/asyncQuestion.controller.ts
```diff
@@ -931,8 +930,7 @@ export class asyncQuestionController {
       return temp;
     });
 
-    res.status(HttpStatus.OK).send(questions);
-    return;
+    return questions as unknown as AsyncQuestion[];
```
AdamFipke: I was gonna say this would be a big nono (since it basically disables typescript) but in this case a lot of the code above is doing stuff that typescript wouldn't be able to calculate (and already suppresses typescript) and would require a big ol' refactor to fix so this is fine

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -199,6 +206,14 @@ export default function AsyncCentrePage(
 
   const displayedQuestions = useMemo(() => applySort, [applySort])
 
+  const totalQuestions = displayedQuestions.length // total length after all filters applied
+
+  const paginatedQuestions = useMemo(() => {
+    const startIndex = (page - 1) * pageSize //calculates where to start slicing
+    const endIndex = startIndex + pageSize // and where to stop slicing
+    return displayedQuestions.slice(startIndex, endIndex)
+  }, [page, pageSize, displayedQuestions])
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx]: I have found a bug.

If I'm on page 1 and set a filter to "verified", it shows these 3 questions (which is correct)

Image

But if I'm on page 2 and set the filter to "verified", it shows 0 questions. It still says there's 3 questions at the bottom oddly enough

Image

Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx]: Also, if I change the pageSize, it will fix it and show all 3 questions again. So my guess is it's another useMemo or useEffect dependency issue. This is kinda weird since displayedQuestions is already inside the paginatedQuestions dependency array.

packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx
```diff
@@ -177,7 +183,8 @@ export default function AsyncCentrePage(
   )
 
   const applySort = useMemo(() => {
-    return applyCreatorFilter.sort((a, b) => {
+    return [...applyCreatorFilter].sort((a, b) => {
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/page.tsx]: good catch with this btw, I think this might've fixed another bug that I think one user mentioned to me a long time back to do with the order of sorting and filtering. I think when I was first making edits to this component I didn't understand react as much. But yeah a lot of stuff with useState, useEffect, useMemo, etc. are supposed to return new objects rather than modify old ones (supposed to be immutable. Stuff shouldn't change inputs no matter how many times it runs). EDIT: found the article it reminded me of https://react.dev/learn/updating-objects-in-state

PR #460, opened 2026-01-13
Fixed pagination for Organization Users table

### LINKED ISSUE #441: Organization Users table is showing incorrect number of pages
**Description**

The `UsersTable.tsx` and its associated endpoint (`@Get(':oid/get_users{/:page}')`) does not show/handle the total number of users (see below)

Image

The endpoint needs to be modified to return the total number of users *in addition to* the 50 (or less) users for that page (the return-type for the endpoint will need to be modified).

The query inside `getUsers()` inside organiztion.service will need to be updated to include the COUNT (or perhaps make a seperate query by doing `OrganizationUser.count({orgId})` kinda thing)

mahigangal:
# Description

The Organization Users table now returns and displays the real total user count. Pagination displays only the actual number of pages, and the Next/Previous controls are enabled only when multiple pages are available. Each page shows up to 50 users. Previously, the total was hardcoded to 5000, which created ~100 phantom pages.

Backend `@Get(':oid/get_users{/:page}')` now uses `getManyAndCount()` to return both the paginated users and the total count, instead of relying on a fixed total. The frontend `UsersTable` now reads `response.total` and passes it to the Pagination component, so page counts and Next/Previous controls accurately reflect the total. Each page still shows up to 50 users.

Screenshot 2026-01-12 at 11 24 02 PM

Closes #441 

## Type of change

# How Has This Been Tested?

Manual Tests: The maximum number of users that can be displayed on a single page is 50. Right now, there are only 6 users in the system; therefore, the system only shows one page with the Next arrow disabled. I tested the functionality by temporarily setting the maximum number of users that can be displayed on a single page to 5. In this case, I could visit page 1 (showing 5 users) and page 2 (showing 1 user) using the Next and Previous arrows. When multiple pages get created, 'Go to _ page' is visible and works to go to a specific page. 

Unit tests (organization.service.spec.ts): Updated assertions to expect {users, total} structure instead of just users array

organization.integration.ts.snap: Updated snapshot to match new API response format with total count

# Checklist:

### DIFF
--- a/packages/server/src/organization/organization.service.ts
+++ b/packages/server/src/organization/organization.service.ts
@@ -196,12 +196,9 @@ export class OrganizationService {
       });
 
     if (search) {
-      const likeSearch = `%${search.replace(/\s+/g, '').toUpperCase()}%`;
-
-      organizationUsers.andWhere(
-        `UPPER(REPLACE(user.name, ' ', '')) LIKE :search`,
-        { search: likeSearch },
-      );
+      organizationUsers.andWhere(`user.name ILIKE :search`, {
+        search: `%${search}%`,
+      });
     }
 
     organizationUsers


packages/server/src/organization/organization.controller.ts
```diff
@@ -1431,9 +1431,8 @@ export class OrganizationController {
     @Param('oid', ParseIntPipe) oid: number,
     @Param('page', ParseIntPipe) page: number,
     @Query('search') search: string,
-  ): Promise<OrgUser[]> {
+  ): Promise<{ users: OrgUser[]; total: number }> {
```
bhunt02: use the DTO class you wrote in common/index.ts

packages/server/src/organization/organization.service.ts
```diff
@@ -185,58 +185,42 @@ export class OrganizationService {
...
-      const likeSearch = `%${search.replace(' ', '')}%`.toUpperCase();
+      const likeSearch = `%${search.replace(/\s+/g, '').toUpperCase()}%`;
+
       organizationUsers.andWhere(
-        new Brackets((q) => {
-          q.where('UPPER("UserModel".name) like :searchString', {
-            searchString: likeSearch,
-          });
-        }),
+        `UPPER(CONCAT(user.firstName, user.lastName)) LIKE :search`,
```
bhunt02: did you make sure to test the search out, btw? i'm getting some red flags - postgres requires quotations around camel case properties and typeorm qb doesnt always infer

mahigangal: Thank you for pointing this out. I have fixed this query with `UPPER(REPLACE(user.name, ' ', '')) LIKE :search` to use the database name column, instead of raw concatenation. I am also removing the spaces from both the search input and the name column to ensure matching. 

I tested the search, and it is working with all edge cases I could think of. Please let me know if there are still any red flags.

packages/server/src/organization/organization.service.ts
```diff
@@ -185,58 +185,42 @@ export class OrganizationService {
...
-        UserModel,
-        'UserModel',
-        'UserModel.id = OrganizationUserModel.userId',
-      )
-      .where('OrganizationUserModel.organizationId = :organizationId', {
+  ): Promise<{ users: OrgUser[]; total: number }> {
+    const organizationUsers = OrganizationUserModel.createQueryBuilder(
+      'orgUser',
+    )
+      .leftJoinAndSelect('orgUser.organizationUser', 'user')
```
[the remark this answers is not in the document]
bhunt02: had a feeling...

bhunt02 (approved):
i'd wait for adam's review, but after addressing comments nothing jumps out as problematic to me

AdamFipke (changes requested):
meant to request changes my bad

AdamFipke (approved):
tested it with the new search, looks to work good!

PR #463, opened 2026-01-17
Added nullable createDateColumn to appropriate entities

### LINKED ISSUE #453: Add createDateColumn to more entities (nullable)
I just realised we don't keep track on when a user is created. This seems like an important metric to track.

Could also add it to queues, question types, semesters, courses (though that could be handled by #451) and possibly more.

It needs to be nullable I guess. Since we need it to be compatible with old data.

It's a small issue. Don't forget to run migrations and update test snapshots!

mahigangal:
# Description

Added nullable createdAt column to the following entities, after further discussion with Adam: 
- user_model: track when a user account is created.
- course_model: track when a course is created for admin analytics.
- queue_model: track when queues are created. 
- question_type_model: track when question types are created.
- semester_model: track when semesters are created.
- organization_model: track when organizations are created.
- user_course_model: track when a user enrolls in a course.
- queue_invite_model: track when a queue invite is created.

I generated the new migrations using `yarn schema:drop && yarn migration:run && yarn typeorm migration:generate`, and a new migration file (1768954613282-remove-created-at.ts) has been added. 

The test snapshots have been updated depending on the new createDateColumn added. Updated integration snapshots to use `expect.any(String)` or `expect.any(Date)` for createdAt fields to avoid timestamp drift across runs.

#### Motivation/context:
The system previously did not track user creation time; this is needed for analytics and admin metrics. Additionally, I also added createdAt to entities where it is applicable. Columns are nullable to ensure compatibility with existing data.

Closes #453 

## Type of change

# How Has This Been Tested?

I checked the database to make sure the new column had been added in the mentioned entities. I updated the snapshots using `yarn test:update-unit-snapshots && yarn test:update-integration-snapshots` to make sure they match the latest changes. Then ran the unit tests and the integration tests.

# Checklist:


packages/server/src/organization/__snapshots__/organization.service.spec.ts.snap
```diff
@@ -32,10 +32,10 @@ exports[`OrganizationService getCourses should return organization courses with
     "semester": {
       "color": "blue",
       "description": "Test Semester Description",
-      "endDate": 2022-12-31T00:00:00.000Z,
+      "endDate": 2022-12-30T08:00:00.000Z,
       "id": 1,
       "name": "Test Semester",
-      "startDate": 2020-09-01T00:00:00.000Z,
```
AdamFipke: by the way you should just be able to change all of the "startDate" and "endDate" things in this file back to the way they were and the tests will pass since your changes should not have affected these.

packages/server/src/queue/queue.service.spec.ts
```diff
@@ -201,7 +201,33 @@ describe('QueueService', () => {
         ours.creatorId,
         Role.STUDENT,
       );
-      expect(lqr).toMatchSnapshot();
+      expect(lqr).toMatchSnapshot({
+        questions: [
+          {
+            questionTypes: [
+              {
+                createdAt: expect.any(Date),
```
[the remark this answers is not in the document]
mahigangal: i have followed this way now for adding the createdAt column. Removed the pattern matchers and updated tests accordingly. A couple of places, tests were not using the factory for creating the entity, so I added static values for the createdAt date.

packages/server/test/__snapshots__/organization.integration.ts.snap
```diff
@@ -58,10 +61,10 @@ exports[`Organization Integration GET /organization/:oid/get_courses{/:page} sho
     "semester": {
       "color": "blue",
       "description": "Test Semester Description",
-      "endDate": "2022-12-31T00:00:00.000Z",
+      "endDate": "2022-12-30T08:00:00.000Z",
       "id": 1,
       "name": "Test Semester",
-      "startDate": "2020-09-01T00:00:00.000Z",
```
AdamFipke: this startDate and endDate should also be changed back to what it was before

packages/server/test/__snapshots__/course.integration.ts.snap
```diff
@@ -183,6 +191,7 @@ exports[`Course Integration POST /courses/:id/create_queue/:room correctly propa
   "allowQuestions": true,
   "config": {},
   "courseId": Any<Number>,
+  "createdAt": 2026-02-02T08:00:23.918Z,
```
AdamFipke: This date doesn't look statically generated (and is why a few of the tests are failing). Since you have added a static value for createdAt to QueueFactory, it likely means that these few tests in particular create their queue entities using the normal QueueModel.create({...}) syntax.

Seeing as how this is for the create_queue endpoint, this makes sense (all of our main services and endpoints create entities like normal rather than use factories. Factories are only used for test or seed data). So, since you can't modify the create_queue endpoint directly to make createdAt static, for specifically these few tests with create_queue you will want to add a pattern matcher that matches createdAt to Any Date, like you have done before previously. You could also run an update query after creating the queue to update createdAt to something static. Either way works.

packages/server/src/insights/insights.service.spec.ts
```diff
@@ -509,7 +509,9 @@ describe('InsightsService', () => {
       ],
     })) as ChartOutputType;
 
-    expect(res.data).toEqual(expected);
+    expect(res.data).toMatchSnapshot(
+      res.data.map(() => ({ date: expect.any(Number) })),
```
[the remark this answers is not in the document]
mahigangal: Any changes to insights have been reverted, as adding the createdAt column doesn't affect it in any capacity. 

I had added the expect.any(Number) pattern matcher, as they were failing before due to timezone‑sensitivity.

PR #479, opened 2026-02-22
Full Delete Course 

### LINKED ISSUE #450: Full Delete Course
We have an Archive Course button, but that course still exists and can still be accessed (just kinda hidden away).

I think some courses were probably accidentally made, so having a way to perma-delete courses would be nice.

Most of it can just be handled by ON DELETE CASCADE for stuff like user courses and questions and stuff. Just make sure to warn the user that "this will permanently delete everything to do with this course, including questions asked etc. Do this only if you accidentally created this course. Perhaps consider Archiving it instead.".

Make sure it deletes all associated chatbot documents and questions too in the chatbot database

frasermuller:
# Description

Closes #450 

This PR adds a new feature of being able to fully delete a course and everything related to it. Below are my changes from my notes:

1. Wrote a query in beekeeper studio  to find all the tables that have a connection to the course_model table so we can change all the tables that have NO ACTION -> CASCADE because they will block course deletion:
table_name

2. Created a new migration file to change the 10 tables ABOVE with NO ACTION to CASCADE.
confirming changes worked on beekeeper studio:
alert_model

3. Made backend endpoint that finds the course and calls .remove(). and the CASCADE should handle the rest. NOTE: only organization level admins can fully delete courses. so if this should be changed to professors aswell or some other role let me know.

4. Made frontend API function that calls the backend endpoint. nothing new to note here.

5. Made the delete button component, it is very similar to the archive course, but with a warning message/confirmation before calling the api function. Also redirects back to /courses after.

6. Added the delete button to the edit course page, but only appears for admins.

SCREENSHOTS OF NEW FEATURE:

Screenshot 2026-02-21 at 3 39 07 PM

© Are you sure you mant so permanently delete this counse

## Type of change

# How Has This Been Tested?

**TEST 1:**

After deleting COSC 310 and going back into beekeeper studio I ran

SELECT id, name FROM course_model;

Which shows 
Test Course
Showing it was deleted

**TEST 2:**

SELECT * FROM course_model WHERE id = 50;
SELECT * FROM queue_model WHERE "courseId" = 50;
SELECT * FROM user_course_model WHERE "courseId" = 50;

Shows nothing returned, which cosc 310’s id was 50, so the CASCADE worked since the other tables with the referenced coursed were deleted.

# Checklist:


packages/server/src/organization/organization.controller.ts
```diff
@@ -641,6 +641,30 @@ export class OrganizationController {
       });
   }
 
+  // For permanently deleting a course
+  @Delete(':oid/delete_course/:cid')
+  @UseGuards(JwtAuthGuard, OrganizationRolesGuard, EmailVerifiedGuard)
+  @OrgRoles(OrganizationRole.ADMIN)
+  async deleteCourse(
```
AdamFipke: in this instance, I would probably want to add some tests for this. Stuff like
- Make sure only org admins can call the endpoint
- Try making a course with a bunch of random connected entities (queues, events, user_course, etc.) and then call this endpoint to make sure that all of those entities are deleted, but it probably doesn't need to be too fancy. In an ideal world, I would also want some way of making sure that no unintended database entities were affected (for example, if the user entities were deleted due to some ON DELETE CASCADE messup in another entity), but that would require a fair bit of work (plus there's a LOT of other tests should also check this).

frasermuller:
I added 2 integration tests:

403 for non-admins (only org admins can delete)
200 for successful deletion by org admin

I wasn't able to test the CASCADE behavior (verifying that queues, user_courses, etc. get deleted too). The test DB uses: synchronize: true, which creates tables from entity decorators, but it doesn't run migrations. The CASCADE constraints are in my migration file but not in the entity decorators, so the test DB doesn't have them.

Tested manually and the CASCADE works, but if there's a better approach for testing this, let me know.

Also had to move it up in the file, cause CI was timing out.

PR #492, opened 2026-03-03
Tablet Navbar issues fixed

### LINKED ISSUE #418: Better Tablet Navbar
Right now, the hamburger menu will show itself only at the `md` breakpoint (like 400 something px)

This poses a problem, where even browser widths as high as 1200px will start to show a horizontal scrollbar because all of the nav elements cannot fit horizontally (though I should note that this is only really bad for professors since they have the Course Settings and Insights nav elements ontop of everything else)

Image

This horizontal scroll also messes up the profile dropdown menu (since it goes off of viewport width, unfortunately)

Image

---

### Solution

Make it so the navbar will go into the mobile "hamburger" menu mode whenever all nav elements cannot fit. Note that this width is very dynamic, since there can be a different number of nav elements depending on whether the user is a student or professor or depending on what course features the course has enabled 

ALSO, we could add an intermediary step (between "hamburger" menu mode and the full desktop navbar) where the paddings of the nav elements are reduced in an effort to fit more nav elements in without needing to fully go to the hamburger menu.

mahigangal:
# Description

This PR fixes the responsive course header behavior for issue #418. Previously, the header switched to the drawer menu only at a fixed breakpoint (768px). On tablet/smaller desktop widths, role/course-dependent nav items (especially professor links like Course Settings and Insights) could overflow, causing a horizontal scrollbar and dropdown placement issues.

The header now uses a 3-mode system:
- desktop
- compact (reduced top-level nav padding)
- drawer (hamburger)

Mode selection is based on measured rendered width (not fixed nav-item-count estimates), so it helps with student vs professor nav differences, enabled course features, zoom/larger font size scenarios.

Additionally, NavigationMenu now supports an optional showViewport prop so hidden measurement navbars can skip rendering an extra viewport element.

No dependencies, env changes, DB changes, or install steps are required.

Closes #418 

## Screenshots

1. Professor course at ~1024px
Screenshot 2026-03-05 at 3 56 56 PM

2. Professor course at ~1280px
Screenshot 2026-03-05 at 3 57 53 PM

3. Student course at ~1024px
Screenshot 2026-03-05 at 3 59 41 PM

4. Narrow tablet width ~820px (student or professor)
Screenshot 2026-03-05 at 4 01 07 PM

5. Queues dropdown near a threshold width (~820px)
Screenshot 2026-03-05 at 4 01 49 PM

6. Profile dropdown near a threshold width (~1024px)
Screenshot 2026-03-05 at 4 02 45 PM

All screenshots show no horizontal scrollbar and correct nav mode for width/role.

## Type of change

# How Has This Been Tested?

Manually tested on course pages with different roles and feature sets, focusing on widths where overflow previously occurred.

Repro steps:

1. Open a student course home page and a professor course home page.
2. Resize viewport across tablet/smaller desktop widths (for example around 820px, 1024px, 1280px).
3. Verify the header switches between desktop, compact, and drawer based on fit.
4. Confirm there is no horizontal scrollbar.
5. Open the profile dropdown and queue dropdown near threshold widths and verify positioning/interaction.
7. Repeat checks at browser zoom levels (125%, 150%) and confirm nav switches mode instead of overflowing.

Zoom testing note: I could only test zoom reliably at desktop width. At 100% zoom, all nav elements are shown in desktop mode; after zooming in enough, the navbar shifts to drawer mode without introducing a horizontal scrollbar.  I couldn’t reliably test zoom at other widths in DevTools device mode.

# Checklist:

### DIFF
--- a/packages/frontend/app/components/HeaderBar.tsx
+++ b/packages/frontend/app/components/HeaderBar.tsx
@@ -188,21 +188,12 @@ const NavBar = ({
   const compactTopLevelClass = isCompactDesktop ? '!pl-1.5 !pr-2' : ''
   const drawerTopLevelItemClass = showDrawerPresentation ? 'w-full' : ''
   const horizontalInsetClass = !showDrawerPresentation ? 'md:pl-8' : ''
-  const horizontalInsetImportantClass = !showDrawerPresentation
-    ? '!md:pl-8'
-    : ''
   const selectedNavItemClass =
     'bg-zinc-300/80 hover:bg-zinc-300/80 focus:bg-zinc-300/80'
   const sortedQueues = useMemo(() => {
     if (!course?.queues) return []
     return sortQueues(course.queues)
   }, [course?.queues])
-  const queueDropdownListClass = showDrawerPresentation
-    ? `grid w-full gap-1 p-4 ${sortedQueues.length > 6 ? 'grid-cols-2' : 'grid-cols-1'}`
-    : `grid gap-1 p-4 md:grid-cols-2 lg:w-[600px] lg:gap-2 ${sortedQueues.length > 6 ? 'w-[95vw] grid-cols-2' : 'w-[60vw]'}`
-  const emptyQueueStateClass = showDrawerPresentation
-    ? 'w-full p-4 text-center text-sm text-gray-500'
-    : `w-[60vw] p-4 text-center text-sm text-gray-500 ${role === Role.PROFESSOR ? 'lg:w-[600px]' : 'lg:w-[400px]'}`
 
   // This is to move the "Queues" and "Profile" submenu to show on the left or right side (there is only one component, so it needs to be moved around like this)
   const setNavigationSubMenuRightSide = useCallback(() => {
@@ -317,7 +308,22 @@ const NavBar = ({
                       <NavigationMenuContent>
                         {/* On mobile, if there are more than 6 queues, put the queue list into two columns */}
                         {sortedQueues.length > 0 ? (
-                          <ul className={queueDropdownListClass}>
+                          <ul
+                            className={cn(
+                              'grid gap-1 p-4',
+                              showDrawerPresentation
+                                ? 'w-full'
+                                : 'md:grid-cols-2 lg:w-[600px] lg:gap-2',
+                              sortedQueues.length > 6
+                                ? 'grid-cols-2'
+                                : showDrawerPresentation
+                                  ? 'grid-cols-1'
+                                  : 'w-[60vw]',
+                              !showDrawerPresentation &&
+                                sortedQueues.length > 6 &&
+                                'w-[95vw]',
+                            )}
+                          >
                             {sortedQueues.map((queue) => (
                               <ListItem
                                 key={queue.id}
@@ -336,7 +342,16 @@ const NavBar = ({
                             ))}
                           </ul>
                         ) : (
-                          <div className={emptyQueueStateClass}>
+                          <div
+                            className={cn(
+                              'p-4 text-center text-sm text-gray-500',
+                              showDrawerPresentation
+                                ? 'w-full'
+                                : role === Role.PROFESSOR
+                                  ? 'w-[60vw] lg:w-[600px]'
+                                  : 'w-[60vw] lg:w-[400px]',
+                            )}
+                          >
                             <p>There are no queues in this course</p>
                             {role === Role.PROFESSOR && (
                               <p>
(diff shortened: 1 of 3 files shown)


packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -595,10 +643,18 @@ const NavBar = ({
...
  */
+
+type NavMode = 'desktop' | 'compact' | 'drawer'
+
 const HeaderBar: React.FC = () => {
   const { userInfo } = useUserInfo()
   const [isDrawerOpen, setIsDrawerOpen] = useState(false)
-  const isDesktop = useMediaQuery('(min-width: 768px)')
+  const [navMode, setNavMode] = useState<NavMode>('drawer')
+  const isPhone = useMediaQuery('(max-width: 640px)')
```
AdamFipke: you *might* want to leave this as 768px instead of 640px since basically the entire site uses `md:` as the breakpoint between mobile and desktop for CSS, so in the case where the browser width is at 700px it might show the desktop navbar but with mobile styles.

packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -633,104 +689,192 @@ const HeaderBar: React.FC = () => {
...
+    nodes.forEach((node) => observer.observe(node))
+    updateNavMode()
+
+    return () => observer.disconnect()
+  }, [updateNavMode, pathname, course, userInfo])
+
+  return (
+    <div ref={availableWidthRef} className="relative w-full">
+      <div className="pointer-events-none invisible absolute left-0 top-0 -z-10 h-0 overflow-hidden">
+        <div ref={regularMeasureRef} className="w-max">
```
AdamFipke: Okay I get this now. 

We're rendering 3 navbar components all the time.

2 of these navbar components are invisible and literally only exist so their width can be measured. 1 navbar has the condensed classnames, and the other the standard desktop navbar.

The final navbar is the real one.

It's kinda creative, but I think rendering 3 navbars might have some performance cost considerations that might not be worth itt so I might wait on merging this in case someone thinks of a better solution.

AdamFipke (changes requested):
See comments. 

Rendering 3 navbars and using 2 of them just for measuring the width of the desktop navbar and condensed navbar to determine what mode the true navbar should be in is a creative approach, but I'm unsure if it'll be worth the performance cost. I will leave this to you to test more. Maybe one of us can come up with another way that's better for performance if it turns out to be noticeably hurting performance. 

Check out React Dev Tools and see if there's lots of unnecessary re-renders. Also try using chrome dev tools to throttle performance and see if it stills feels responsive.

Image

packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -633,104 +689,192 @@ const HeaderBar: React.FC = () => {
...
+      setNavMode('compact')
+      return
+    }
+
+    setNavMode('drawer')
+  }, [isPhone])
+
+  useLayoutEffect(() => {
+    updateNavMode()
+  }, [updateNavMode, pathname, course, userInfo])
```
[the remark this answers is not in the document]
AdamFipke: Huh, so I went and manually tested it and maybe this is less performance intensive than I thought? Like it really does only trigger the re-render at the breakpoints. Moving the window width a lot doesn't seem to cause any re-renders. So maybe debouncing isn't really that necessary.
image

Although I guess this only shows the visible navbar. The invisible ones would also be re-rendering a bunch.

AdamFipke (commented):
just one small comment, should be good to merge after

AdamFipke (approved):

packages/frontend/app/components/HeaderBar.tsx
```diff
@@ -350,45 +400,47 @@ const NavBar = ({
...
                         Schedule
                       </Link>
                     </NavigationMenuItem>
                   )}
                   {(role === Role.TA || role === Role.PROFESSOR) && (
-                    <NavigationMenuItem
-                      className={
-                        isACourseSettingsPage
-                          ? // the hover:border-none is because the inner link has a hover effect that adds another border
-                            'md:border-helpmeblue bg-zinc-300/80 md:border-b-2 md:bg-white md:hover:border-none md:focus:border-none'
```
AdamFipke, line 370: I tried `ctrl+f`ing "border-helpmeblue" and I see that the original desktop styles have actually been completely removed (presumably got misinterpreted for some reason after my first round of feedback). I was expecting that the changes were at least tested. I'll look into fixing it, but I might just revert the changes done by this PR

AdamFipke: Yeah, the more I look at it now, the more random broken changes I see. It's not worth my time to try to salvage this.

PR #507, opened 2026-04-03
Improve uploads security

### LINKED ISSUE #499: Improve Get/Upload Image security
- Change Upload to all use ParseFilePipeBuilder (or use MaxFileSizeValidator and FileTypeValidator ) probably https://docs.nestjs.com/techniques/file-upload
- Validate who is allowed to get the image (is the user an admin? if so allow everything. Otherwise, only allow you to retrieve PFPs from those in the same course I think)
- Change GET image endpoints to just use the given orgId or userId to get the photoURL rather than use a user-supplied photoURL.

Add tests:
- For all upload endpoints and get file endpoints
- Test when user tries to access images they shouldn't be able to access (course-permissions)
- Test when user tries to access files they shouldn't be able to access (like passing `../../.env` into photoURL or some illegal characters. Could use the big list of naughty strings https://github.com/minimaxir/big-list-of-naughty-strings)
- Upload files too large, too small
- Upload files of incorrect file types. Upload files with correct file types but incorrect mime types (and vise-versa).

AdamFipke:
# Description

It's A LOT smaller than what it looks like. Like over 1k lines are just from tests or linter things

- Adds a series of permission checks to getProfilePicture endpoint so that you can't just get any user's profile picture as long as you are logged in. Now actually checks to see whether you *should* be able to see the pfp. Though I should note that this doesn't work for user's external photo urls (e.g. if it was retrieved from Login with Google).
- Makes it so users no longer feed in the photoUrl in order to get the organization's logo/banner. Instead it just uses the org ID and retrieves the photourl from the DB
- Similar case with user.photoUrl, except the requester still sends the photoUrl inside the endpoint url (for browser caching reasons in case the user changes their pfp). In the future, when we start storing the pfp inside the DB we can just store an identifier for the browser caching issue (i.e. `get_pfp/${userId}/${a random string, maybe a uuid}` where said random string is stored in the DB). We don't actually use the photoUrl that the requester gives us, we still first get the user entity and then use user.photoUrl before putting that inside `fs.stat()`
- Changes the file upload decorators to use Nestjs's ParseFilePipeBuilder (so that magic numbers get checked). **IMPORTANT:** From now on, if you need to get the user upload, please use:
```ts
@UploadedFile(
      new ParseFilePipeBuilder()
        .addFileTypeValidator({
          // Note that nestjs filetypevalidator comes with mime type and magic number validation build in
          fileType: 'jpg|jpeg|png|gif|avif|webp',
        })
        .addMaxSizeValidator({
          maxSize: 5 * 1024 * 1024, // 5MB limit per file
        })
        .build(),
    )
    file: Express.Multer.File,
```

Note: Also added `--experimental-vm-modules` in the node options for testing since nestjs FileTypeValidator uses eval('import('file-type')') which doesn't work in jest, causing the magic byte validation to not work.

Closes #499 

## Type of change

# How Has This Been Tested?

Added a boatload of tests for get_pfp. Manually tested in a bunch of areas to make sure it looks okay.

# Checklist:


packages/server/src/mail/weekly-summary.builder.ts
```diff
@@ -89,15 +89,22 @@ export interface CourseStatsData {
   suggestArchive: boolean;
 }
 function validateHtml(html: string): void {
-  const originalTags = html.match(/<(?!\/)(?!br|hr|img|input|meta|link|area|base|col|embed|source|track|wbr)([a-zA-Z][a-zA-Z0-9]*)[^>]*(?<!\/)>/g) || [];
```
AdamFipke, line 92: I don't think I modified this file at all, I think this was just the linter. Same goes for all of the weekly summary stuff actually. Idk what happened here exactly (though iirc, I was messing with node import settings so that might've triggered it)

PR #561, opened 2026-06-09
Add chatbot agent groups

Whiteknight07:
# Description

Adds chatbot agent groups. One visible parent course (LANTERN) is linked to several hidden "agent" courses (Analyst, Communicator, Strategist, Thrive). Each agent is a normal course under the hood, so it gets its own chatbot prompt, documents, settings, and chat history. Students only enroll in the parent course and pick an agent from a dropdown in the chatbot; their questions are routed to that agent's course.

How it works:
- `SuperCourseModel` gains a `purpose` enum (`course_clone_group` | `chatbot_agent_group`) and its `courses` relation is now ManyToMany, so a course can be in a clone group and an agent group at the same time. A migration backfills existing super courses as `course_clone_group`.
- Courses gain `chatbotAgentName`, `chatbotAgentDescription`, and `chatbotAgentOrder`. A course with an agent name is an agent; the course in the group without one is the parent.
- New endpoint `GET /chatbot/course/:courseId/agents` returns the agents for a course (archived agents are hidden unless the parent course is also archived).
- The HelpMe-course bypass guard is renamed to `CourseRolesConditionalBypassGuard` and extended: members of the parent course may use the chatbot endpoints of agent courses in the same group, with their parent-course role. Everyone else still gets a 404.
- Cloning a course that belongs to an agent group clones the whole group (all agent courses, their agent fields, and their chatbot data) into a new agent group. Clone groups and agent groups stay separate.
- `yarn seed:chatbot-agent-group` sets up the LANTERN courses on production (expects the `UBC` organization and `2026S Both Terms` semester to exist, enrolls no one). See `docs/chatbot_agent_groups.md`.

## Type of change

- [x] This change requires a documentation update (added `docs/chatbot_agent_groups.md`)
- [x] This change requires a database query to update old data on production: handled by the included migration (backfills `purpose` and the new join table); after deploy, run `yarn seed:chatbot-agent-group` once to create the LANTERN courses.

# How Has This Been Tested?

- [x] Integration tests for the agents endpoint (empty list, sorted agents, archived-agent hiding, access blocked for non-members)
- [x] Integration tests for the bypass guard (parent member can ask in agent course, outsider gets 404)
- [x] Integration tests for cloning courses with agent groups (from the parent course and from an agent course)
- [x] Existing course clone unit tests updated for the `purpose` column
- [x] Manually tested locally: seeded the group, logged in as a student in the parent course, asked questions to each agent, and verified separate chat histories
- [x] Screenshots of the agent selector (mobile and desktop) — see Screenshots below

# Checklist:

- [x] I have performed a code review of my own code

<details>
<summary>Screenshots</summary>

### 1. Mobile
Mobile agent selector

### 2. Laptop
Laptop agent selector

### 3. Laptop (dropdown)
Laptop agent dropdown open

</details>


packages/server/migration/1780375600000-chatbot-agent-course-groups.ts
```diff
@@ -0,0 +1,37 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class chatbotAgentCourseGroups1780375600000
+  implements MigrationInterface
+{
+  name = 'chatbotAgentCourseGroups1780375600000';
```
AdamFipke: we have a command for generating migrations, please use that. I think the command is mentioned inside DEVELOPING.md

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -211,6 +212,38 @@ export class ChatbotController {
...
+    @Param('courseId', ParseIntPipe) courseId: number,
+  ): Promise<ChatbotAgentCourse[]> {
+    const course = await CourseModel.findOne({
+      where: { id: courseId },
+      relations: { superCourse: { courses: true } },
+    });
+
+    if (course?.superCourse?.purpose !== 'chatbot_agent_group') {
+      return [];
+    }
```
AdamFipke: Can't you change this so it's something like:
```suggestion
    const supercourse = await SuperCourseModel.findOne({
      where: { courseId: courseId, purpose: 'chatbot_agent_group' },
      relations: {  courses: true  },
    });

    if (!supercourse) {
      return [];
    }
```

One less join this way. Plus if the query doesn't find anything, no data needs to be deserialized so it's a little bit faster that way too. (note that this might not be a complete suggestion so probably don't commit it directly)

packages/server/src/course/super-course.entity.ts
```diff
@@ -17,6 +17,9 @@ export class SuperCourseModel extends BaseEntity {
   @Column('text')
   name: string;
 
+  @Column('text', { nullable: true })
+  purpose?: string;
```
AdamFipke: adding a 'purpose' column is a pretty smart way of doing this i think. Though you will also want to adjust the course clone feature (since it also uses SuperCourse) so that it has its own purpose (something like "connection_between_clones" or something else if you can think of something better). Also, the way clone course creates a Super Course is that it will find any pre-existing SuperCourse whose courseId is the one that's being cloned and then *re-use that SuperCourse* if one doesn't yet exist. You will probably just want to slightly adjust the query so that it only looks for `purpose: 'connection_between_clones'` so it doesn't use the supercourse for multi-agents. Ideally, integration tests are also added to make sure that this check works.

You should probably also make this an enum rather than a string (and put the enum inside common/index.ts, similar to mail-services.entity.ts) and then re-generate the migrations.

--- 

**EDIT**: I just realised that SuperCourse is OneToMany not ManyToMany. Meaning that what I suggested above wouldn't quite work (since each course is only allowed one supercourse). I do also realise there's an important edge case here of "what happens when you clone a course with agent courses?". There's at least a couple approaches from here:
1. Disallow courses with a supercourse of 'chatbot_agent_group' from being cloned.
2. Allow courses with a supercourse of  'chatbot_agent_group' to be cloned. Need to change SuperCourse to be ManyToMany with course (easy part). Two options from here:
   1. Only allow "main" course to be cloned?? This I think would unfortunately be a pain to implement.
   2. When cloning one agent course, its 'chatbot_agent_group' supercourse is cloned and all other agent courses are also cloned. Significantly easier to implement. I guess we also would create several 'connection_between_clone' supercourses as well, each connecting the cloned agent course with its original agent course if that makes sense (don't forget to clone over `chatbotAgentName`, `chatbotAgentDescription`, and `chatbotAgentOrder`. I believe the only extra logic for this would need to be added to the cloneCourse() method (might be called something slightly different).

Option 2.2 is probably gonna be best. I see it being only a little bit more work than option 1 since you'll want to make integration tests for it anyway (since cloning a course with agent courses is going to be *very* rare so we'll rarely be manually testing it). Plus it future-proofs it in case the professor goes "hey that worked great this semester, let's do it again next semester"

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -211,6 +212,38 @@ export class ChatbotController {
     );
   }
 
+  @Get('course/:courseId/agents')
+  @UseGuards(CourseRolesGuard)
+  @Roles(Role.STUDENT, Role.TA, Role.PROFESSOR)
+  async getChatbotAgents(
```
AdamFipke, line 220: maybe add a handful of integration tests for this endpoint?

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -290,8 +320,58 @@ const Chatbot: React.FC<ChatbotProps> = ({
...
+                      }))}
+                      onChange={(newValue) => {
+                        if (newValue !== selectedAgentCourseId) {
+                          setTempAgentCourseId(newValue)
+                        }
+                      }}
+                    />
+                  </Popconfirm>
+                ) : (
+                  <Select
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]: do you think you could add a screenshot or two of how this looks to the PR description? On both mobile and desktop?

packages/server/src/guards/course-roles-conditional-bypass.guard.ts
```diff
@@ -33,6 +34,30 @@ export class CourseRolesBypassHelpMeCourseGuard extends RolesGuard {
         role: Role.STUDENT,
       });
     }
+
+    // Agent courses are hidden from students, but students enrolled in the
+    // non-agent parent course should still be able to ask those agent chatbots.
```
AdamFipke, line 39: This guard should probably be renamed as it is no longer only the "HelpMe Course" Bypass. Something like `CourseRolesBypassHelpMeCourseGuard` -> `CourseRolesConditionalChatbotBypass` (rename file too). Maybe if you can think of something better, feel free to mention it. Like `CourseRolesBypassHelpMeCourseAndAgentCourses` is probably most descriptive but a little too long haha.

packages/server/src/seed/seed-chatbot-agent-group.command.ts
```diff
@@ -0,0 +1,161 @@
...
+import { UserCourseModel } from 'profile/user-course.entity';
+import { UserModel } from 'profile/user.entity';
+import { SemesterModel } from 'semester/semester.entity';
+
+const agents = [
+  {
+    name: 'Analyst',
+    description:
+      'Research foundations, statistics, research methods, terminology, philosophy of science, and critical appraisal.',
+  },
```
AdamFipke, line 18: Profs won't have any way of adjusting this (and it'd be a bit of a pain for us to update it if they have a request since we'd either need to query the DB directly or make another command or migration for it). Maybe that's fine but I just thought i'd mention it

packages/server/src/chatbot/chatbot.controller.ts
```diff
@@ -211,6 +212,38 @@ export class ChatbotController {
...
+  ): Promise<ChatbotAgentCourse[]> {
+    const course = await CourseModel.findOne({
+      where: { id: courseId },
+      relations: { superCourse: { courses: true } },
+    });
+
+    if (course?.superCourse?.purpose !== 'chatbot_agent_group') {
+      return [];
+    }
+
```
AdamFipke, line 231: Also random thought about archived "agent" courses: We probably would want to hide them from the UI if the main course is non-archived (maybe the prof found one of the agents unnecessary, or maybe they needed to fix there being duplicate agents for some reason). But maybe don't hide the archived agents from the UI if the main course is *also* archived (since maybe the prof is looking back at an archived course to showcase to someone). 

There would be different ways about implementing this (either put the logic on the frontend or backend, or even right in the query itself). The plus side about doing it on the backend is it would be easier to make tests to make sure it's working and not breaking anything.

packages/server/src/guards/course-roles-helpme-bypass.guard.ts
```diff
@@ -33,6 +34,30 @@ export class CourseRolesBypassHelpMeCourseGuard extends RolesGuard {
...
       });
     }
+
+    // Agent courses are hidden from students, but students enrolled in the
+    // non-agent parent course should still be able to ask those agent chatbots.
+    if (!user?.courses?.find((c) => Number(c.courseId) === Number(courseId))) {
+      const agentCourse = await CourseModel.findOne({
+        where: { id: Number(courseId) },
+        relations: { superCourse: { courses: true } },
+      });
```
AdamFipke:
```suggestion
      const superCourse = await SuperCourseModel.findOne({
        where: { courseId: courseId, purpose: 'chatbot_agent_group' },
        relations: { courses: true  },
      });
```
Like the other comment, this will probably be a little more efficient (less join, less serialization). Syntax might be slightly incorrect so don't commit directly. Also code below can be simplified a bit due to this

packages/server/src/seed/seed-chatbot-agent-group.command.ts
```diff
@@ -0,0 +1,161 @@
...
+
+@Injectable()
+export class SeedChatbotAgentGroupCommand {
+  @Command({
+    command: 'seed:chatbot-agent-group',
+    describe: 'creates the LANTERN chatbot agent group demo courses',
+  })
+  async createLanternAgentGroup(): Promise<void> {
+    const organization = await OrganizationModel.findOneOrFail({
+      where: { name: 'UBCO' },
```
AdamFipke:
```suggestion
      where: { name: 'UBC' },
```
On prod it's UBC

packages/server/src/seed/seed-chatbot-agent-group.command.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+    const semester =
+      (await SemesterModel.findOne({ where: { name: 'Test Semester' } })) ??
```
AdamFipke:
```suggestion
      (await SemesterModel.findOne({ where: { name: '2026S Both Terms' } })) ??
```
Should probably put a real semester there that matches prod. Though also idk when this course is taking place exactly (or if at all, in which case it should use the "Test Course" semester)

I have also promoted you to Admin on HelpMe so you can find the full list of semesters here: https://coursehelp.ubc.ca/organization/settings (scroll down)

packages/server/src/seed/seed-chatbot-agent-group.command.ts
```diff
@@ -0,0 +1,161 @@
...
+      (await SemesterModel.findOne({
+        where: { organizationId: organization.id },
+      }));
+    const superCourse = await this.findOrCreateSuperCourse(organization.id);
+    const parentCourse = await this.findOrCreateCourse('LANTERN', semester?.id);
```
AdamFipke: should the main course name to just be "LANTERN"? Is there like a course code like "LANT 123" or something? Also some other parts might need to be updated since there is at least one hardcoded check for it

packages/server/src/seed/seed-chatbot-agent-group.command.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    await this.attachCourseToGroup(parentCourse, superCourse, organization.id);
+    await this.enrollStudentOne(parentCourse);
+
+    for (const [index, agent] of agents.entries()) {
+      const course = await this.findOrCreateCourse(agent.name, semester?.id);
```
AdamFipke:
```suggestion
      const course = await this.findOrCreateCourse("some prefix like LANT 123" + agent.name, semester?.id);
```
I think we should prefix the agent name with some proper prefix so that way we don't end up with a course named "Strategist" or "Communicator" etc.

packages/server/src/seed/seed-chatbot-agent-group.command.ts
AdamFipke: Also, does the course already exist on HelpMe? Since in that case this course name should match that course.

packages/server/src/seed/seed-chatbot-agent-group.command.ts
```diff
@@ -0,0 +1,161 @@
...
+      return existing;
+    }
+
+    return CourseModel.create({
+      name,
+      semesterId,
+      timezone: 'America/Los_Angeles',
+      sectionGroupName: '001',
+      enabled: true,
+    }).save();
```
AdamFipke: this should match the same logic as createCourse in organization.controller.ts.


```suggestion
    return CourseModel.create({
      name,
      semesterId,
      timezone: 'America/Los_Angeles',
      sectionGroupName: '001', // You should double-check to make sure the LANTERN course is actually section 001 on canvas/workday or whatever (can also just leave as empty string)
      zoomLink: '',
      enabled: true,
      courseInviteCode: crypto.randomBytes(6).toString('hex'), // need random courseInviteCode
    }).save();
```
I'd just double check the sectionGroupName, and also make sure courseInviteCode is set.

~~Also, you're missing the stuff for creating an OrganizationCourseModel (!!) and the CourseSettingsModel for these new courses.~~ nvm I see that that logic is placed inside attachCourseToGroup. Also I guess for the agent courses you could have the "queues" and "anytime question" features be disabled to make it simpler for profs but that probably doesn't matter too much.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -105,16 +107,42 @@ const Chatbot: React.FC<ChatbotProps> = ({
   // used to temporarily store what question type the user is trying to change to
   const [tempChatbotQuestionType, setTempChatbotQuestionType] =
     useState<ChatbotQuestionType | null>(null)
+  const [agents, setAgents] = useState<ChatbotAgentCourse[]>([])
+  const [hasLoadedAgents, setHasLoadedAgents] = useState(false)
+  const [selectedAgentCourseId, setSelectedAgentCourseId] = useState<
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]: fyi because this state exists inside the Chatbot.tsx component (and not in the ChatbotProvider.tsx), this state will reset when the user switches between the chatbot on the main course page and the little "Chatbot" widget in the bottom corner rather than it being shared between the two components. In its current implementation, if the user selects an agent from the menu on the main course page, starts a chat, and then clicks onto another page, the selector will reset - meaning if the user then tries to continue the conversation, the wrong agent will be selected and they won't be able to select the agent they had before without it resetting the chat (since switching agents resets the chat).

To fix this, just lift up some of the state into ChatbotProvider.tsx. Probably the `agents` state and `selectedAgentCourseId` state, just be sure to add them to the existing useEffect to reset those states when switching courses. You may also want to add other states like `tempAgentCourseId`, but it's not obvious what it's for.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -146,7 +176,7 @@ const Chatbot: React.FC<ChatbotProps> = ({
 
     const data = {
       question:
-        chatbotQuestionType === 'System'
+        !selectedAgentCourseId && chatbotQuestionType === 'System'
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:171]: As a small tiny thing, I think you might want to add another `chatbotQuestionType` - "Agent". It might make some of the conditionals a little more explicit rather than checking for selectedAgentCourseId imo (e.g. "this part of the UI is only for Agent and Course chatbots"). 

Or... maybe not since I realised these values correspond with the current Segmented component lmao.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]: I'm assuming the flow is something like: When the Segmented is on "Course", an extra Select shows up somewhere (or better yet, a dropdown *within* the Segmented somehow) that lets the user choose an agent. When the Segmented is on "System", this extra Select is gone. 

 I should also note that something tells me you've probably run out of UI space around here, since the Chatbot is already really cramped on the top bar...

Image

... which is why if you can somehow make it so the "Course" part of the segmented shows a dropdown inside...

... or maybe when there's Agents available, you just replace the whole Segmented with a Select whose options include the list of agents *and* "General HelpMe System"

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -290,8 +320,58 @@ const Chatbot: React.FC<ChatbotProps> = ({
...
+                      value: agent.courseId,
+                      label: agent.agentName,
+                      title: agent.description,
+                    }))}
+                    onChange={(newValue) => {
+                      setSelectedAgentCourseId(newValue)
+                      resetChat()
+                    }}
+                  />
+                )
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:393]: as another thing, I realised that the "Convert to Anytime Question" feature might not work but I looked at the code for it and I think it should be fine actually.

I would still manually test it. Or hey, as another idea maybe the feature could be adjusted so that converted questions include a little bit of text in the answer text saying what agent was selected. That might be really easy to do once the `agents` state is stored inside ChatbotProvider, which can then easily be passed into ConvertChatbotQToAnytimeQModal on `/async_centre/page.tsx`.

packages/server/migration/1780375600000-chatbot-agent-course-groups.ts
AdamFipke: Also you might want to come back to this as a last step after resolving the other comments since you'll be doing some more DB changes

packages/server/src/seed/seed-chatbot-agent-group.command.ts
```diff
@@ -0,0 +1,161 @@
...
+        asyncQueueEnabled: true,
+        adsEnabled: true,
+        queueEnabled: true,
+      }).save();
+    }
+  }
+
+  private async enrollStudentOne(parentCourse: CourseModel): Promise<void> {
+    const student = await UserModel.findOne({
+      where: { email: 'studentOne@ubc.ca' },
```
AdamFipke: I think instead of enrolling a test user (that probably doesn't exist on production) into the new courses (and as a student i guess), maybe we could make it so it enrolls the LANTERN professor with the professor role? Obviously we wouldn't want to hard-code their email into our codebase, so maybe we could pass it as a param into the command... 

...Or maybe another option is to just enroll no-one and you or I can just add them in after the fact on production (you can do so by editing the course here https://coursehelp.ubc.ca/organization/courses )

packages/server/src/cli.ts
```diff
@@ -2,9 +2,19 @@ import { NestFactory } from '@nestjs/core';
 import { CommandModule, CommandService } from 'nestjs-command';
 import { AppModule } from './app.module';
 
-(async () => {
+async function bootstrap(): Promise<void> {
```
AdamFipke, line 5: @bhunt02 you might be interested in reviewing this stuff since I think you've used the nestjs command stuff more than me, but I think it probably looks fine

AdamFipke (changes requested):
Some comments. It's mostly all minor stuff, with the biggest items (all of which are still somewhat small) being:
- Clone course also uses SuperCourse, so some stuff will need to be tweaked otherwise cloning an Agent course (or the LANTERN course) might result in some bad data
- Making sure the UI for selecting the agents looks good
- Integration tests

I also noticed that the PR description doesn't follow the PULL_REQUEST_TEMPLATE for some reason?

packages/server/migration/1780375600000-chatbot-agent-course-groups.ts
Whiteknight07: gotcha my bad for it, ill have this sorted a the end

packages/server/src/chatbot/chatbot.controller.ts
Whiteknight07: Fixed!

packages/server/src/seed/seed-chatbot-agent-group.command.ts
Whiteknight07: Come to think of it, there should a UI for this so that profs can manage agents (change prompts for them, activate/deactivate agents, add agents etc and cloning too)

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (Whiteknight07) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]: For courses in a chatbot_agent_group, we replaced that Segmented with a single agent Select, per your suggestion since LANTERN students only need the four agents and the parent course isn’t its own chatbot persona. I left out “General HelpMe System” for the MVP; happy to add it as an extra option in the same dropdown or as a pill (as the agent dropdown serves a diff purpose as system) if you want that available inside LANTERN too. What do you think? Ive added screenshot in the PR desc

Comment (Whiteknight07) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]: Also the textbox outside of the view for phone view and people will have to scroll to get it in view when the page loads up. Might need to fix that as well

packages/server/src/course/super-course.entity.ts
Whiteknight07: Implemented option 2.2: cloning the parent (or any agent) clones the whole group into a new chatbot_agent_group, with course_clone_group super courses linking each clone to its original when 'associate' is checked. Covered by two integration tests.

packages/server/src/chatbot/chatbot.controller.ts
Whiteknight07: Added in chatbot.integration.ts

packages/server/src/guards/course-roles-conditional-bypass.guard.ts
Whiteknight07: Done

packages/server/src/chatbot/chatbot.controller.ts
Whiteknight07: implemented on the backend for this. Archived agents are hidden unless parent itself is archived.

packages/server/src/seed/seed-chatbot-agent-group.command.ts
Whiteknight07: Done

packages/server/src/seed/seed-chatbot-agent-group.command.ts
Whiteknight07: Did 2026S for now, also I wasnt able to see the full list of semesters.

packages/server/src/seed/seed-chatbot-agent-group.command.ts
Whiteknight07: LANTERN course does not exist on helpme. For the naming I have kept just LANTERN for now, but even I am not fully sure based on the proposal for it if this is the name the department wants to use.

packages/server/src/seed/seed-chatbot-agent-group.command.ts
Whiteknight07: Yup did, LANTERN Analyst... so on so forth

packages/server/src/seed/seed-chatbot-agent-group.command.ts
Whiteknight07: I set sectionGroupName to be 001 and we generate a random invite code ( the same way organization.controller.ts. Have not disabled queues and anytime questions yet (we might need them for lantern courses)

packages/server/src/seed/seed-chatbot-agent-group.command.ts
Whiteknight07: I went with the last option of enrolling nobody.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (Whiteknight07) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]: Lifted both into the provider and reset them in the existing course-change effect, so the selection survives switching between page chatbot and the widget. tempAgentCourseId is the pending value for the confirm popup, so it stayed local.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (Whiteknight07) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx:171]: Left as-is

packages/server/src/seed/seed-chatbot-agent-group.command.ts
AdamFipke: ah i see your cwl login has changed to your new ubc email (it's no longer the @student.ubc.ca one, which was the one I promoted since there was no other account), so that's what happened there. I have promoted your new ubc account to admin now so you should be able to seee  it now.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
Comment (AdamFipke) [reply packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]: > I left out “General HelpMe System” for the MVP; happy to add it as an extra option in the same dropdown or as a pill (as the agent dropdown serves a diff purpose as system) if you want that available inside LANTERN too. What do you think? 

I'm thinking we could still add "HelpMe System" to the Select options, but I think we should add a little divider to separate it from the other select options. I think even adding some kind of subtitle of `${course.name} Agents` above the 4 agent options would be best. The "option group" in the antd docs is what i mean https://ant.design/components/select#select-demo-optgroup 

> Also the textbox outside of the view for phone view and people will have to scroll to get it in view when the page loads up. Might need to fix that as well

I believe the chatbot height is a static height on mobile (it should be anyway, since otherwise it might cause some jarring movements as you chat with it). It's just that in this particular case with the phone height you set and that there's no queues or anything that they chatbot textbox appears just out of view. Normally, on a full mobile browser with a search bar and menu buttons etc. the viewport is actually pretty short so it'd be a little rough to fit it all. Moreover, as a little fun fact, if the prof disables all features except the chatbot, the course homepage becomes *just* a fullscreen chatbot. So yeah it's pretty much  just a combination of the specific browser height + course configurations that lead to what you saw there. There probably isn't something to fix there.

packages/server/src/course/super-course.entity.ts
```diff
@@ -17,6 +24,13 @@ export class SuperCourseModel extends BaseEntity {
   @Column('text')
   name: string;
 
+  @Column({
+    type: 'enum',
+    enum: SuperCoursePurpose,
+    nullable: true,
+  })
+  purpose?: SuperCoursePurpose;
```
AdamFipke:
```suggestion
  @Column({
    type: 'enum',
    enum: SuperCoursePurpose,
  })
  purpose: SuperCoursePurpose;
```
I'm feeling like purpose shouldn't be optional, that way you have a warning if you try to create a SuperCourse without a purpose. Technically, this will put the database in an illegal state when the changes are first pulled (since a non-nullable column will have null values), but that's what the migration is for!

You might even be able to get away without running the command to re-generate the migrations, and you probably only need to add `NOT NULL` to the query that adds the `purpose` column. But I would still suggest to try running it to make sure it works.

packages/server/migration/1781189195359-chatbot-agent-course-groups.ts
```diff
@@ -0,0 +1,77 @@
...
+      `CREATE INDEX "IDX_424f337188cbf825994e856c7c" ON "super_course_course_model" ("courseId") `,
+    );
+    await queryRunner.query(
+      `CREATE TYPE "public"."super_course_model_purpose_enum" AS ENUM('course_clone_group', 'chatbot_agent_group')`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "super_course_model" ADD "purpose" "public"."super_course_model_purpose_enum"`,
+    );
+    await queryRunner.query(
+      `UPDATE "super_course_model" SET "purpose" = 'course_clone_group' WHERE "purpose" IS NULL`,
```
AdamFipke: oh nice, i thought this wanna gonna be a gotcha but it looks like you remembered. Same goes for the other things like from changing the columns from OneToMany to ManyToMany (unless typeorm was able to detect that properly on it's own and automagically write the correct migration for it so that the relationship is kept)

packages/server/src/course/course.entity.ts
AdamFipke: wait, shouldn't this be deleted? We shouldn't have both a `superCourse` and `superCourses` attributes. Migration file might need to be re-generated, and some queries might need to be adjusted (anything that was using `course.superCourse`)

packages/server/src/course/course.service.ts
```diff
@@ -532,6 +533,352 @@ export class CourseService {
...
+        manager,
+        originalAgentCourse.id,
+        clonedAgentCourse.id,
+        cloneData,
+        chatToken,
+      );
+    }
+  }
+
   async cloneCourse(
```
AdamFipke, line 882: the git diff for this file was absolute wank but I think (?) I got the gist of it. I decided not to go deep into verifying the logic for it, but assuming you manually tested it to make sure it works, then it should be fine.

AdamFipke (changes requested):
thanks for implementing those changes. I have a couple more comments, the only major one being deleting the old `course.superCourse` attribute and verifying everything is still working. Also adding the "HelpMe System" under the Select under an option group (which I mentioned [here](https://github.com/ubco-db/helpme/pull/561#discussion_r3407537792)) would also be sweet. Then after that it should be good to merge probably.

There are a few other comments mentioned in the previous review that I just thought I'd consolidate here. I'm not sure if you want to just stick with what you've got here now or to implement any of these:
- UI so profs can manage agents (names, descriptions, order, add agents. Maybe some consolidated view for deactivating/activating agents, adjusting their prompts, seeing questions across all agents in a unified view, etc.). Maybe we can do this later after verifying that this is something they want.
- (tiny) LANTERN is the course name, but the department might want a different name
- Verify that the Convert to Anytime Question feature still works with different agents selected (the conversation *should* be put into a new anytime question in the main lantern course). Optionally, include what agent the user had selected as part of the questionText when doing the conversion ([should be pretty trivial to do](https://github.com/ubco-db/helpme/pull/561#discussion_r3393895351)).

Whiteknight07:
> Also adding the "HelpMe System" under the Select under an option group (which I mentioned [here](https://github.com/ubco-db/helpme/pull/561#discussion_r3407537792)) would also be sweet.

Added this! Here are the screenshots: 

<details>
  <summary>Screenshots</summary>
 
helpme-chatbot-agent-dropdown
helpme-chatbot-system-selected

 
 
</details>

packages/server/src/course/super-course.entity.ts
Whiteknight07: This should be handled now. `purpose` is required, the migration was regenerated, and the old `course.superCourse` relation was removed.

Whiteknight07:
> * UI so profs can manage agents (names, descriptions, order, add agents. Maybe some consolidated view for deactivating/activating agents, adjusting their prompts, seeing questions across all agents in a unified view, etc.). Maybe we can do this later after verifying that this is something they want.

Im going to make this into a separate PR and do it there. 

> * (tiny) LANTERN is the course name, but the department might want a different name

I'm not very sure what the department wants so for now I'll keep it LANTERN. If they want a different name, I'll put this up as a follow-up on the UI PR. Does that sound right? 



> * Verify that the Convert to Anytime Question feature still works with different agents selected (the conversation _should_ be put into a new anytime question in the main lantern course). Optionally, include what agent the user had selected as part of the questionText when doing the conversion ([should be pretty trivial to do](https://github.com/ubco-db/helpme/pull/561#discussion_r3393895351)).

I did manually test Convert to Anytime Question feature and it works, also added what agent the user had selected when converting to anytime question. (I used userQuestionText instead of questionText as I felt that is a better way to put this info than in questionText)  

<details>
<summary>Screenshot: </summary>
Screenshot 2026-06-18 at 10 27 36 PM

</details>

packages/server/migration/1781189195359-chatbot-agent-course-groups.ts
```diff
@@ -16,16 +19,13 @@ export class ChatbotAgentCourseGroups1781189195359
...
-    await queryRunner.query(
-      `ALTER TABLE "super_course_model" ADD "purpose" "public"."super_course_model_purpose_enum"`,
+      `ALTER TABLE "course_model" DROP COLUMN "superCourseId"`,
     );
     await queryRunner.query(
-      `UPDATE "super_course_model" SET "purpose" = 'course_clone_group' WHERE "purpose" IS NULL`,
+      `CREATE TYPE "public"."super_course_model_purpose_enum" AS ENUM('course_clone_group', 'chatbot_agent_group')`,
     );
     await queryRunner.query(
-      `INSERT INTO "super_course_course_model" ("superCourseId", "courseId") SELECT "superCourseId", "id" FROM "course_model" WHERE "superCourseId" IS NOT NULL ON CONFLICT DO NOTHING`,
```
AdamFipke: you might want to re-add these back to the migration file so it'll migrate our current prod data correctly (also make sure the order is right, like so that we insert into super_course_course_model before dropping course_model.superCourseId). 

Also ideally this custom migration is verified (so that when we push to production and run the migrations it won't error and mess up our data). Though last I checked, it was difficult to manually apply the migrations of just *one* migration file, so the way to check might be to checkout to `main`, create some `super_course_model`s with some random courses, and then manually copy-paste each sql query in this file and run them in order. You won't want to have `yarn dev` running while you do this since typeorm might try to do some auto-synchronize stuff and try to revert your queries while you're running them.

packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx
```diff
@@ -256,6 +278,63 @@ const Chatbot: React.FC<ChatbotProps> = ({
...
+      popupMatchSelectWidth={false}
+      options={[
+        ...(hasHelpMeSystem
+          ? [
+              {
+                label: 'HelpMe System',
+                options: [
+                  {
+                    value: 'System',
+                    label: 'HelpMe System',
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/components/chatbot/Chatbot.tsx]:
```suggestion
                    label: 'System Inquiries',
```

Idk just a thought. I don't really want to have the text "HelpMe System" repeated twice, but I couldn't think of any text that'd be significantly better. 

Image

So doing this should change it so it's:
```
HelpMe System
     System Inquiries
Course Agents
      ...
```

Yeah I think I like that more, but there's probably something even better. I've also thought about removing the tag group for HelpMe System completely, but I kinda like how it parallels with the course agents, since it sorta communicates "Here you can select the System Inquiries agent from HelpMe, or you can select one of your Course Agents". 

I've also been careful to avoid just straitup using "HelpMe" since a lot of profs will introduce our system as the "HelpMe Chatbot" -> meaning students might try to select "HelpMe" from the dropdown expecting a course chatbot only to find it's the system chatbot

packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertChatbotQToAnytimeQModal.tsx
```diff
@@ -79,21 +86,21 @@ const ConvertChatbotQToAnytimeQModal: React.FC<
             generatedAbstract.length > 100 ||
             generatedAbstract.includes('\n')
           ) {
-            generatedAbstract = fallbackGenerateAbstract(generatedAbstract)
+            generatedAbstract = fallbackGenerateAbstract(userQuestionText)
```
Comment (AdamFipke) [inline packages/frontend/app/(dashboard)/course/[cid]/async_centre/components/modals/ConvertChatbotQToAnytimeQModal.tsx]:
```suggestion
            generatedAbstract = fallbackGenerateAbstract(generatedAbstract)
```
Could go either way with this tbh but I think I like how it was before with using the generatedAbstract to create a new one.

AdamFipke (changes requested):
new changes look good, just had a few more small comments

AdamFipke (approved):
looks good! I shall merge

PR #566, opened 2026-06-25
Alerts Rework, Improved Async Question Emails, and background jobs for Upload Chatbot Documents

### LINKED ISSUE #435: Highlight specific Anytime Question/Comment after clicking email/notification link
Problem: You get an email notification about a new comment on your anytime question. You click on the link, but now you're searching through dozens of questions and comments to find where the new one is so you can reply.

Solution: Add a simple query param to all of these email links (inside asyncQuestion.service.ts) that is then handled by the Anytime Question page.tsx. 

You will probably want to be able to pass an anytimeQuestionId and/or commentId, and then:
- expand the corresponding question (and its comment section if commentId is given)
- Scroll to it (using browser API's `scrollTo`)
- Add some kinda subtle animation or highlight (like adding the `glowy` class for a couple seconds, with a fade in and out)

This query param could be re-used once #414 is merged to also use this query param to take you to the exact comment/question it's referring to.

AdamFipke:
# Description

Previous PR: #414 
I've made considerable changes to the original implementation that I figured it'd be better to just re-make it to clean up the PR a little. It's still accessible there in case you want to see the discussions n whatnot.

### What this PR does (major stuff):

**Adds a totally reworked Alert System**

Uses EventSource/SSE to have the clients subscribe to new events regarding alerts (new alert, deleted alert, updated alert (being read)). **This allows the backend to create/delete/update any alert at any time on the backend and the frontend be instantly synchronized.** It even works with multiple tabs open (dismissing an alert on one will dismiss on all)! 

See some implementation details below:

- Uses similar logic to how Queue Questions work with SSE (re-uses most of the backend logic, but the frontend I did fairly differently). Basically: Frontend creates new EventSource to subscribe to SSE events (it opens up a connection with `content-type: 'text/event-stream'` and `connection: 'keep-alive'`). The frontend takes any events, deserializes them, and then loads it into some state variable. In order for the backend to keep track of all the subscribed connections, it stores them all in an array of `directConnections`. Though I guess this comes with scalability concerns if there's multiple instances of the server since each server may have their own list of `directConnections`, but if one server wants to push a new event to another server instance, it can't. Thus, it uses redis's pub/sub system, where each server's SSEService subscribes to redis. Then if one instance publishes all instances receive the event (this whole part is overkill for our case but it's already set up).
- In order to make it so you don't need to remember to do `alertsSSEService.notifyUserOfNewAlert()` in every part of the codebase where an alert is created (or deleted/updated), I created a typeorm subscriber (`alerts.subscriber.ts`) that will automatically send the notifications. **So in order to take advantage of the new alerts system, you can just Create/Delete/Update alerts as normal**, the subscriber takes care of notifying. IMPORTANT NOTE FOR UPDATE & DELETE: Sometimes, not enough data is sent to the subscriber; for example if you do `AlertModel.delete({ where: courseId })` only the `courseId` is passed to the delete subscriber (rather than a list of what alerts were deleted, but sometimes I don't even get the `courseId` meaning I have literally nothing), meaning the users won't have their notifications deleted (a warning will appear in the console if this happens). To get around this, please **first query all alerts (.getMany() or .find()) and *then* delete/update** (e.g. `AlertModel.remove(alertsToDelete)`) (see deleteAdminNoticeAlerts() inside alerts.controller.ts for an example).
- When you first load onto the page (or switch course), it gets most of the unread alerts:
  - First 100 feed alerts, choosing unread first, then by sentAt. (this technically means that if you have more than 100 unread feed alerts, stuff might be a little desynced but meh this would be so rare)
  - All unread MODAL (limit 20) and TOAST alerts (limit 60)
- Has different behaviour based on whether you are within a specific course page or on a `/courses` level page.
  - When on a course page:
    - Only retrieves Modal alerts with either `null` courseId or for the specified course (`null` courseId alerts are like "global" alerts)
  - When not on a course page:
    - Modal: fetches ONLY `null` courseId alerts (since otherwise you'd get spammed with a million alerts from different courses)
  - Toast and Feed alerts will ignore the target courseId and always fetch ALL Toast/Feed alerts (previously I had it so it would only fetch global or course-specific Feed alerts, similar to how Modal works, but I realised that I think it just makes sense have the bell have *all* your notifications on it, especially since each feed alert says what course it's from).
- Old feed alerts (alerts read more than 30 days ago) will be auto-deleted (alerts-clean.service.ts)

Here are a list of the current alerts and what delivery mode each of them are in:
```ts
export const FEED_ALERT_TYPES = [
  AlertType.ASYNC_QUESTION_UPDATE,
  AlertType.ADMIN_NOTICE,
] as const
export const MODAL_ALERT_TYPES = [
  AlertType.REPHRASE_QUESTION,
  AlertType.EVENT_ENDED_CHECKOUT_STAFF,
  AlertType.PROMPT_STUDENT_TO_LEAVE_QUEUE,
  AlertType.ADMIN_NOTICE,
] as const
export const TOAST_ALERT_TYPES = [
  AlertType.CHATBOT_DOCUMENT_PROCESSED,
  AlertType.ADMIN_NOTICE,
] as const
export enum AsyncQuestionUpdateSubtype {
  COMMENT_ON_MY_POST = 'commentOnMyPost',
  COMMENT_ON_OTHERS_POST = 'commentOnOthersPost',
  HUMAN_ANSWERED = 'humanAnswered',
  STATUS_CHANGED = 'statusChanged',
  UPVOTED = 'upvoted',
  ENDORSED = 'endorsed',
}
```

image

It also works with the existing Modal alerts (I tested and made sure it worked with the TA Checkout and PromptStudentToLeaveQueue and RephraseQuestion modals).

image

Adds a new Notification Bell for FEED alerts.
- These alerts are paginated (since you could have a history of thousands of alerts), so it only fetches the first 100 alerts and then fetches more pages as the user gets closer to the end. (I'm proud of managing to work out the state with this)

image

Feed notifications will get dismissed when the bell is closed. But users have the option of viewing dismissed notifications

image
On mobile (notice the hamburger menu also gets the Badge with number of alerts).

image
image

When opening a notification for a new anytime question or anytime question comment, it now takes the user directly to the comment (rather than dumping you on the async_question page and expecting you to find it). Will switch you to the right page and expand comments for the correct question. I updated the Anytime Question Emails as well to use this.

**Adds a "Create Admin Notice" page**
image

- System Admins (userRole == 'admin', NOT org admins) have the ability to create "notices"
- Targets of noticies: specific user, specific course (optionally specific role in course), specific org (optionally specific role in org), or everyone
- Able to create all 3 alert types (Feed, Modal, Toast)
- Will email other admins if someone makes a modal alert that targets more than 1000 users
- These alerts can be deleted, which will instantly update on the frontend for anyone using the system (thanks to how I made the new alert system with SSE)
- I mostly just used this for testing alerts. I don't think it'll be used that much other than for maybe giving a modal alert to all professors about canvas being down or something.

**Reworks the Add Chatbot Document process (again)**

image

Added nice upload bars so you can see the progress for each upload (as well as the size and status of the file).

image

Now when you call the endpoint to upload a document, it just uploads the document and then creates a job (via BullMQ which uses redis), and then when that's done the server creates a Toast alert that is shown on the fronted (This alert is guaranteed to show, even if they close the tab). 

image

Here all the "this document was successfully processed" toasts get merged into one (so the user isn't spammed). Still works when refreshing the page. Error toasts will still be seperate (since they might have big error messages). 

image

Also updates the modal state (in the case that the user decides to not leave the page - but also calls getDocuments() so the table of uploaded documents gets updated).

**Improvements to Async Question Emails**

Now actually shows a card (sorta matching the actual system) inside the email. Most notably this actually provides more information (like what question got the new comment exactly).

image
Staff commenting on your post.

image
Student marking a question as needing faculty help (here I used a different email previewer). You can hover the underlined "private" and it will show "only visible to staff" (`title` attribute, doesn't work on mobile though `:(`)

### Medium sized stuff:
- Adds new alert for when your async question comment was endorsed
- Removed the "Dev tools" page that used to be available to org admins. Either removed or moved the stuff that was there to other (admin-only) pages.
- Upgrades Next.js version (15.3.6 -> 16.3.0). I mainly did this since it upgrades the React version, allowing me to use useEffectEvent (very nice, it's like used for event handlers inside useEffects, I wrote about it inside `useEventSource.ts`). It also comes with some other performance whatevers.
  - **note that middleware.ts is now called proxy.ts**
- Updates NEWDEVS docs to explain state, different ways of synchronizing state (short/long polling, SSE, websockets). Also updated the section on React, Hooks, Tailwind. Added a section on Forms.
- Updated Clone Course so that it also queues stuff in the background and then updates the frontend with a Toast alert (didn't end up making a full BullMQ job for it since it's not worth the work). 

### Minor stuff:
- Added userIds to the Users table (for org admins). Allows you to search too. The Users table now is sorted by userId ASC.
- Fixed a bug where doing a full-delete of a course wouldn't update the frontend state and it would still appear in the list of courses.
- Put the Anytime Question Tags underneath the abstract instead of underneath the comment section
- RephraseQuestionModal now has the option to dismiss the modal vs opening the EditQuestionModal (uses a query param which immediately gets consumed to open the modal). Also fixed an issue previously where it would set the query param but not actually do anything with it.
- Upgraded some other package versions 
  - `class-transformer` and `reflect-metadata` inside common/index.ts and server/index.ts (to match frontend) since otherwise date deserialization wouldn't work. 
- Replaced reconnecting-eventsource with eventsource-client on the frontend. Mainly because I didn't like how reconnecting-eventsource would continuously spam our backend with requests even while down (responding with 500 errors). useEventSource now also has some custom reconnection logic (retries 7 times on dev, 15 times on prod. Does exponential backoff. Also has jitter (so if the server shuts off, every single client won't make a request at the exact same time)).
- Added some code to hopefully suppress the "antd message: static function can not consume context yada please use MessageContextProviderWhatever instead" since I'd actually really rather not. I like being able to just do `message.success()` without needing to add a hook to use it (and there's nothing to really style with them anyway).
- This fixes an issue where sending the "rephrase question" alert wouldn't get sent right away, and how it'd only appear once there's an update to someone else's question (since the queue-sse service only sends events when questions get updated, not when alerts do).
- Fixed issue where not all course settings would get copied over when cloning a course (asyncAuthorsAllowPrivate and that other new one, it's because there was no linter warning to add it there. Now it just clones over everything from CourseSettingsModel regardless).
- Fixed issue where full deleting courses wouldn't work if the course had any queue questions or chatbot questions (certain relations didn't have CASCADE: Interaction -> Chatbot question and Queue -> Queue question. I thought about making them null instead but then that would mean we'd need a million nullish checks everywhere and i don't want to do that. Also we already have a warning that it deletes all questions).

Closes # (issue number)

## Type of change

# How Has This Been Tested?

Been doing a LOT of manual testing to make sure the new alert system works (mostly via using the admin alerts system).

I've also tested that all of the current alerts work with the new system (rephrase question, prompt student to leave queue, event ended checkout staff, clone course, upload chatbot document, create async question comment, endorse comment, async question need help, async question got human answer, probably more).

I've also added some new tests for the new or changed endpoints. Used claude opus with a big ol' prompt and then verified it myself. Yeah there'd be more benefit to making the tests myself (i might find some issues in the process, make better tests, or ) but it's not worth the effort for the amount of long-term value it brings for this size of project.

clone_course integration tests were also updated a bit (they no longer just mock the whole cloneCourse service function), but they're not that comprehensive (only testing if cloning over course settings works, not chatbot documents or anything yet).

# Checklist:


packages/frontend/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev --turbopack",
```
AdamFipke, line 6: turbopack is default enabled now with the new next.js version (I think that includes production builds too? So that might be interesting, hopefully all goes well)

packages/common/index.ts
```diff
@@ -1311,6 +1326,104 @@ export enum asyncQuestionStatus {
...
+
+/**
+ * Anonymous animal names used for anonymizing student identities.
+ * Each student gets a deterministic animal per question via getAnonAnimal().
+ */
+export const ANONYMOUS_ANIMAL_AVATAR = {
+  URL: '/anon_animals',
+  ANIMAL_NAMES: [
+    'Alligator',
+    'Chipmunk',
```
AdamFipke, line 1348: a bunch of these things got moved to `/common/index.ts` from various areas of the async question frontend since they are now used in the backend when constructing emails notices for anytime questions.

packages/frontend/app/components/AlertsContainer.tsx
```diff
@@ -1,82 +0,0 @@
...
-import { useRouter } from 'next/navigation'
-import StudentRephraseModal from '../(dashboard)/course/[cid]/queue/[qid]/components/modals/StudentRephraseModal'
-import { API } from '../api'
-import EventEndedCheckoutStaffModal from '../(dashboard)/course/[cid]/queue/[qid]/components/modals/EventEndedCheckoutStaffModal'
-import PromptStudentToLeaveQueueModal from '../(dashboard)/course/[cid]/queue/[qid]/components/modals/PromptStudentToLeaveQueueModal'
-
-type AlertsContainerProps = {
-  courseId: number
-}
-const AlertsContainer: React.FC<AlertsContainerProps> = ({ courseId }) => {
```
AdamFipke, line 16: renamed to ModalAlertsContainer (the whole pack is: ToastAlertsContainer, ModalAlertsContainer, NotificationBell, and AlertsContext)

packages/frontend/app/contexts/AsyncToasterContext.tsx
```diff
@@ -1,100 +0,0 @@
...
-export const useAsyncToaster = () => useContext(AsyncToasterContext)
-
-const toastOptions: ExternalToast = {
-  richColors: true,
-  dismissible: true,
-  duration: Infinity,
-  closeButton: true,
-}
-
-export const AsyncToasterProvider: React.FC<PropsWithChildren> = ({
```
AdamFipke, line 45: turned into ToastAlertsContainer. Also got rid of runAsyncToast since if there's any job that takes long enough to justify using it, you should instead just create the alert from the backend (that way it still gets sent even if the user closes the page)

packages/frontend/app/hooks/useEventSource.ts
```diff
@@ -1,69 +1,233 @@
...
   count: number
 }
-
 interface SourceAndCount {
-  eventSource: EventSource
+  eventSource: EventSourceClient
   listeners: Record<string, ListenerAndCount>
   isLiveSetters: Set<(live: boolean) => void>
 }
+/* This "EVENTSOURCES" is BS nonsense.
```
AdamFipke, line 14: hi, so i finally went and understood what the hell is going on with useEventSource. I wrote about it here

packages/frontend/app/hooks/useUpdateAlertsWhenLastStaffChecksOut.ts
```diff
@@ -1,30 +0,0 @@
...
-import useSWR from 'swr'
-import { API } from '../api'
-import { useEffect, useMemo } from 'react'
-import { debounce } from 'lodash'
-
-/*
-    This hook will call mutate for useSWR on the alerts data endpoint
-    whenever the last staff member checks out of a queue (only for students).
-*/
-export function useUpdateAlertsWhenLastStaffChecksOut(
```
AdamFipke, line 11: this is one of the big advantages of the new alerts system.

Now we don't need a hook to try to fetch the newly created alert when the last staff member checks out. Now SSE handles this for us

packages/frontend/eslint.config.js
```diff
@@ -20,14 +20,13 @@ const compat = new FlatCompat({
 });
 
 module.exports = defineConfig([
+    ...compat.extends(
```
AdamFipke, line 23: this was something to do with upgrading the eslint version, which i needed to do apparently due to the nextjs version upgrade breaking `eslint --fix` command (idk)

packages/server/src/alerts/alerts.service.ts
```diff
@@ -4,87 +4,54 @@ import {
...
+    deliveryMode: alert.deliveryMode,
+    readAt: alert.readAt,
+    courseId: alert.courseId,
+    courseName: alert.course ? alert.course.name : undefined,
+  };
+};
 
 @Injectable()
 export class AlertsService {
-  async removeStaleAlerts(alerts: AlertModel[]): Promise<Alert[]> {
```
AdamFipke, line 17: removed this (since it's kinda jank to check if you should remove any stale alerts before each and every fetch). Now does it when question status change, on clean queue, and when last staff member checks out.

PR #574, opened 2026-07-21
Add professor access to chatbot agent courses

Whiteknight07:
# Description

This change gives LANTERN professors access to each chatbot agent course.

Before this change, a professor could be a member of the parent course but not the agent courses. Staff endpoints require direct course membership. Thus, these endpoints returned `403` for the professor.

The command `seed:chatbot-agent-group` now does these tasks:

- It finds the professors in the LANTERN parent course.
- It creates each missing agent-course membership with the `professor` role.
- It does not create agent-course memberships for students.
- It does not change an existing membership role or its `favourited` value.
- It documents the setup limits and the membership limits.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] This change requires `yarn install`.
- [ ] This change requires a production environment variable change.
- [ ] This change requires a development environment variable change.
- [ ] This change requires a database query or migration.

# How Has This Been Tested?

- [x] All four tests in `seed.integration.ts` pass.
- [x] The access test verifies that a professor receives `200`.
- [x] The access test verifies that a parent-course student receives `403`.
- [x] ESLint, Prettier, TypeScript, and `git diff --check` pass.

# Checklist

- [x] I reviewed my code.
- [x] I added comments where they are necessary.
- [x] I updated the documentation.
- [x] My changes cause no new warnings.
- [x] I added tests that verify the new function.
- [x] The relevant new and existing tests pass locally.
- [x] All required work is in the `main` branch.
- [x] This change has no user-interface changes.


docs/chatbot_agent_groups.md
```diff
@@ -1,9 +1,9 @@
 # Chatbot Agent Groups
```
AdamFipke: I think maybe this documentation could live as a comment inside `seed-chatbot-agent-group.command.ts`, mostly because this isn't many lines and that way it'd be closer to the actual implementation, but then again it probably doesn't really matter since when you `ctrl+p` to find the agent_groups files, you also see the .md file in docs so idk it's probably fine.

AdamFipke (approved):
Sure looks good I think. Reviewed the logic and I think it makes sense, just find all professors in the main course and adds them to the agent courses if they're not already in it.

I did leave one comment, but it's more of a nitpick more than anything, up to you if you want to follow it

Whiteknight07:
I just realised you did review this PR, ive addressed the comment change and yeah you are right that chatbot agent groups doesnt really warrant a new md file

